PR #19, opened 2019-12-13
add support for using document fields as contexts to filter on during suggest filtering

### LINKED ISSUE #17: Enable suggestion lookup to filter on contexts which are fields on the indexed documents
We need to be able to filter for _contexts_ during suggest lookup. This functionality currently works only when we  build the suggest index from a localSource (i.e. local file). However we do not set up the DocumentDictionary with contexts in case of using [nonLocalSource](https://github.com/Yelp/platypus/blob/master/src/main/proto/luceneserver.proto#L490) i.e. map context to a field already indexed.

Will need to update DocumentDictionary construction duing buildSuggest and update [lookup API](https://github.com/Yelp/platypus/blob/master/src/main/java/org/apache/platypus/server/luceneserver/BuildSuggestHandler.java#L439) to support this.

umeshdangat:
closes #17

### DIFF
--- a/src/test/resources/addSuggestDocsWithContext.csv
+++ b/src/test/resources/addSuggestDocsWithContext.csv
@@ -1,0 +1,3 @@
+text,weight,payload,context
+the cat meows,1,payload1,lucene
+the dog barks,2,payload2,notlucene
--- a/src/main/java/org/apache/platypus/server/luceneserver/BuildSuggestHandler.java
+++ b/src/main/java/org/apache/platypus/server/luceneserver/BuildSuggestHandler.java
@@ -433,16 +433,25 @@ public class BuildSuggestHandler implements Handler<BuildSuggestRequest, BuildSu
             String suggestField = nonLocalSource.getSuggestField();
 
             String payloadField = nonLocalSource.getPayloadField();
+            String contextField = nonLocalSource.getContextField();
 
             DocumentDictionary dict;
 
             if (weightCase.equals(SuggestNonLocalSource.WeightCase.WEIGHTFIELD)) {
                 // Weight is a field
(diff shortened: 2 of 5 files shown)


src/test/resources/addSuggestDocsWithContext.csv
```diff
@@ -0,0 +1,3 @@
+text,weight,payload,context
+the cat meows,1,payload1,lucene
```
felfilali: Not sure if you want to support multiple contexts (and context types) in this review.

I think here context should be a list or a map <string, List> :
could could add 2 geocontext:
geocontext1: {precision:5, boost: 3}
geocontext2: {precision:4, boost: 2}

Another context, using yelp use case as an example, would be the categorie of the a biz:

catcontext1: { names [restaurant, cafes], boost: 3}



I coul

umeshdangat: Lucene seems to  currently support contexts only on AnalyzingInfixSuggester.  Also it seems like it supports only store field for context. It makes term queries for all values on this field with a SHOULD clause.  https://github.com/apache/lucene-solr/blob/master/lucene/suggest/src/java/org/apache/lucene/search/suggest/analyzing/AnalyzingInfixSuggester.java#L537

If we want to support multiple contexts we might need to implement our own Lookup and also our own InputIterator (which itself contains the contexts to be built out in the first place). This is doable and we could probably ticket this if needed.

felfilali: Yes I think we should support it. Can make a ticket and resolve

felfilali: added issue. resolving this for now.

PR #20, opened 2019-12-13
add indexing test for lat_lon point type

umeshdangat:


src/test/resources/registerFieldsLatLon.json
```diff
@@ -0,0 +1,41 @@
...
+    {
+      "name": "license_no",
+      "type": "INT",
+      "multiValued": true,
+      "storeDocValues": true
+    },
+    {
+      "name": "lat_lon",
+      "type": "LAT_LON",
+      "multiValued": true,
```
felfilali: i think some types should default to multivalued like lat_lon

PR #35, opened 2020-01-23
adding Indexing classes and Suggest test

felfilali:
I deleted the previous Draft after applying the recommended changes and made this new one in order to have a cleaner commit history


build.gradle
```diff
@@ -160,4 +160,5 @@ test {
     if (!project.hasProperty('includePerfTests')) {
         exclude '**/YelpReviewsTest.class'
     }
+    systemProperties System.properties
```
umeshdangat: not sure what this line does

felfilali: This transfers the -D params given to ./gradlew to sys.properties in this particular task (test). 
If removed, and you ran `./gradlew test -Dflag:my_flag  --tests "org.apache.platypus.server.YelpSuggestTest", System.properties("my_flag") will result in a NullPointerExcption

umeshdangat: ok, thanks for explaining

felfilali:
@umeshdangat @sarthakn7 was able to run YelpReviewsTest on this branch with the big file. I encountered some issues due to disk space issues. I also made excluded YelpSuggestTest

PR #40, opened 2020-02-04
Add analyzers

sarthakn7:
Added ability to use predefined Lucene analyzers as well as create custom analyzers. 
Note:
1. ConditionalTokenFilter code would compile fine but it's only implementation needs a file as param. Sending a file as a param (i.e. getting file contents, writing it to a file and then using that as a param) isn't supported yet, so this code is useless right now.
2. The analyzers for suggest haven't been modified here - just using standard analyzer for now.


src/main/proto/analysis.proto
```diff
@@ -0,0 +1,41 @@
...
+
+// Used to be able to check if a value was set
+message IntObject {
+    int32 int = 1;
+}
+
+message CustomAnalyzer {
+    repeated NameAndParams charFilters = 1; // Available char filters as of Lucene 8.2.0: htmlstrip, mapping, persian, patternreplace
+    NameAndParams tokenizer = 2;
+    repeated NameAndParams tokenFilters = 3;
```
umeshdangat: probably comment pointing to org.apache.lucene.analysis.TokenFilter and stating the possible tokenFilters to be used?

src/main/java/org/apache/platypus/server/luceneserver/AnalyzerCreator.java
```diff
@@ -0,0 +1,126 @@
...
+
+public class AnalyzerCreator {
+
+    private static final String LUCENE_ANALYZER_PATH = "org.apache.lucene.analysis.{0}Analyzer";
+
+    static Analyzer getAnalyzer(org.apache.platypus.server.grpc.Analyzer analyzer) {
+        if (!analyzer.getPredefined().isEmpty()) {
+            String predefinedAnalyzer = analyzer.getPredefined();
+
+            if ("standard".equals(predefinedAnalyzer)) {
```
umeshdangat: nit: maybe make this  string and "classic" a constant?

src/main/java/org/apache/platypus/server/luceneserver/AnalyzerCreator.java
```diff
@@ -0,0 +1,126 @@
...
+                    String className = MessageFormat.format(LUCENE_ANALYZER_PATH, predefinedAnalyzer);
+                    return (Analyzer) AnalyzerCreator.class.getClassLoader().loadClass(className).getDeclaredConstructor().newInstance();
+                } catch (InstantiationException | IllegalAccessException | NoSuchMethodException | ClassNotFoundException | InvocationTargetException e) {
+                    throw new AnalyzerCreationException("Unable to find predefined analyzer: " + predefinedAnalyzer, e);
+                }
+            }
+        } else if (analyzer.hasCustom()) {
+            return getCustomAnalyzer(analyzer.getCustom());
+        } else {
+            return null;
```
umeshdangat: it seems like the calling code checks for `null` and falls back to using StandardAnalyzer could we potentially just return StandardAnalyzer here followed by a warning log message?

src/main/java/org/apache/platypus/server/luceneserver/BuildSuggestHandler.java
```diff
@@ -117,10 +117,10 @@ private Lookup getSuggester(IndexState indexState, BuildSuggestRequest buildSugg
             int options = 0;
             FuzzySuggester fuzzySuggester = buildSuggestRequest.getFuzzySuggester();
             if (fuzzySuggester.getAnalyzer() != null) {
-                indexAnalyzer = queryAnalyzer = RegisterFieldsHandler.getAnalyzer(indexState, null, "analyzer");
+                indexAnalyzer = queryAnalyzer = AnalyzerCreator.getStandardAnalyzer();
```
umeshdangat: probably want to ticket this ?

src/main/java/org/apache/platypus/server/luceneserver/AnalyzerCreator.java
```diff
@@ -0,0 +1,126 @@
...
+
+            if ("standard".equals(predefinedAnalyzer)) {
+                return new StandardAnalyzer();
+            } else if ("classic".equals(predefinedAnalyzer)) {
+                return new ClassicAnalyzer();
+            } else {
+                // Try to dynamically load the analyzer class
+                try {
+                    String className = MessageFormat.format(LUCENE_ANALYZER_PATH, predefinedAnalyzer);
+                    return (Analyzer) AnalyzerCreator.class.getClassLoader().loadClass(className).getDeclaredConstructor().newInstance();
```
umeshdangat: Should we handle creation of analyzers that take in parameters like stopWords, stemExclusions, etc ? We would also need to pass in the stopwords as a list from the client.

  /** Builds an analyzer with the given stop words.
   * @param stopWords stop words */
  public StandardAnalyzer(CharArraySet stopWords) {
    super(stopWords);
  }
followed by 
analyzer .setMaxTokenLength(maxTokenLength);

or 
            analyzer = new EnglishAnalyzer(stopWords, stemExclusions);

src/main/java/org/apache/platypus/server/luceneserver/BuildSuggestHandler.java
sarthakn7: Added #45 .

src/main/java/org/apache/platypus/server/luceneserver/AnalyzerCreator.java
sarthakn7: Ticketed to address separately - #46 .

src/main/java/org/apache/platypus/server/luceneserver/AnalyzerCreator.java
sarthakn7: Apart from throwing an exception from AnalyzerCreator.getAnalyzer if no analyzer is present, I have also refactored the calling code in RegisterFieldsHandler to hopefully make it more clear.

src/main/java/org/apache/platypus/server/luceneserver/AnalyzerCreator.java
sarthakn7: Done!

src/main/proto/analysis.proto
sarthakn7: There are 101 possible token filters, can't write each one of them. I added links to the Lucene pages though. We can probably have a separate page for listing the token filters in the future if needed.

PR #43, opened 2020-02-05
Enable restoring replica on same host as primary

### LINKED ISSUE #42: Restoring replica on same host as primary fails since backup contains stateDir Path of Primary
When we back up data we back up stateDir and indexes of primary.
StateDir contains a file that has mapping of index name to indexDir Path
```
{
 "test_index" : "/path/to/indexDir/of/primary"
}

```
**Bug repro steps**
1. Run primary and replica on same host
2. Backup primary
3. Down replica 
4. Restore replica using backed up resource info on same host as primary

In current code we re-use the state file for primary from above. This points to primary indexDir instead of the replica indexDir.

One solution is for replica to simply upload its own stateDir in a different namespace and upon restore since the Mode of the server is known to be replica we can download the correct StateDir which in turn points to the old (existing) indexDir path.

This issue does not arise as if we run primary and replica on separate hosts but with same stateDir path.

umeshdangat:
Closes #42 

- Created a separate backupIndex flow for backing up just the state if server is Replica. This allows replica to backup and restore its own state files which contain paths to its indexDir.

- Also added BackupIndexCommand to client so it can be invoked from command line

- Download data for replica from s3 once it comes up. This means we get only the "diff" from primary instead of all the data.


src/main/java/org/apache/platypus/server/luceneserver/Restorable.java
```diff
@@ -38,7 +39,8 @@
     default void restoreDir(Path source, Path target) throws IOException {
         Path baseSource = source.getParent();
         Path tempCurrentLink = baseSource.resolve(getTmpName());
-        Path downloadedData = source.resolve(target.getFileName());
+        Path downloadedFileName = Files.list(source).collect(Collectors.toList()).get(0).getFileName();
```
sarthakn7: This can be `Files.list(source).findFirst().get().getFileName()`

PR #51, opened 2020-02-08
Add match and match phrase queries

### LINKED ISSUE #47: Add match and match phrase queries
Match and MatchPhrase are Elasticsearch query types which analyze the query text to create a boolean query consisting of terms or phrase queries. We can create equivalent queries using `MultiFieldQueryParser` in Lucene.

sarthakn7:
Fixes #47 .


src/test/java/org/apache/platypus/server/grpc/LuceneServerTest.java
```diff
@@ -466,6 +490,66 @@ public void testSearchDisjunctionMaxQuery() throws IOException, InterruptedExcep
...
+        //manual refresh
+        grpcServer.getBlockingStub().refresh(RefreshRequest.newBuilder().setIndexName(grpcServer.getTestIndex()).build());
+
+        SearchResponse searchResponse = grpcServer.getBlockingStub()
+                .search(SearchRequest.newBuilder()
+                        .setIndexName(grpcServer.getTestIndex())
+                        .setStartHit(0)
+                        .setTopHits(10)
+                        .addAllRetrieveFields(RETRIEVED_VALUES)
+                        .setQuery(Query.newBuilder()
```
umeshdangat: Maybe add tests for custom user input analyzer? Same for MatchPhraseQuery

src/main/java/org/apache/platypus/server/luceneserver/QueryNodeMapper.java
```diff
@@ -106,6 +115,40 @@ private DisjunctionMaxQuery getDisjunctionMaxQuery(org.apache.platypus.server.gr
...
 
+    private Query getMatchQuery(MatchQuery matchQuery, IndexState state) {
+        Analyzer analyzer = isAnalyzerDefined(matchQuery.getAnalyzer())
+                ? getAnalyzer(matchQuery.getAnalyzer())
+                : state.searchAnalyzer;
+
+        MultiFieldQueryParser multiFieldQueryParser = new MultiFieldQueryParser(null, analyzer);
+
+        // This created query will be TermQuery if only one token is found after analysis, otherwise BooleanQuery
+        Query query = multiFieldQueryParser.createBooleanQuery(matchQuery.getField(), matchQuery.getQuery(), occurMapping.get(matchQuery.getOperator()));
```
umeshdangat: Seems like the base class QueryBuilder throws exception for any clauses other than MUST and SHOULD. What do you think about making a more restrictive proto clause with only those two options. Might be easier on the user to not worry about RuntimeError due to invalid clause.

```
    if (operator != BooleanClause.Occur.SHOULD && operator != BooleanClause.Occur.MUST) {
      throw new IllegalArgumentException("invalid operator: only SHOULD or MUST are allowed");
    }
```

src/main/java/org/apache/platypus/server/luceneserver/QueryNodeMapper.java
```diff
@@ -106,6 +115,40 @@ private DisjunctionMaxQuery getDisjunctionMaxQuery(org.apache.platypus.server.gr
         return new DisjunctionMaxQuery(disjuncts, disjunctionMaxQuery.getTieBreakerMultiplier());
     }
 
+    private Query getMatchQuery(MatchQuery matchQuery, IndexState state) {
+        Analyzer analyzer = isAnalyzerDefined(matchQuery.getAnalyzer())
+                ? getAnalyzer(matchQuery.getAnalyzer())
+                : state.searchAnalyzer;
+
+        MultiFieldQueryParser multiFieldQueryParser = new MultiFieldQueryParser(null, analyzer);
```
umeshdangat: Would the input queries not have boosts? Seems like MultiFieldQueryParser supports boosts on fields as well.

src/test/java/org/apache/platypus/server/grpc/LuceneServerTest.java
sarthakn7: Added

src/main/java/org/apache/platypus/server/luceneserver/QueryNodeMapper.java
sarthakn7: Added MatchOperator with only MUST and SHOULD.

src/main/java/org/apache/platypus/server/luceneserver/QueryNodeMapper.java
sarthakn7: I will add a query level boost separately (and per-field boost for multi-match).

PR #58, opened 2020-02-20
Add boost query

### LINKED ISSUE #53: Add boost query
Boost query will use [BoostQuery](https://lucene.apache.org/core/8_2_0/core/org/apache/lucene/search/BoostQuery.html) in Lucene.

sarthakn7:
Fixes #53 .

Added a boost parameter to Query which will wrap the Query around a [BoostQuery](https://lucene.apache.org/core/8_2_0/core/org/apache/lucene/search/BoostQuery.html) with the specified boost. Also refactored the query tests:
1. Moved the tests into QueryTest
2. Extracted the common parts of the tests into testQuery function to which only the query to test and the way to test is provided. This function additionally also tests all the queries with a boost.


src/test/java/org/apache/platypus/server/grpc/QueryTest.java
```diff
@@ -0,0 +1,414 @@
...
+        testQuery(querySupplier, responseTester);
+    }
+
+    /**
+     * Search with the query and then test the response. Additional test with boost will also be performed on the query.
+     *
+     * @param querySupplier {@link Supplier} that provides the query to test with
+     * @param responseTester {@link Consumer} that tests a {@link SearchResponse}
+     */
+    private void testQuery(Supplier<Query> querySupplier, Consumer<SearchResponse> responseTester) {
```
umeshdangat: Not sure I understand why we need a Supplier here since the supplier seems to return the static query. Seems like we could simply pass in `Query query` instead and the code would still work?

sarthakn7: The supplier will return a new query object every time its called. It will still work as expected if we pass just Query because the object is not being modified, the supplier was more of a precaution.

umeshdangat: I meant if we change this method signature

`private void testQuery(Supplier<Query> querySupplier, Consumer<SearchResponse> responseTester)`

to 

`private void testQuery(Query query, Consumer<SearchResponse> responseTester)`

This code should still work right?

PR #59, opened 2020-02-22
concurrent multipart s3 upload/download, lz4 compression, client tool for json input

umeshdangat:
* client tool can now accept indexing docs from json files, 1 json doc per line
* use lz4 for compression
* multipart s3 upload and parallel s3 download

lz4 vs gzip speed comparisons for a 12G index.

```
umesh@devbox:~$ ls -ltrh /nail/home/umesh/nrtsearch/compressed.*
-rw-r--r-- 1 umesh users 9.4G Feb 21 12:26 /nail/home/umesh/nrtsearch/compressed.lz4
-rw-r--r-- 1 umesh users 7.5G Feb 21 12:40 /nail/home/umesh/nrtsearch/compressed.gzip
umesh@devbox:~$ du -sh /nail/home/umesh/nrtsearch/primary_index/
12G     /nail/home/umesh/nrtsearch/primary_index/
umesh@devbox:~/pg/platypus (client_add_docs_from_json_files) $ ./tar_impl.sh
buildTar with lz4 took 37 seconds
buildTar with gzip took 848 seconds
umesh@devbox:~/pg/platypus (client_add_docs_from_json_files) $ ./tar_extract_impl.sh
extractTar with lz4 took 59 seconds
extractTar with gzip took 105 seconds
```

* parallel s3 upload took around ~3 mins for 12G index
* parallel s3 download takes around ~10seconds per GB of data

### DIFF
--- a/src/main/java/org/apache/platypus/server/utils/TarImpl.java
+++ b/src/main/java/org/apache/platypus/server/utils/TarImpl.java
@@ -23,6 +23,8 @@
 package org.apache.platypus.server.utils;
 
 import com.google.inject.Inject;
+import net.jpountz.lz4.LZ4FrameInputStream;
+import net.jpountz.lz4.LZ4FrameOutputStream;
 import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
 import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
 import org.apache.commons.compress.archivers.tar.TarArchiveOutputStream;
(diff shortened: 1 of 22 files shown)


src/test/java/org/apache/platypus/server/grpc/GrpcServer.java
```diff
@@ -13,6 +14,8 @@
 import org.apache.commons.csv.CSVParser;
 import org.apache.platypus.server.luceneserver.GlobalState;
 import org.apache.platypus.server.utils.Archiver;
+import org.apache.platypus.server.utils.ArchiverImpl;
```
sarthakn7: Additional imports in this class seem to be added by mistake

src/main/java/org/apache/platypus/server/utils/TarImpl.java
```diff
@@ -126,4 +145,30 @@ private static void addFilestoTarGz(String filePath, String parent, TarArchiveOu
             }
         }
     }
+
+    public static void main(String[] args) throws IOException {
```
sarthakn7: Would it be better to have this in tests (even though it doesn't test with asserts) ?

src/test/java/org/apache/platypus/server/grpc/GrpcServer.java
umeshdangat: removed

src/main/java/org/apache/platypus/server/utils/TarImpl.java
umeshdangat: moved to TarImplTest

Issue #16, opened 2019-12-13
buildDocs broken for Java13


umeshdangat:
buildDocs is broken post Java13 upgrade. Stacktrace [here](https://gist.github.com/umeshdangat/4ca22a91b3cb8a97711143019caba695). Probably the dependency plugin `pseudomuto/protoc-gen-doc` does not work with Java13. 

Workaround: buildDocs still works with Java12. Until this is fixed, ./gradlew buildDocs will work if your JAVA_HOME is set to java12.  The output of buildDocs is an `index.html` file which can be rendered in the browser.

sarthakn7:
The `pseudomuto/protoc-gen-doc` plugin currently does not support JDK 13, as `docker-java` does not support it. We need to update `pseudomuto/protoc-gen-doc` dependency once a new version with JDK 13 support comes out.

Ref: https://github.com/bmuschko/gradle-docker-plugin/issues/885

sarthakn7:
Not doing this since we have replaced the protoc-gen-doc generated docs with swagger docs.

PR #61, opened 2020-03-02
Enable REST support via grpc-rest gateway

umeshdangat:


grpc-gateway/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+
+RUN mkdir -p "$GOPATH/src" "$GOPATH/bin" && chmod -R 777 "$GOPATH"
+#install required go protoc plugins to build grpc-gateway server
+RUN go get -u \
+    github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway \
+    github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger \
+    github.com/golang/protobuf/protoc-gen-go
+
+#run protoc plugin for protoc-gen-go
+RUN protoc -I/code/src/main/proto -I/code/build/extracted-protos/main -I/code/build/extracted-include-protos/main --plugin=protoc-gen-go=$GOPATH/bin/protoc-gen-go --go_out=plugins=grpc:$OUTPATH /code/src/main/proto/analysis.proto /code/src/main/proto/luceneserver.proto /code/src/main/proto/search.proto
```
sarthakn7: Do we need to keep the generated files around ? If we are generating them in Docker during the build, why do we need to keep them in the repo ?

umeshdangat: The purpose of this Dockerfile is to generate 
1. grpc-gateway reverse proxy source code and
2. swagger json files 
These are copied back into src/main/proto once the task buildGrpcGateway is run.

We regenerate them on each run because there could be changes in the src/main/proto/luceneserver.proto thereby needing that source code and swagger.json to be updated.

We keep them in the repo because it 
1. it allows easy access to swagger spec
2. can run REST server out of the box by running `go run http_wrapper <grpc_port> <rest_port>`
3. Seems like it is not a very uncommon thing to keep generated go code in the repo for example:
https://github.com/grpc-ecosystem/grpc-gateway/tree/master/examples/proto/examplepb
https://github.com/elastic/cloud-on-k8s/blob/master/pkg/apis/elasticsearch/v1/zz_generated.deepcopy.go

build.gradle
```diff
@@ -134,30 +130,11 @@ applicationDistribution.into('bin') {
...
-        image 'pseudomuto/protoc-gen-doc'
-        volumes "src/main/docs": "/out", "src/main/proto": "/protos"
-        clean true
-    }
+task buildGrpcGateway(type: Exec) {
+    workingDir = '.'
+    commandLine = './build_grpc_gateway.sh'
 }
 
-task buildDocs() {
```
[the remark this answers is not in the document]
umeshdangat: Its original motivation was to provide documentation. It is currently broken for jdk13 #16  as you mentioned. Since swagger allows us to provide documentation I did not see it necessary to have this broken code exist since its functionality is provided by swagger. If we do plan to have multiple ways to provide documentation then it might make more sense to have working code checked in the repo?

grpc-gateway/Dockerfile
```diff
@@ -0,0 +1,42 @@
+FROM ubuntu:19.10
+
+#RUN add-apt-repository ppa:longsleep/golang-backports
```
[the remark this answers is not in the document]
umeshdangat: done

grpc-gateway/Dockerfile
sarthakn7: 1. I agree with keeping the generated swagger spec, but not sure about the go code
2. Don't get it, don't we still have to run `./gradlew clean installDist` to build the java protos and the jar ? We can generate go files in the same step I believe.
3. Yes, because of go's structure it makes it easy to import the generated code if you keep it. But assuming we're not going to actively work with the go code, not sure if we need to keep them.

If we do keep the files, does it make sense to keep them in a separate package rather than in src/main/proto ?

umeshdangat: > We can generate go files in the same step I believe.

`./gradlew installDist`  outputs the code needed to run java-grpc. This command runs fairly quickly.
`./gradlew buildGrpcGateway` generates the code needed to run rest over grpc. This has a lot deps including go deps like plugins and takes longer time.
When running travis build we run both commands as we need to provide both servers. Idea behind decoupling these two commands and not running them as a single command is:
1. easier developer experience. It is much faster to iterate when changing proto java objects and testing java grpc server only.
2. We might not always want to add every new grpc endpoint to be exposed via REST so we dont need to rebuild  the go server and reverse proxy in case a dev is only interested in modifying the grpc java files.  However eventually  we run builds on travis we will ensure that the correct go server is built.

Overall, I feel there is more benefit in keeping the building of go code separate from java code.
Also I do not see any harm in keeping the generated go code in the repo, since it is a pretty common practice, allows devs to take the rest server for a quick spin, with not many drawbacks of doing so.

> But assuming we're not going to actively work with the go code, not sure if we need to keep them
    Most go files except `http_wrapper.go` are auto generated. For the wrapper code I plan to make changes to it potentially like supporting running REST on same port as grpc.

sarthakn7: Alright. What about 
`If we do keep the files, does it make sense to keep them in a separate package rather than in src/main/proto ?`

This can also be done during the refactor though.

umeshdangat: moved to folder `grpc-gateway`

PR #67, opened 2020-03-07
fix doc value lookup for numeric and sorted_numeric types

umeshdangat:


src/main/java/org/apache/platypus/server/luceneserver/SearchHandler.java
```diff
@@ -842,9 +890,8 @@ public int docID() {
                             advance = sortedNumericDocValues.advanceExact(docID);
                             if (advance) {
                                 for (int i = 0; i < sortedNumericDocValues.docValueCount(); i++) {
-                                    long val = sortedNumericDocValues.nextValue();
-                                    compositeFieldValue.addFieldValue(FieldValue.newBuilder()
-                                            .setLongValue(val));
+                                    Long val = sortedNumericDocValues.nextValue();
```
sarthakn7: We can avoid autoboxing here. It's usages seem to either unbox to primitives or cast to int

src/main/java/org/apache/platypus/server/luceneserver/SearchHandler.java
```diff
@@ -854,8 +901,8 @@ public int docID() {
                             docID = hit.doc - leaf.docBase;
                             advance = numericDocValues.advanceExact(docID);
                             if (advance) {
-                                long val = numericDocValues.longValue();
-                                compositeFieldValue.addFieldValue(FieldValue.newBuilder().setLongValue(val));
+                                Long val = numericDocValues.longValue();
```
sarthakn7: Can avoid autoboxing here too

src/main/java/org/apache/platypus/server/luceneserver/SearchHandler.java
```diff
@@ -962,6 +1009,34 @@ else if (fd.fieldType != null && fd.fieldType.stored()) {
...
+        } else if (fieldValueType.equals(FieldDef.FieldValueType.FLOAT)) {
+            float value;
+            if (isSortedNumeric) {
+                value = NumericUtils.sortableIntToFloat(val.intValue());
+            } else {
+                value = Float.intBitsToFloat(val.intValue());
+            }
+            compositeFieldValue.addFieldValue(FieldValue.newBuilder().setFloatValue(value));
+        } else if (fieldValueType.equals(FieldDef.FieldValueType.BOOLEAN)) {
+            boolean value = val.intValue() == 1 ? true : false;
```
sarthakn7: This can be simplified to `boolean value = val.intValue() == 1 ;`

PR #75, opened 2020-03-27
Alok protobuf make clientlib module

alok-sm:


clientlib/build.gradle
```diff
@@ -0,0 +1,72 @@
...
+    }
+    generateProtoTasks {
+        all()*.plugins { grpc {} }
+    }
+}
+
+publishing {
+    publications {
+        maven(MavenPublication) {
+            groupId = 'org.apache.platypus'
```
sarthakn7: @umeshdangat are we going to change the name of the repo ? If yes, then we probably shouldn't publish to the public maven repo with this name....although I guess there's no harm even if we do?

sarthakn7: I also would like to confirm that it's fine publishing with org.apache in the package when this is not an official apache project but uses apache license.

umeshdangat: Given: https://issues.sonatype.org/browse/OSSRH-56201, @alok-sm and I decided we will simply publish this jar on internal nexus server for now.

alok-sm:
@umeshdangat @sarthakn7 Updated this PR to change the package name

src/main/java/yelp/platypus/server/grpc/LuceneServer.java
```diff
@@ -17,7 +17,36 @@
  *
  */
 
-package org.apache.platypus.server.grpc;
+package yelp.platypus.server.grpc;
```
umeshdangat, line 20: should we have the package name be `com.yelp.nrtsearch`? We were thinking of removing references to the word platypus since its already used for a different project internally.

alok-sm: Makes sense to include that here.

sarthakn7: Also I think the files in main should be under `com/yelp/nrtsearch` directory not just `yelp/nrtsearch`.

alok-sm:
@umeshdangat @sarthakn7 Testing done for the above commit:
./gradlew clean && ./gradlew installDist && ./gradlew test && ./gradlew test -PincludePerfTests=* --tests "com.yelp.nrtsearch.server.YelpReviewsTest.runYelpReviews"  --info

clientlib/build.gradle
```diff
@@ -0,0 +1,79 @@
...
+    }
+    generateProtoTasks {
+        all()*.plugins { grpc {} }
+    }
+}
+
+publishing {
+    publications {
+        maven(MavenPublication) {
+            groupId = 'com.yelp.platypus'
```
sarthakn7, line 67: this should be com.yelp.nrtsearch

alok-sm: good catch! :)

alok-sm:
This build still throws a 
```
> Task :clientlib:compileJava
file or directory '/home/travis/build/Yelp/platypus/clientlib/src/main/java', not found
```

This is fine since it's only a warning saying that it can't find java source files in the clientlib module and this is fine because there isn't any java source files there anyway. This will fix itself in the next PR when I add the clientlib java code here.

PR #82, opened 2020-04-07
make totalHits an object

### LINKED ISSUE #80: totalHits missing in response when hits are zero
I noticed that `totalHits` field in `searchResponse` was completely absent when there were no hits instead of being set to 0. I set the `totalHits` field to 100 instead of setting it to actual hit count in the result, and the field appeared in response with value 100. This happened with the `LuceneServerClient` as well as when using the REST gateway.
My guess is that this has something to do with 0 being default value for long type and either how GRPC encodes the message or how the grpc-gateway and the `LuceneServerClient` are converting the GRPC response into json. Note that we probably won't see this issue when using grpc only since the grpc clients should ideally handle this properly even across platforms.

umeshdangat:
closes #80


sarthakn7:
The tests are failing as they are asserting `searchResponse.getTotalHits()` against ints. Other than that looks good

PR #84, opened 2020-04-08
new endpoint to list all indexes

umeshdangat:
- add indices rpc to allow listing stats for all indices
  - includes numDocs, status, size in bytes of an index
- fix NPE when taxoreader not initialized
- fix NPE when index name not saved/commited

sample usage
`curl -XPOST localhost:8900/v1/indices -d '{}'`

sample output (when index has no data)
```
{
  "indicesResponse": [
    {
      "indexName": "test2",
      "statsResponse": {
        "state": "started",
        "taxonomy": {
          "numOrds": 1,
          "segments": "org.apache.lucene.facet.taxonomy.directory.DirectoryTaxonomyReader@14d0422"
        },
        "currentSearcher": {
          "segments": "IndexSearcher(StandardDirectoryReader(segments:1:nrt); executor=null)"
        }
      }
    },
    {
      "indexName": "test1",
      "statsResponse": {
        "state": "not_started"
      }
    }
  ]
}
```

sample output for index with data
```
{
  "indicesResponse": [
    {
      "indexName": "lists_v0",
      "statsResponse": {
        "maxDoc": 5128793,
        "numDocs": 5128793,
        "dirSize": "11932372929",
        "state": "started",
        "taxonomy": {},
        "currentSearcher": {
          "numDocs": 5128793,
          "segments": "IndexSearcher.....
 }
      }
    }
  ]
}
```


clientlib/proto/yelp/nrtsearch/luceneserver.proto
```diff
@@ -460,10 +467,11 @@ message StatsResponse {
      * counted you should call {@link IndexWriter#commit()} first.
      */
     int32 numDocs = 3;
-    string state = 4; //state of the index
-    Taxonomy taxonomy = 5; //Taxonomy(facets) stats
-    repeated Searcher searchers = 6; //Searcher stats
-    Searcher currentSearcher = 7; //Current Searcher stats
+    int64 dirSize = 4; //size of the this indexDir
```
sarthakn7, line 470: we can add this new field to the stats test too

src/main/java/com/yelp/nrtsearch/server/luceneserver/StatsRequestHandler.java
```diff
@@ -0,0 +1,139 @@
...
+                statsResponseBuilder.setCurrentSearcher(searcher.build());
+            } finally {
+                shardState.release(s);
+            }
+        }
+        return statsResponseBuilder.build();
+
+    }
+
+    public static IndicesResponse getStatsResponse(GlobalState globalState) throws IOException, HandlerException {
```
sarthakn7, line 117: Should this be `getIndicesResponse` ? Should this also be in a separate handler ?

umeshdangat: the handler API doesnt work well for this once since we get an empty request anyway.

clientlib/proto/yelp/nrtsearch/luceneserver.proto
umeshdangat: fixed

PR #85, opened 2020-04-09
enable parallel segment search, use it by default

### LINKED ISSUE #86: If Primary down when Replica coming up, do we choose to use existing replica index or fail replica startup?
When [sendNewReplica()](https://github.com/apache/lucene-solr/blob/531015245042507d71845b6d584e7e7389303093/lucene/replicator/src/java/org/apache/lucene/replicator/nrt/ReplicaNode.java#L286) fails example when Primary is not available then the code fails and prevents replica from starting up.

```
if (t.getMessage().startsWith("replica cannot start") == false) {
        message("exc on start:");
        t.printStackTrace(printStream);
      } else {
        dir.close();
      }
      throw IOUtils.rethrowAlways(t);

```

Choosing consistency:
I guess we could choose to not have replica come up in the case where Primary is not up which is what happens right now.
Choosing availablity:
Since replica downloads data from s3 and it has data upto the last commit point we could alternatively consume the Exception within sendNewReplica() and choose to not throw it. This would mean replica is not connected to Primary but it can still serve search on the existing frozen index.

umeshdangat:
Lucene enabled parallel segment search in this PR: https://issues.apache.org/jira/browse/LUCENE-6294
Another fix for Early termination when numHits is met was backported to 8.3: https://issues.apache.org/jira/browse/LUCENE-8939

This functionality is currently missing from our ES deployments since elasticsearch (at least the version we have in production) does not support it.

Being able to search over segments in parallel should be a crucial performance win (w.r.t latency).

Note: The  "leaves"/"segments" are divided into "leafSlices" in lucene. Each LeafSlice is then searched upon by a thread.  Better Segment To Thread Mapping Algorithm was merged in lucene 9.0 (master): https://issues.apache.org/jira/browse/LUCENE-8757. I am borrowing some of that code here so we can more easily customize how we slice the leaves for individual threads.

Fixes #86 #87

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/StatsRequestHandler.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/StatsRequestHandler.java
@@ -60,14 +60,16 @@ public class StatsRequestHandler implements Handler<StatsRequest, StatsResponse>
         for (Map.Entry<Integer, ShardState> entry : indexState.shards.entrySet()) {
             ShardState shardState = entry.getValue();
             statsResponseBuilder.setOrd(entry.getKey());
-            IndexWriter.DocStats docStats = shardState.writer.getDocStats();
+            if (shardState.writer != null) { //replica
+                IndexWriter.DocStats docStats = shardState.writer.getDocStats();
+                statsResponseBuilder.setMaxDoc(docStats.maxDoc);
+                statsResponseBuilder.setNumDocs(docStats.numDocs);
+            }
             String[] fNames = shardState.indexDir.listAll();
             long dirSize = 0;
             for (int i = 0; i < fNames.length; i++) {
                 dirSize += shardState.indexDir.fileLength(fNames[i]);
             }
-            statsResponseBuilder.setMaxDoc(docStats.maxDoc);
-            statsResponseBuilder.setNumDocs(docStats.numDocs);
             statsResponseBuilder.setDirSize(dirSize);
             // TODO: snapshots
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/yelp/nrtsearch/server/luceneserver/StatsRequestHandler.java
```diff
@@ -60,14 +60,16 @@ private StatsResponse process(IndexState indexState) throws IOException {
         for (Map.Entry<Integer, ShardState> entry : indexState.shards.entrySet()) {
             ShardState shardState = entry.getValue();
             statsResponseBuilder.setOrd(entry.getKey());
-            IndexWriter.DocStats docStats = shardState.writer.getDocStats();
+            if (shardState.writer != null) { //replica
```
sarthakn7, line 63: does this mean that maxDoc and numDocs would be available only for replicas ?

umeshdangat: correct

sarthakn7: can we have an `else` or `else if` that follows the previous behavior so that we get maxDoc and numDocs with primary and standalone nodes as well ?

umeshdangat: sorry the comment is misleading, the code should work already for both replicas and standalone. `writer` is null only for replicas and hence we cannot access the DocStats

umeshdangat: fixed the comment

PR #94, opened 2020-04-28
Keep replica connected to primary on replication port

umeshdangat:
We have made the design decision that a replica can come up even if the primary is down. This allows replica to serve search requests on the data it has e.g. downloaded from s3.  However at some point when the primary is up we want to be able to connect to primary so that we get the newNRT updates. 
  In the approach here the replica
- spawns a new thread in the background once its "started"
- this thread will continue executing until this index is "closed"
- every N time_interval it calls primary to see if it is registered in its list of nodes, if not it sends "addReplica" else does nothing.

In the primary we need to ensure that we maintain only one unique copy per replica. We use a ConcurrentQueue and ensure that the replica (defined as replicaId + hostname + port) is not duplicated in this Queue.  

There are multiple scenarios in which a replica might try to add itself more than once to Primary which is why we ensure we add it only once. 
Example: 
- Just before closing the index the replica found it to be not connected and sent an "addReplica" request.
- There were no newNRT points so Primary never tries to send any data to this now downed index on replica. If it did then primary would remove this replica upon getting an error.
-  Upon StartIndex on the same replica it restarts its background thread to addItself. If the primary did not check for duplicates it would have added this replica twice for the given index.


grpc-gateway/luceneserver.swagger.json
```diff
@@ -299,6 +299,39 @@
         ]
       }
     },
+    "/v1/get_nodes_info": {
```
sarthakn7: I don't see this specified in the proto files. Was something missed ? There are some other inconsistencies too.

clientlib/src/main/proto/yelp/nrtsearch/luceneserver.proto
```diff
@@ -835,3 +838,16 @@ message SearcherVersion {
     int64 version = 1; //returns the version recorded in the commit that the reader opened.  This version is advanced every time a change is made with IndexWriter
     bool didRefresh = 2; //true if refresh happened
 }
+
+message GetNodesRequest {
+    string indexName = 1; //name of the started index whose binary connections we wish to see
+}
+
+message GetNodeResponse {
```
sarthakn7, line 846: Nit: rename this to `GetNodesReponse` (plural to match the request)

src/main/java/com/yelp/nrtsearch/server/luceneserver/GetNodesInfoHandler.java
```diff
@@ -0,0 +1,57 @@
...
+import com.yelp.nrtsearch.server.grpc.NodeInfo;
+import com.yelp.nrtsearch.server.grpc.ReplicationServerClient;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.Collection;
+
+public class GetNodesInfoHandler implements Handler<GetNodesRequest, GetNodeResponse> {
+    Logger logger = LoggerFactory.getLogger(GetNodesInfoHandler.class);
```
sarthakn7, line 36: Nit: make this `private static final`

src/main/java/com/yelp/nrtsearch/server/luceneserver/GetNodesInfoHandler.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    public GetNodeResponse handle(IndexState indexState, GetNodesRequest getNodesRequest) throws HandlerException {
+        GetNodeResponse.Builder builder = GetNodeResponse.newBuilder();
+        ShardState shardState = indexState.getShard(0);
+        if (!shardState.isPrimary() || !shardState.isStarted()) {
+            logger.warn("index \"" + indexState.name + "\" is not a primary or was not started yet");
+        } else { //shard is a primary and started
+            Collection<NRTPrimaryNode.ReplicaDetails> replicasInfo = shardState.nrtPrimaryNode.getNodesInfo();
+            for (NRTPrimaryNode.ReplicaDetails replica : replicasInfo) {
+                ReplicationServerClient replicationServerClient = replica.getReplicationServerClient();
```
sarthakn7, line 47: Might be better to expose `inetSocketAddress` in `ReplicaDetails` and use it to get host and port instead of getting the `ReplicationServerClient`

src/main/java/com/yelp/nrtsearch/server/luceneserver/ShardState.java
```diff
@@ -953,4 +959,43 @@ public Long call() throws Exception {
     }
 
 
+    public static class KeepAlive implements Runnable, Closeable {
+        Logger logger = LoggerFactory.getLogger(KeepAlive.class);
```
sarthakn7, line 963: Nit: make this `private static final`

grpc-gateway/luceneserver.swagger.json
sarthakn7: Resolving since this got fixed right after I reviewed

PR #95, opened 2020-05-05
Add plugin interface

aprudhomme:
Adds interface for dynamically loading plugins to extend server functionality. Includes a basic AnalysisPlugin interface that allows for registration of named Analyzers.

The config can now specify a list of plugin names to load and a path to search for their locations. The loader finds the directory in the search path that matches the plugin name. All jars in the directory are added to a custom class loader. The directory contains a metadata config file that specifies the plugin class name that should be loaded.


src/main/java/com/yelp/nrtsearch/server/plugins/AnalysisPlugin.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.util.Collections;
+import java.util.Map;
+
+/**
+ * Plugin interface for providing custom analysis implementations. Currently allows for the registration
+ * of named {@link Analyzer} providers. These analyzers will be used when the name is provides as the
+ * predefined AnalyzerType for queries and field registration.
+ */
+public interface AnalysisPlugin {
```
sarthakn7, line 32: If there are any issues with the current custom analyzer api, why not extend the current api instead of adding analysis plugin ?

aprudhomme: Extending the custom analyzer api would limit you to just functionality provided by lucene/nrtsearch. This interface allows you to provide a class that extends the lucene Analyzer and does arbitrary analysis processing.

sarthakn7: +1 Makes sense

sarthakn7:
Sorry for not mentioning this before, but would it be possible to put the classes required externally (like the plugin interface) in a separate module like the clientlib ? Can skip for now as well if its too much work.

umeshdangat:
@aprudhomme do you want to address @sarthakn7  comment in this PR or should we merge this PR and make a separate module for `plugins` in another PR? I am good with merging this as is.

aprudhomme:
I'm not sure we are going to want to pull plugins out of the core, as they are fundamentally designed to extend core systems. Also, they are going to be tightly coupled to the server public api, as well as its dependency libraries, through the classloader. This may also involve needing to relocate chucks of core functionality into the plugin module as we add more advanced plugin types.

sarthakn7:
@aprudhomme we don't want to pull plugins out the core, just the interfaces that external plugins would need to implement for e.g. `AnalysisPlugin`

aprudhomme:
@sarthakn7 What exactly are you trying to accomplish by pulling out the interfaces? The plugin is still going to need server classes, like config. Also, the interfaces themselves may need to include core classes in their signatures.

sarthakn7:
@aprudhomme per my understanding as of now we have to add a dependency for the entire nrtsearch jar to external plugins, instead of that we can expose only the classes the external plugins need to build the plugin, lucene dependencies can also be added externally only. The current approach does still work, so we can merge it for now and pull this out later into a separate module if you want to do that instead. @umeshdangat also mentioned that he is fine with merging this as is.

aprudhomme:
@sarthakn7 @umeshdangat I think we should merge this as is.

@sarthakn7 The thing is, the server jar (and its dependencies) is already a runtime dependency. Any classes loaded from the plugin will first be resolved through the parent class loader. If the plugin is not built against the server, it may end up loading unexpected class implementations at runtime (especially if there is a dependency version miss match). The server only needs to be a compile time dependency, so it won't bloat the plugin library.

One thing that was punted on in this review was some kind of enforcement that the plugin was build against the current server version. This will be important to ensure expected runtime behavior.

umeshdangat:
I am okay merging this PR as is.

sarthakn7:
Alright, let's merge this as is.

PR #100, opened 2020-05-12
Added v1/state to return schema, synchronize parallel start_index restore calls

umeshdangat:
Changes:

- POST v1/state
This returns the state of an index once started.  Includes settings, live_settings, fields, suggestors

```
umesh@devXX-XXdevc:~/pg/nrtsearch (add_get_state) $ curl -XPOST localhost:8900/v1/state -d '{"ind
exName": "talk_v1"}' | jq .
{
  "response": "{\"state\":{\"settings\":{\"mergeMaxMBPerSec\":0.0,\"nrtCachingDirectoryMaxMergeSizeMB\":5.0,\"nrtCachingDirectoryMaxSizeMB\":60.0,\"concurrentMergeSchedulerMaxThreadCount\":8,\"concurrentMergeSchedulerMaxMergeCount\":16,\"indexVerbose\":false,\"indexMergeSchedulerAutoThrottle\":false,\"directory\":\"MMapDirectory\"},\"liveSettings\":{\"maxRefreshSec\":60.0,\"minRefreshSec\":0.5,\"indexRamBufferSizeMB\":250.0},\"fields\":{\"geoquad\":{\"name\":\"field1\",\"type\":\"INT\",\"search\":true,\"storeDocValues\":true,\"multiValued\":true},\"id\":{\"name\":\"id\",\"type\":\"INT\",\"storeDocValues\":true,\"multiValued\":true},\"text\":{\"name\":\"text\",\"type\":\"TEXT\",\"search\":true,\"storeDocValues\":true,\"tokenize\":true,\"multiValued\":true},\"field2\":{\"name\":\"field2\",\"type\":\"INT\",\"storeDocValues\":true,\"multiValued\":true},\"title\":{\"name\":\"title\",\"type\":\"TEXT\",\"search\":true,\"storeDocValues\":true,\"tokenize\":true,\"multiValued\":true},\"timestamp\":{\"name\":\"timestamp\",\"type\":\"INT\",\"storeDocValues\":true}},\"suggest\":{}}}"
}

```

- GET v1/status
Made this endpoint a GET

- Fixed issues in case two parallel start_index in restore mode calls made. We download data only once while the other call fails.

Error outputs when issuing parallel restore start_index commands:

```
umesh@devXX-XXdevc:~/pg/nrtsearch (add_get_state) $ curl -XPOST localhost:8900/v1/start_index -d @/nail/home/umesh/scratch/nrtsearch/talk/startIndexReplicaRestore.json
{"error":"error while trying to start index: talk_v1\nIndex talk_v1 already restored","code":3,"message":"error while trying to start index: talk_v1\nIndex talk_v1 already restored"}

```


src/main/java/com/yelp/nrtsearch/server/luceneserver/StartIndexHandler.java
```diff
@@ -58,9 +57,19 @@ public StartIndexResponse handle(IndexState indexState, StartIndexRequest startI
         final String primaryAddress;
         final int primaryPort;
         Path dataPath = null;
-        if (startIndexRequest.hasRestore()) {
-            RestoreIndex restoreIndex = startIndexRequest.getRestore();
-            dataPath = downloadArtifact(restoreIndex.getServiceName(), restoreIndex.getResourceName(), INDEXED_DATA_TYPE.DATA);
+        if (startIndexRequest.hasRestore() && !shardState.isStarted()) {
```
sarthakn7, line 60: This will make sure that we download the data and start index only once when restoring, but won't handle parallel start index requests without restore. Is that fine ?

umeshdangat: correct. Parallel start Index requests without restore are handled in [shardState.startPrimary](https://github.com/Yelp/nrtsearch/blob/master/src/main/java/com/yelp/nrtsearch/server/luceneserver/ShardState.java#L576) (or shardState.startReplica)  That method is synchronized and once a thread enters it, first thing it does is checks if index is already started and throws IllegalStateException

PR #105, opened 2020-05-19
support prometheus exposition format for metrics endpoint

umeshdangat:
/status/metrics endpoint needs to emit prometheus based metrics on the [supported exposition format](https://prometheus.io/docs/instrumenting/exposition_formats/). 

This is little tricky since our gRPC server uses protoBuf format. We also have a reverse-proxy (grpc-gateway) which uses json format (for all req/responses till now).  We need to support text/plain format for metrics endpoint. 

This can be done as follows: 
- The gRPC server emits metrics response as a  [google.api.HttpBody] (https://github.com/googleapis/googleapis/blob/master/google/api/httpbody.proto) proto object.
- Next we configure our reverse proxy to read HttpBody responses using a special marshaler. 
This allows the reverse proxy to read this data as binary rather than using the default marshaler which is Json.
https://grpc-ecosystem.github.io/grpc-gateway/docs/httpbody.html

Sample output:
https://gist.github.com/umeshdangat/8f879631fc35c7a4a0594f92cb024f72


umeshdangat:
@vim345 Yes the eventual implementation is pretty neat, but I had to dig up the interface for [marshaler](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/marshaler.go#L8) . Initially I was thinking of writing my own marshaler but the [HTTPBodyMarshaler](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/marshal_httpbodyproto.go#L39) seemed to fit this use case nicely.   We just have to make sure in that our RPC returns HTTPBody type.

Re: glog, I just used whatever the grpc-gateway [source examples had ](https://github.com/grpc-ecosystem/grpc-gateway/blob/db7fbefff7c04877cdb32e16d4a248a024428207/examples/internal/server/a_bit_of_everything.go#L9). Dont mind changing it to something else to.

PR #112, opened 2020-05-28
Add ScriptService, ScriptPlugin, generalized doc value loading

aprudhomme:
Scripts
Rather than defining a javascript expression, virtual fields are now defined by a Script message which contains:
lang: Script language specifier
source: Script source to compile
params: map of String to simple type, available to script execution

ScriptService
This class is responsible for compiling a script definition into a concrete script factory. It manages the set of registered ScriptEngine, and delegates compilation to the engine handling the script's language. Compiled factories are cached in an LRU cache to avoid recompilation of identical scripts.

The javascript expression handling has been moved into a ScriptEngine for the language 'js'.

ScoreScript
Script that produces a double value for each document. Compiling a ScoreScript produces a factory for creating a lucene DoubleValueSource. This script type is used for virtual fields and FunctionScoreQuery. Script execution has access to the document's score, doc values, and the script params.

Doc values
Reading doc value data for a document is now wrapped by LoadedDocValues<> implementations, which present a List interface of the appropriate value type. These values are made available to scripts through the DocLookup passed during creation. These are also used when filling doc value fields for the query response.

ScriptPlugin
A plugin interface has been added to allow for custom ScriptEngine implementations to be provided.


src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
```diff
@@ -112,15 +113,11 @@ private PhraseQuery getPhraseQuery(com.yelp.nrtsearch.server.grpc.PhraseQuery ph
...
-        Expression expr;
-        String exprString = functionScoreQuery.getFunction();
-        try {
-            expr = JavascriptCompiler.compile(exprString);
-        } catch (ParseException pe) {
-            // Static error (e.g. bad JavaScript syntax):
-            throw new IllegalArgumentException(String.format("could not parse expression: %s", exprString), pe);
-        }
-        return new FunctionScoreQuery(getQuery(functionScoreQuery.getQuery(), state), expr.getDoubleValuesSource(state.exprBindings));
+        ScoreScript.Factory scriptFactory = ScriptService.getInstance().compile(functionScoreQuery.getScript(), ScoreScript.CONTEXT);
```
umeshdangat, line 116: Nice ! This will allow us to use to binds script parameters into the expression using `getDoubleValuesSource(bindings)`.

JavascriptCompiler.compile(exprString) is a hotspot right now for us

src/main/java/com/yelp/nrtsearch/server/utils/LRUCache.java
```diff
@@ -0,0 +1,235 @@
...
+        T value;
+        volatile long lastAccessTimeMs;
+
+        CacheEntry(T value) {
+            this.value = value;
+            lastAccessTimeMs = System.currentTimeMillis();
+        }
+    }
+
+    private final Map<K, CacheEntry<V>> cache = new HashMap<>();
```
umeshdangat, line 51: could we use `LinkedHashMap` here to get the LRU behavior like below ? Ofc you will need to still use the locks for concurrency?

```
public LRUCache(int capacity) {
            CAPACITY = capacity;
            map = new LinkedHashMap<Integer, Integer>(capacity, 0.75f, true){
                protected boolean removeEldestEntry(Map.Entry eldest) {
                    return size() > CAPACITY;
                }
            };
        }
        public int get(int key) {
            return map.getOrDefault(key, -1);
        }
        public void set(int key, int value) {
            map.put(key, value);
        }
```

sarthakn7: Can also use https://github.com/ben-manes/caffeine

aprudhomme: I guess it would be worth trying the Guava cache. It will occasionally add some extra work on get calls, which I was trying to avoid, but it does simplify things.

umeshdangat:
@aprudhomme I will merge this once we fix our release process upstream so we dont break our existing client queries !

PR #118, opened 2020-06-15
Use generalized config reader

aprudhomme:
Adds a YamlConfigReader to allow lookups into configuration by providing a dot separated key path. Accessor are provided for simple types (bool,int,long,float,double,str) and lists of simple types. The generic accessors allow for extensibility (and could perhaps be combined with an ObjectMapper).

This will allow config reading to move outside the main LuceneServerConfiguration class, which will make configuring different core services easier and allow plugins to lookup custom config.


umeshdangat:
@aprudhomme The integration tests seem to fail  while trying to parse the yaml
https://travis-ci.com/github/Yelp/nrtsearch/builds/171296014#L13949

aprudhomme:
@umeshdangat fixed the test.

PR #125, opened 2020-06-24
increase max message size on channel

umeshdangat:
We need to increase grpc max message length as mentioned here https://github.com/tensorflow/serving/issues/1382 in order to avoid getting the below error on larger responses:

{"error":"grpc: received message larger than max (7750076 vs. 4194304)","code":8,"message":"grpc: received message larger than max (7750076 vs. 4194304)"}


ubaidillahSriyudi:
does anyone know how to do it on Dart Flutter ? 
i got gRPC: Received message larger than max, i use go for the server

PR #128, opened 2020-06-25
set max grpc message size on server

umeshdangat:
In #125 we increase message size limit on the channel(client) side but seems like we also need to do it on the server side as per 
https://github.com/grpc/grpc-java/search?q=defaults+to+4+MiB&unscoped_q=defaults+to+4+MiB
and https://github.com/grpc/grpc/issues/7927#issuecomment-244993104


umeshdangat:
@sarthakn7 build failed due to spotless, re-ran `spotlessApply`. This should hopefully allow us to run a MatchAllDocs query (queryText field absent) to retrieve all indexed ids without the grpc_max_size  violation issue now

PR #133, opened 2020-07-07
Add structure to field definition classes

aprudhomme:
A first pass at refactoring FieldDef to give it more structure and move type specific code into discrete classes. The goal is to make this all more understandable and make it easy to to add/augment field type.

There is a general hierarch:
- Field
  - Virtual Field
  - Indexable Field
    - Text Fields
    - Other fields (Numbers/bool/data time/lat lon/etc)

For field properties that didn't really fit the hierarchy well, I added them as optional trait interfaces (Sortable/Bindable/TermQueryable/etc).

In general, I tried to stay pretty close to the status quo logic, though some of it looks a bit wrong. Things like atom fields always being searchable, whether you ask for it or not. Let me know what things should be addressed here. Also let me know if you think some properties are in the wrong place (should analyzers only apply to text fields?).

This is meant to be a move in the right direction. Things like highlighting and facets will probably need to change quite a bit, it is just hard to do that now without those features being functional.


src/main/java/com/yelp/nrtsearch/server/luceneserver/field/IndexableFieldDef.java
```diff
@@ -0,0 +1,268 @@
...
+import org.apache.lucene.document.Document;
+import org.apache.lucene.document.FieldType;
+import org.apache.lucene.index.DocValuesType;
+import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.search.similarities.BM25Similarity;
+import org.apache.lucene.search.similarities.ClassicSimilarity;
+import org.apache.lucene.search.similarities.Similarity;
+
+/** Base class for all field definition types that can be written into the index. */
+public abstract class IndexableFieldDef extends FieldDef {
```
umeshdangat, line 32: could this be an interface with `default` methods instead?

aprudhomme: You could change them to interfaces, but the default functions wouldn't be helpful in most cases. Since the interface can't store state, every type would need to redefine the standard functions (getName/hasDocValues/isMultiValue/etc) and do its own grpc Field definition processing. It felt like it was going to be a lot of code duplication, so I was trying to use the abstract class to structure the field building through its protected methods. I am open to discussion on this.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/BooleanFieldDef.java
```diff
@@ -0,0 +1,133 @@
...
+
+    if (hasAnalyzer(requestField) && !requestField.getSearch()) {
+      throw new IllegalArgumentException("No analyzer allowed when search=false");
+    }
+  }
+
+  @Override
+  protected DocValuesType parseDocValuesType(Field requestField) {
+    if (requestField.getStoreDocValues() || requestField.getSort() || requestField.getGroup()) {
+      if (requestField.getMultiValued()) {
```
karthikalle, line 61: Up for discussion: Do we want to support multiple values for boolean fields, or validate and enforce single values?

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/IndexableFieldDef.java
```diff
@@ -0,0 +1,268 @@
...
+            : requestField.getDocValuesFormat();
+
+    String similarityStr = requestField.getSimilarity();
+    if (!similarityStr.isEmpty()) {
+      if (similarityStr.equalsIgnoreCase("DefaultSimilarity")) {
+        this.similarity = new ClassicSimilarity();
+      } else if (similarityStr.equalsIgnoreCase("BM25Similarity")) {
+        // TODO pass custom k1, b values so that we can call Constructor BM25Similarity(float k1,
+        // float b)
+        this.similarity = new BM25Similarity();
```
karthikalle, line 95: Do we want `Postings` and `Similarity` in `IndexableFieldDef`? Since these might only be needed for text fields, could we move them to `TextBaseFieldDef`?

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/BooleanFieldDef.java
aprudhomme: I would lean towards disallowing multivalued boolean fields altogether. However, I would punt this until later. The boolean type generally needs some work. It currently can't even be used in term queries.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/IndexableFieldDef.java
aprudhomme: I wasn't entirely sure the best level to put these properties. But if that is the case, I can move them.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/AtomFieldDef.java
```diff
@@ -0,0 +1,142 @@
...
+      } else {
+        return DocValuesType.BINARY;
+      }
+    }
+    return DocValuesType.NONE;
+  }
+
+  @Override
+  protected void setSearchProperties(FieldType fieldType, Field requestField) {
+    // This is how things are currently expected, if the field is searchable it
```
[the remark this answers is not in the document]
aprudhomme: I too am unclear on what a highlighted keyword would look like. Highlighting isn't really functional at the moment anyways.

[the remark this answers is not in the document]
karthikalle: Makes sense to fix this later.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/BooleanFieldDef.java
karthikalle: Sounds good, we could do this later.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/IndexableFieldDef.java
karthikalle: We could also do it in a subsequent review. Closing for now.

aprudhomme: Yeah. I was going to move it quickly, but boolean needs posting because it is indexed. I think we should fix boolean first and sort this out later.

PR #137, opened 2020-07-20
Restructure search, splitting into query and fetch phases

aprudhomme:
A first attempt to clean up the search logic and pull it out of SearchHandler. The search request is processed into a SearchContext with all the information needed to execute the query. The search is performed by the SearchExecutor based on this context info.

Search execution is divided into a query phase (find top docs and scores) and fetch phase (populate fields in response). This may be useful in the future if we want a coordinator service to support query then fetch.

The goal was to structure things in such a way that there will be more natural points of extension for future search features (rescore, aggregations, custom fetch ops, etc).

A lot of the facet/highlighting logic has been removed. This code was not functional and cluttered up the search logic. It should be added back in independently, with purpose and tests.

This is meant as a request for comments. Existing test coverage of core functionality is a bit spotty. I would want to improve that a bit before merging something like this.


umeshdangat:
> Search execution is divided into a query phase (find top docs and scores) and fetch phase (populate fields in response). This may be useful in the future if we want a coordinator service to support query then fetch. The goal was to structure things in such a way that there will be more natural points of extension for future search features (rescore, aggregations, custom fetch ops, etc).

 I am not sure we want to be as opinionated (at least at the moment) as elasticsearch when it comes to having such phases since we are not doing a scatter gather differentiation across nrtsearch nodes anytime soon. The more important immediate tasks are supporting facets, highlights, score components, custom similarity. 


-Adding these to the existing codebase should in fact be easier in the existing codebase. Currently  we have access to raw Lucene objects like Collectors, adding Custom collectors, Sort, SortComparators, Custom DrillSidewaysQuery etc. All these should allow us to build in the functionality we need to extend our features. Here is an example branch of a facets implementation on the existing codebase: https://github.com/Yelp/nrtsearch/pull/146.
-We were able to support all the above features in Lucy (which was not elasticsearch based)
-it seems easier to add these features now using whatever tools lucene provides us (e.g. see how I store facets in the custom  DrillSidewaysImpl using state) rather than being restricted by these new interfaces and having to refactor them as we add things (e.g. VirtualFieldDef not being indexable and hence making it harder to support facets in current codebase)

I do agree that we need to refactor some of the SearchHandler and make it more modular but probably easier to get some of these features working with more testing before we do major refactors so we
- are more certain of any potential regressions
- have a more complete view of things to move around in order to support some advanced features


Overall, I feel like we should try to get some of the above features implemented with better test coverage. Post this/or if we are unable to add in these features we can consider refactoring this codebase. This approach probably saves us time with respect to getting things working faster.

aprudhomme:
I think we are going to need to start getting opinionated on some of these things. The SearchHandler is not a sustainable model. Yes, having the bulk of search logic in single large class allows for a lot of flexibility, but maintenance and developer understandability will suffer.

We need to create some sort of structure. Query and fetch may not be lucene concepts, but they create a very clear separation of concerns. If the intent in to have external parties eventually use/develop within this system, we are going to need to start giving more serious though to our code organization and interfaces.

Oh course, it would be great to have a better idea exactly what features will need to be supported in the future, but how long do you wait. The longer we wait to do an overhaul, the harder it will be. The FieldDef refactor was not perfect, but I would argue it had a net positive effect on the usability of the codebase.

I agree that test coverage is an issue. Though, it is a pretty big issue in general and should probably get a priority bump when possible.

That being said, if this magnitude of this change is un agreeable, I think there are still some good things here that can be integrated into the current search logic. Retrieving fields by segment slice should be more resource efficient and allow for a clear custom retrieval operation interface to be created.

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/SearchRequestProcessor.java
```diff
@@ -0,0 +1,284 @@
...
+
+    Query rewrittenQuery = context.searcherAndTaxonomy().searcher.rewrite(query);
+    diagnostics.setRewrittenQuery(rewrittenQuery.toString());
+
+    Query drillDownQuery = addDrillDowns(context.indexState(), rewrittenQuery);
+    diagnostics.setDrillDownQuery(drillDownQuery.toString());
+
+    context.setQuery(drillDownQuery);
+    context.setCollector(buildCollector(context, searchRequest));
+    context.searchResponse().setDiagnostics(diagnostics);
```
umeshdangat, line 103: it is good that we have access to a mutable searchResponse within the searchcontext. Maybe we need to change DrillSidewaysImpl to accept searchReponse in its constructor to fill out the  facetResult list.

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/SortParser.java
```diff
@@ -0,0 +1,142 @@
...
+          throw new IllegalArgumentException(
+              String.format("field: %s does not support sorting", fieldName));
+        }
+
+        sf = ((Sortable) fd).getSortField(sub);
+      }
+      sortFields.add(sf);
+    }
+
+    return new Sort(sortFields.toArray(new SortField[0]));
```
umeshdangat, line 81: Should we  allow the SortField to take in a custom `FieldComparatorSource`. 

```
/** Creates a sort, possibly in reverse, with a custom comparison function.
   * @param field Name of field to sort by; cannot be <code>null</code>.
   * @param comparator Returns a comparator for sorting hits.
   * @param reverse True if natural order should be reversed.
   */
  public SortField(String field, FieldComparatorSource comparator, boolean reverse) {
    initFieldType(field, Type.CUSTOM);
    this.reverse = reverse;
    this.comparatorSource = comparator;
  }
```

Not entirely sure how this would work on our custom scoring scripts since for virutal fields this seems to create a `FieldComparator.DoubleComparator`. Wondering if might be good to have our own custom comparator. Probably add a test for now to show how this might work on Script/virtual fields?

umeshdangat:
@aprudhomme Do you want to try rebasing this on master?

aprudhomme:
@umeshdangat I merged in master and got the facet tests working. I have also tried to refine the collector classes and add interfaces for custom collection. I provided a simple max value collector as an example.

aprudhomme:
This ended up getting a bit big. I am going to break this down into some smaller PRs.

PR #144, opened 2020-08-05
Added serviceName and nodeName to prometheus metrics

erikyangs:
**Summary**
serviceName and nodeName are now dimensions in `/status/metrics`:
```grpc_server_handled_latency_seconds_count{grpc_type="UNARY",grpc_service="luceneserver.LuceneServer",grpc_method="metrics",serviceName="nrtsearch-generic",nodeName="lucene_server",} 3.0```

**Changes**
Most files are copy-pasted from me.dinowernli:java-grpc-prometheus. Classes in `java-grpc-prometheus` could not be extended because of classes being package private (e.g. `ServerMetrics`) and no default constructors being available (e.g. `LuceneServerMonitoringServerInterceptor`). me.dinowernli:java-grpc-prometheus is [under Apache 2.0 License](https://github.com/grpc-ecosystem/java-grpc-prometheus/blob/master/LICENSE).

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
+++ b/src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
@@ -125,8 +125,7 @@ public class LuceneServer {
                 .withLatencyBuckets(luceneServerConfiguration.getMetricsBuckets())
                 .withCollectorRegistry(collectorRegistry),
             serviceName,
-            nodeName
-        );
+            nodeName);
     /* The port on which the server should run */
     server =
         ServerBuilder.forPort(luceneServerConfiguration.getPort())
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java [hunk 1]
```diff
@@ -81,8 +83,6 @@
 import java.util.concurrent.Future;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.stream.Collectors;
-import me.dinowernli.grpc.prometheus.Configuration;
```
umeshdangat, line 84: Hmm, this is unfortunate that we basically cannot extend `me.dinowernli`.  I can think of following alternatives

1. Use [raw prometheus simpleclient](https://github.com/prometheus/client_java#instrumenting) which this library was doing anyway. Might be more re-work though.

2. Maybe use [asm](https://dzone.com/articles/fully-dynamic-classes-with-asm) to generate dynamic class that could extend ServerMetrics 

3. use your implementation here as is but since we are pretty much rolling out our own implementation of `me.dinowernli`  should we remove that dep from the build.gradle? You will probably also need to fix the tests using it in  `GrpcServer`

As long as this allows us to easily add new metrics going forward I am okay with the current approach you take.

PR #145, opened 2020-08-06
Added configuration option for buffer size for addDocuments

erikyangs:
Added configuration option in `LuceneServerConfiguration` to control addDocuments buffer length.

**Example in configuration.yaml**
```
addDocumentsMaxBufferLen: 1000
```


src/main/java/com/yelp/nrtsearch/server/config/LuceneServerConfiguration.java
```diff
@@ -46,6 +46,7 @@
   private static final Path DEFAULT_PLUGIN_SEARCH_PATH =
       Paths.get(DEFAULT_USER_DIR.toString(), "plugins");
   private static final String DEFAULT_SERVICE_NAME = "nrtsearch-generic";
+  private static final int DEFAULT_ADD_DOCUMENTS_MAX_BUFFER_LEN = 100;
```
umeshdangat, line 49: The PR seems fine but I was wondering if we want to make the MAX_BUFFER_LEN as per jvm start config like you do here or should we make it a part of the AddDocumentRequest so that each indexing request  could have a different setting and we do not need to restart the server for this to take effect. One example would be to run a backfill vs real time indexing.

src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
```diff
@@ -495,25 +496,74 @@ public void startIndex(
...
+            new ConcurrentHashMap<>();
+        // Map of {indexName: count}
+        Map<String, Long> countMap = new ConcurrentHashMap<>();
+        private static final int DEFAULT_MAX_BUFFER_LEN = 100;
+
+        private int getAddDocumentsMaxBufferLen(String indexName) {
+          try {
+            return globalState
+                .getIndex(indexName)
+                .getSaveState()
```
sarthakn7, line 511: @erikyangs sorry didn't look at this PR before you merged it. The `getSaveState` method is synchronized https://github.com/Yelp/nrtsearch/blob/master/src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java#L636 and we are calling this for every batch of requests. Apart from being slow, this might also reintroduce the deadlocks we had before.  I was thinking about adding a getter for `addDocumentsMaxBufferLen` in `IndexState`, but that would also be slightly slow since the variable is volatile.
cc: @umeshdangat

umeshdangat: +1 on adding getter on `volatile int addDocumentsMaxBufferLen`.  It should be faster to read volatile variable than  read from JSONObject.

PR #146, opened 2020-08-07
add facets

umeshdangat:
Added support for facets of type:
- NUMERIC_RANGE see [RangeFacetsExample](https://github.com/apache/lucene-solr/blob/d7dc53ff7c3a16110aac4120e5bfdf7721b21bcd/lucene/demo/src/java/org/apache/lucene/demo/facet/RangeFacetsExample.java) for reference. This works on numeric field types to collect facets in ranges. Note these dont work on multivalued numeric fields though.

- SORTED_SET_DOC_VALUES see [SortedSetFacetsExample](https://github.com/apache/lucene-solr/blob/d7dc53ff7c3a16110aac4120e5bfdf7721b21bcd/lucene/demo/src/java/org/apache/lucene/demo/facet/SimpleSortedSetFacetsExample.java) for reference. This works on text field types. 
Pros compared to HIERARCHY type 
we dont need to maintain a separate index for facets
Cons compared to HIERARCHY type
 we cannot structure terms by hierarchy 

- HIERARCHY see [MultiCategoryListsFacetsExample](https://github.com/apache/lucene-solr/blob/d7dc53ff7c3a16110aac4120e5bfdf7721b21bcd/lucene/demo/src/java/org/apache/lucene/demo/facet/MultiCategoryListsFacetsExample.java) for reference. This works on  text field types.  All the unique values for a taxonomy facet are stored in a separate Lucene index (often called the sidecar index). But, this implementation supports hierarchical facets.
during index time request can contain list of  `paths` which determine the hierarchy and during search time request can bucket at required level of hierarchy.

- FLAT Similar HIERARCHY in that it uses taxonomy index (sidecar) but does not deal with hierarchy.

Simplified tests using base class from branch: https://github.com/Yelp/nrtsearch/pull/137/commits/905d4a2a9b4f270eeb74862f1401d1f2bb370528

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/facet/DrillSidewaysImpl.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/facet/DrillSidewaysImpl.java
@@ -330,12 +330,6 @@ public class DrillSidewaysImpl extends DrillSideways {
               String.format("each facet request must have either topN or labels"));
         }
       }
-      //
-      //            if (facetResult == null) {
-      //                facetResults.add(null);
-      //            } else {
-      //                facetResults.add(buildFacetResultJSON(facetResult));
-      //            }
       if (facetResult != null) {
         grpcFacetResults.add(buildFacetResultGrpc(facetResult));
       }
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/facet/DrillSidewaysImpl.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/facet/DrillSidewaysImpl.java
@@ -32,6 +32,7 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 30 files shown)


src/main/java/com/yelp/nrtsearch/server/luceneserver/facet/DrillSidewaysImpl.java
```diff
@@ -0,0 +1,363 @@
...
+            c = drillDowns;
+          }
+
+          LongRangeFacetCounts longRangeFacetCounts =
+              new LongRangeFacetCounts(indexableFieldDef.getName(), c, ranges);
+          facetResult =
+              longRangeFacetCounts.getTopChildren(
+                  0,
+                  indexableFieldDef.getName(),
+                  facet.getPathsList().toArray(new String[facet.getPathsCount()]));
```
karthikalle, line 182: Not needed in this review: Do we want to structurally put these in individual FieldDefs? `FacetResult getFacetResults(facet, dsDimMap, drillDowns`?

PR #147, opened 2020-08-10
Removed finishIndexingJob

sarthakn7:
Removed `finishIndexingJob` when adding documents, and added refresh to tests with each add document call


umeshdangat:
```
com.yelp.nrtsearch.server.grpc.LuceneServerTest > testSearchPostUpdate FAILED
    java.lang.AssertionError: expected:<2> but was:<0>
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:645)
        at org.junit.Assert.assertEquals(Assert.java:631)
        at com.yelp.nrtsearch.server.grpc.LuceneServerTest.testSearchPostUpdate(LuceneServerTest.java:344)

```

Seems like you might want to try adding a manual `refresh` to these tests?

PR #149, opened 2020-08-13
Add faceting on virtual fields

umeshdangat:
Enables support for numeric_range faceting on VirtualFields. This allows us to facet/group by custom scripts. See `testVirtualField` for example. Scripts give us a DoublesValueSource which is used to bucket results by.


umeshdangat:
ping @sarthakn7 @karthikalle @erikyangs for review !

PR #151, opened 2020-08-14
added volatile to addDocumentsMaxBufferLen

erikyangs:


umeshdangat:
seems like we don't need to see the most recent write before every  indexing batch so probably okay to leave the variable non-volatile !

PR #152, opened 2020-08-28
Add server tests jar to be used in plugins

karthikalle:
The `ServerTestCase` will be used in tests for the `maptype` plugin.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,7 +20,7 @@ sourceCompatibility = 1.14
 targetCompatibility = 1.14
 
 allprojects {
-    version = '0.1.5'
+    version = '0.1.6'
     group = 'com.yelp.nrtsearch'
 }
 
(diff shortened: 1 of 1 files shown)


aprudhomme:
You might want to add in some changes to ServerTestCase from my other branch https://github.com/Yelp/nrtsearch/pull/137/files#diff-e3c1ea76185caf8c4054d0d80465a0c3
Mainly for the hacky fix to let the grpc cleanup rule run multiple times, otherwise you can only have one subclass.

umeshdangat:
@aprudhomme @karthikalle  regarding having multiple subclasses for ServerTestCase I thought I already fixed that?  In master we already have [multiple classes extending ServerTestCase](https://github.com/Yelp/nrtsearch/search?q=%22extends+ServerTestCase%22&unscoped_q=%22extends+ServerTestCase%22)

aprudhomme:
Yeah, I guess redefining the rule in each class will work as well.

aprudhomme:
Actually, wouldn't the rule instance passed into the GrpcServer instance be the one from the base class? Redefining the rule would keep you from having an error, but I don't think the rule will function correctly. I think the subclass will need pass the rule to the base class during initialization.

PR #155, opened 2020-09-10
Add support for updateDocuments if docId is true

### LINKED ISSUE #153: Use IndexWriter.updateDocuments() in AddDocumentHandler
Currently we don't provide an API to update documents in NRTSearch, but instead have the addDocuments request. It is up to the clients to send delete and add documents externally, and has the following problems:

1. Could lead to inconsistencies where refreshes can happen between these requests.
2. If there are two add documents requests, these would end up as two separate documents than being updated.

It would be good to have the updates happen atomically in a single request. We could use the [updateDocuments()](https://lucene.apache.org/core/8_4_0/core/org/apache/lucene/index/IndexWriter.html#updateDocuments-org.apache.lucene.index.Term-java.lang.Iterable-) from lucene for this.

1. Change [addDocuments in AddDocumentsHandler](https://github.com/Yelp/nrtsearch/blob/97cb292633a70b37d8c8520ef520131b259f2cb0/src/main/java/com/yelp/nrtsearch/server/luceneserver/AddDocumentHandler.java#L140) to use updateDocuments()
2. Add a notion of `doc_id` to FieldDefs which will be used with the TermNode for deletion.

karthikalle:
Fixes #153


umeshdangat (approved):

sarthakn7 (commented):
Can a field of any type be the `docId` ? If not, it would be good to add a check for the allowed fields and to also specify the allowed fields with the `docId` description in the proto

src/test/resources/registerFieldsBasicWithDocId.json
```diff
@@ -0,0 +1,134 @@
+{
+  "indexName": "test_index",
+  "field": [
+    {
+      "name": "doc_id",
+      "type": "ATOM",
+      "storeDocValues": true,
+      "docId": true
```
aprudhomme, line 8: Since the updateDocument method uses a Term, I think it is a requirement for the field to be indexed. Because of the wonky state of ATOM, it will always be indexed, but that is not true for all field types.
I am also a bit unsure how this works with non keyword fields. It would be good to have tests for other supported field types.

aprudhomme:
If allowing any indexable field to be a primary term is too broad, this could maybe be done by adding a field property https://github.com/Yelp/nrtsearch/tree/master/src/main/java/com/yelp/nrtsearch/server/luceneserver/field/properties

This interface could provide methods to populate the doc_id field on a lucene Document, and to create the update Term from a lucene Document.

src/test/resources/registerFieldsWithFalseDocValues.json
```diff
@@ -0,0 +1,24 @@
+{
+  "indexName": "test_index",
+  "field": [
+    {
+      "name": "doc_id",
+      "type": "INT",
+      "docId": true,
+      "multiValued": true
```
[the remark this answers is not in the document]
karthikalle: Good catch! I am changing the tests to not rely on these json files and also check for the error message to make sure. Will send these in an updated review.

src/main/java/com/yelp/nrtsearch/server/luceneserver/AddDocumentHandler.java
```diff
@@ -191,6 +158,56 @@ public Document next() {
...
     }
 
+    private void updateDocuments(
+        Queue<Document> documents, IndexState indexState, ShardState shardState)
+        throws IOException {
+      String docIdFieldName = indexState.getDocIdField().getName();
+      for (Document nextDoc : documents) {
+        nextDoc = handleFacets(shardState, nextDoc);
+        shardState.writer.updateDocument(
+            new Term(docIdFieldName, nextDoc.getBinaryValue(docIdFieldName)), nextDoc);
```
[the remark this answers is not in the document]
karthikalle: What do we want the desired behavior if it is null? Should we fail indexing request for just that document or for the complete request?

src/main/java/com/yelp/nrtsearch/server/luceneserver/RegisterFieldsHandler.java
```diff
@@ -59,6 +59,7 @@ public FieldDefResponse handle(final IndexState indexState, FieldDefRequest fiel
     final Map<String, FieldDef> pendingFieldDefs = new HashMap<>();
     final Map<String, String> saveStates = new HashMap<>();
     Set<String> seen = new HashSet<>();
+    boolean docIdFieldFound = false;
```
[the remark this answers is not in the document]
karthikalle: That's fair, I made some changes which will avoid checking for this and rely on not setting the docIdField in indexState.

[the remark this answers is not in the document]
karthikalle: Actually, I will leave the check as it is and use it before adding fields to the index state.

src/main/java/com/yelp/nrtsearch/server/luceneserver/RegisterFieldsHandler.java
```diff
@@ -169,6 +178,10 @@ private FieldDef parseOneFieldType(
                 : currentField.getFacetIndexFieldName();
         indexState.facetsConfig.setIndexFieldName(fieldName, facetFieldName);
       }
+      // Register DocIdField
+      if (indexableFieldDef.isDocId()) {
+        indexState.setDocIdField(indexableFieldDef);
```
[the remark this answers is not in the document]
karthikalle: ++ for catching that. I have moved adding the docIdfield to where we add fields to the indexState. That way all validations will be finished before we add fields.

src/test/resources/registerFieldsBasicWithDocId.json
karthikalle: I added tests for which field types would be supported and make sure exceptions are thrown for those which don't.

karthikalle:
> If allowing any indexable field to be a primary term is too broad, this could maybe be done by adding a field property https://github.com/Yelp/nrtsearch/tree/master/src/main/java/com/yelp/nrtsearch/server/luceneserver/field/properties
> 
> This interface could provide methods to populate the doc_id field on a lucene Document, and to create the update Term from a lucene Document.

I think it makes sense to have the Key capability only on a few fields (fields like boolean, text, lat-long may not be useful as keys).

I have added a new property to use it only for ATOM, INT and LONG. We could add it to other fields if needed as well. 

One other thing I had to do is set the search properties in IndexableFieldDef as we always need to set it `IndexOptions.DOCS`.

karthikalle:
> Can a field of any type be the `docId` ? If not, it would be good to add a check for the allowed fields and to also specify the allowed fields with the `docId` description in the proto

Added a Keyable interface to only support it for a few fields.

src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java
```diff
@@ -133,6 +134,7 @@
 
   private static final Pattern reSimpleName = Pattern.compile("^[a-zA-Z_][a-zA-Z_0-9]*$");
   private ThreadPoolExecutor searchThreadPoolExecutor;
+  private Keyable keyableField;
```
aprudhomme, line 137: It might be better track this as a FieldDef, and verify it is instanceof Keyable when set. Then you will have access to the field name and any other properties.

src/main/java/com/yelp/nrtsearch/server/luceneserver/RegisterFieldsHandler.java
```diff
@@ -111,16 +121,28 @@ public FieldDefResponse handle(final IndexState indexState, FieldDefRequest fiel
...
           jsonParser.parse(saveStates.get(ent.getKey())).getAsJsonObject();
-      indexState.addField(ent.getValue(), fieldAsJsonObject);
+      FieldDef fieldDef = ent.getValue();
+      indexState.addField(fieldDef, fieldAsJsonObject);
+      if (fieldDef instanceof Keyable) {
+        Keyable keyableField = (Keyable) fieldDef;
+        if (keyableField.isKey()) {
+          indexState.setKeyableField(keyableField);
+        }
+      }
```
aprudhomme, line 145: This could also be done by holding a reference to the key FieldDef when building it above, since there will be at most one field.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/IndexableFieldDef.java
```diff
@@ -118,6 +120,20 @@ protected void validateRequest(Field requestField) {
               "field: %s cannot have both group and multivalued set to true. Cannot  group on multiValued fields",
               requestField.getName()));
     }
+    if (requestField.getKey()) {
+      if (requestField.getMultiValued()) {
```
aprudhomme, line 124: Could probably also verify that this is Keyable here as well.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/IndexableFieldDef.java
```diff
@@ -157,7 +173,13 @@ protected FacetValueType parseFacetValueType(Field requestField) {
...
    * @param requestField field from request
    */
-  protected void setSearchProperties(FieldType fieldType, Field requestField) {}
+  protected void setSearchProperties(FieldType fieldType, Field requestField) {
+    if (this instanceof Keyable) {
+      fieldType.setIndexOptions(IndexOptions.DOCS);
+      fieldType.setTokenized(false);
+      fieldType.setOmitNorms(true);
+    }
+  }
```
aprudhomme, line 182: I'm not sure this is the best place to put this. The subclasses could change these values without realizing it. Also, this is only needed for the field instance that is actually the key. I would say move this logic into the subclass implementations.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/IntFieldDef.java
```diff
@@ -32,10 +33,13 @@
 import org.apache.lucene.search.SortField;
 
 /** Field class for 'INT' field type. */
-public class IntFieldDef extends NumberFieldDef {
+public class IntFieldDef extends NumberFieldDef implements Keyable {
```
aprudhomme, line 36: Numeric fields will only index the default FieldType if store is true. To index the key when store is false, you will need to override parseDocumentField() and add this field if isKey and not store.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/properties/Keyable.java
```diff
@@ -0,0 +1,54 @@
...
+   * @return name of the keyable field
+   */
+  String getName();
+
+  /**
+   * One FieldDef has to be registered with key=true in order to be used for updating the documents.
+   *
+   * @return whether the field is registered with key=true
+   */
+  boolean isKey();
```
aprudhomme, line 43: I'm not sure that these methods are needed on the interface. getName can be provided by the FieldDef.

isKey may not need to be exposed publicly. The index key can be identified from the IndexState. If we did want to expose it, it might be better as a property of FieldDef/IndexableFieldDef.

aprudhomme (commented):
This seems like a reasonable approach. Added some suggestions on how the Keyable interface could be minimized a bit.

src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java
karthikalle: A good reason to have this as Keyable is that I could directly use it to for `getTerm()`, instead of having to cast it again while updating the documents. Name is the only field that is needed now, but since we made the interface have the name which is anyway defined for all FieldDefs, we would get access to it directly.

If it's okay, we could change this to be a `FieldDef` later if we need access to other properties.

src/main/java/com/yelp/nrtsearch/server/luceneserver/RegisterFieldsHandler.java
karthikalle: Makes sense, I have changed the code to reflect this.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/IndexableFieldDef.java
karthikalle: Good catch! I added the check now.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/IndexableFieldDef.java
karthikalle: I have added a `setKeyableSearchProperties` to make sure that some they are set irrespective of subclass implementations. This is to make sure that the key doesn't have tokenization (like ATOM fields), and is also indexed .

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/properties/Keyable.java
karthikalle: I was able to get rid of `isKey()`. It might be okay to leave the `getName()` since it's anyway going to be retrieved from the FieldDef and would avoid having to get it from FieldDef and pass it around to Keyable instances.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/IntFieldDef.java
karthikalle: To keep things simple, I have made `store=true` as mandatory for Keyable fields. We don't really need it in docvalues, but have to guarantee indexing which is taken care of by making `store=true`.

src/main/java/com/yelp/nrtsearch/server/luceneserver/RegisterFieldsHandler.java
```diff
@@ -111,16 +113,34 @@ public FieldDefResponse handle(final IndexState indexState, FieldDefRequest fiel
...
     for (Map.Entry<String, FieldDef> ent : pendingFieldDefs.entrySet()) {
       JsonObject fieldAsJsonObject =
           jsonParser.parse(saveStates.get(ent.getKey())).getAsJsonObject();
-      indexState.addField(ent.getValue(), fieldAsJsonObject);
+      FieldDef fieldDef = ent.getValue();
+      indexState.addField(fieldDef, fieldAsJsonObject);
+    }
+    if (!indexState.hasKeyableField()) {
+      indexState.setKeyableField(keyableField);
     }
```
aprudhomme, line 144: Should keyableField be checked for null too? It looks like passing null to setKeyableField is a noop, but might be better to avoid the call entirely.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/IndexableFieldDef.java
```diff
@@ -118,6 +121,27 @@ protected void validateRequest(Field requestField) {
...
+      throw new IllegalArgumentException(
+          String.format("field: \"%s\" is not a keyable field type", requestField.getName()));
+    }
+    if (requestField.getMultiValued()) {
+      throw new IllegalArgumentException(
+          String.format(
+              "field: %s cannot have both multivalued and docId set to true. Only single docId is supported",
+              requestField.getName()));
+    }
+    if (!requestField.getStore()) {
```
aprudhomme, line 140: I hope we can remove this requirement at some point.

src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java
aprudhomme: The idea was to allow access to the name without creating a duplicate in the interface. This is fine for now.

I'm thinking this could maybe be done better by giving field properties a common interface to get their FieldDef, but that is future work.

aprudhomme (approved):
Looks good to me

karthikalle:
Closing in favor or #159

PR #157, opened 2020-09-12
Added tests to verify searcher pruning and segment file deletion after merges

sarthakn7:
There are three tests added and each test takes just over a minute to complete since we have to wait for a minute for the searcher to get pruned. I think this is fine as devs would mostly run only specific tests via an IDE but we can skip these tests later if needed.

### DIFF
--- a/src/test/java/com/yelp/nrtsearch/server/grpc/MergeBehaviorTests.java
+++ b/src/test/java/com/yelp/nrtsearch/server/grpc/MergeBehaviorTests.java
@@ -1,0 +1,405 @@
+/*
+ * Copyright 2020 Yelp Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


src/test/java/com/yelp/nrtsearch/server/grpc/MergeBehaviorTests.java
```diff
@@ -0,0 +1,405 @@
...
+/**
+ * These tests verify the behavior of segment merges on the searchers and on the segment files. For
+ * old searchers to be pruned: 1. 60 seconds must pass after the searcher stopped being the live
+ * searcher 2. if opened with a snapshot, the snapshot must be released
+ *
+ * <p>For old segment files (which have been merged into new segments) to be deleted: 1. Any
+ * searchers referencing the segments must be pruned 2. If the segments were committed, a new commit
+ * after the merge must be issued
+ */
+public class MergeBehaviorTests {
```
umeshdangat, line 51: If these tests are "slow" could we do something similar to what we did with `YelpReviewsTest`
https://github.com/Yelp/nrtsearch/blob/master/build.gradle#L127

We typically dont run these by default when one invokes: "./gradlew test" but we do run them on travis and if they fail the build fails on travis:
https://github.com/Yelp/nrtsearch/blob/master/.travis.yml#L14

umeshdangat (approved):

PR #159, opened 2020-09-15
Add a new _ID field for updating documents

### LINKED ISSUE #153: Use IndexWriter.updateDocuments() in AddDocumentHandler

karthikalle:
This is another way to fix #153 and seemed better than #155 

We can declare the fields this way and that will be used to update the documents:

```
{
      "name": "doc_id",
      "type": "_ID",
      "store": true
}
``` 
Few reasons why this seems cleaner :

1. The changes are smaller and confined to the field def. We could add more code to the field def without breaking it in any other place.
2. Better notation of '_ID' which is always going to be stored as a string. This won't overload other field definitions to just update the documents.

### DIFF
--- a/clientlib/src/main/proto/yelp/nrtsearch/luceneserver.proto
+++ b/clientlib/src/main/proto/yelp/nrtsearch/luceneserver.proto
@@ -341,7 +341,7 @@ enum FieldType {
     // TODO need tests for internal:
     INTERNAL = 10; //Internal field, currently only for holding indexed facets data.
     CUSTOM = 11; // Field type specified by name.
-    _ID = 12;
+    _ID = 12; // Field which will be used as document IDs
 }
 
 //How the tokens should be indexed.
--- a/clientlib/src/main/proto/yelp/nrtsearch/luceneserver.proto
+++ b/clientlib/src/main/proto/yelp/nrtsearch/luceneserver.proto
@@ -341,6 +341,7 @@ enum FieldType {
     // TODO need tests for internal:
     INTERNAL = 10; //Internal field, currently only for holding indexed facets data.
     CUSTOM = 11; // Field type specified by name.
+    _ID = 12;
 }
 
 //How the tokens should be indexed.
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java
@@ -134,6 +134,7 @@ public class IndexState implements Closeable, Restorable {
 
   private static final Pattern reSimpleName = Pattern.compile("^[a-zA-Z_][a-zA-Z_0-9]*$");
   private ThreadPoolExecutor searchThreadPoolExecutor;
+  private IdFieldDef idFieldDef = null;
 
   public ShardState addShard(int shardOrd, boolean doCreate) {
     if (shards.containsKey(shardOrd)) {
@@ -228,12 +229,7 @@ public class IndexState implements Closeable, Restorable {
   }
 
   public IdFieldDef getIdFieldDef() {
-    for (Map.Entry<String, FieldDef> entry : fields.entrySet()) {
-      if (entry.getValue() instanceof IdFieldDef) {
-        return (IdFieldDef) entry.getValue();
-      }
-    }
-    return null;
+    return idFieldDef;
   }
 
   /** Tracks snapshot references to generations. */
@@ -792,6 +788,9 @@ public class IndexState implements Closeable, Restorable {
           && facetValueType != IndexableFieldDef.FacetValueType.NUMERIC_RANGE) {
         internalFacetFieldNames.add(facetsConfig.getDimConfig(fd.getName()).indexFieldName);
       }
+    }
+    if (fd instanceof IdFieldDef) {
+      idFieldDef = (IdFieldDef) fd;
     }
   }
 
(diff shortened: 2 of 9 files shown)


clientlib/src/main/proto/yelp/nrtsearch/luceneserver.proto
```diff
@@ -341,6 +341,7 @@ enum FieldType {
     // TODO need tests for internal:
     INTERNAL = 10; //Internal field, currently only for holding indexed facets data.
     CUSTOM = 11; // Field type specified by name.
+    _ID = 12;
```
aprudhomme, line 344: nit: comment

src/main/java/com/yelp/nrtsearch/server/luceneserver/RegisterFieldsHandler.java
```diff
@@ -127,6 +132,18 @@ public FieldDefResponse handle(final IndexState indexState, FieldDefRequest fiel
     return reply;
   }
 
+  private void verifyOnlyOneIdFieldExists(Map<String, FieldDef> pendingFieldDefs, FieldDef fieldDef)
```
aprudhomme, line 135: This will also need to make sure there is not an _id field already registered in the index state.

src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java
```diff
@@ -226,6 +227,15 @@ public ThreadPoolExecutor getSearchThreadPoolExecutor() {
     return searchThreadPoolExecutor;
   }
 
+  public IdFieldDef getIdFieldDef() {
+    for (Map.Entry<String, FieldDef> entry : fields.entrySet()) {
+      if (entry.getValue() instanceof IdFieldDef) {
+        return (IdFieldDef) entry.getValue();
+      }
+    }
```
aprudhomme, line 235: It might be nice to avoid this repeated field searching. Could maybe have IndexState.addField() store the IdFieldDef when it is added?

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/IdFieldDef.java
```diff
@@ -0,0 +1,145 @@
...
+      throw new IllegalArgumentException("Cannot index multiple values into _id fields");
+    }
+    String fieldStr = fieldValues.get(0);
+    if (hasDocValues()) {
+      BytesRef stringBytes = new BytesRef(fieldStr);
+      document.add(new BinaryDocValuesField(getName(), stringBytes));
+    }
+    if (isStored()) {
+      document.add(new FieldWithData(getName(), fieldType, fieldStr));
+    }
```
aprudhomme, line 93: I don't think it is required for the value to be stored. Stored is a property of FieldType and will not affect the other indexing options.
You should be able to always add FieldWithData to the document. Even if stored was false in the field message, the value should still be indexed. I think indexing is enough to make updateDocument work.

That being said, it would probably be good to force the field to be retrievable somehow (doc value or stored).

aprudhomme (commented):
I like this way better.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/IdFieldDef.java
karthikalle: That's true. I changed the check to be either store=true or docvalue=true in `validateRequest`

src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java
```diff
@@ -133,6 +134,7 @@
 
   private static final Pattern reSimpleName = Pattern.compile("^[a-zA-Z_][a-zA-Z_0-9]*$");
   private ThreadPoolExecutor searchThreadPoolExecutor;
+  private IdFieldDef idFieldDef = null;
```
aprudhomme, line 137: This may need to be declared volatile (or AtomicReference) since getIdFieldDef is unsynchronized, and threads could technically miss the update. I don't think there will be a big performance hit, since there will be no writes in stead state.

I'm not sure if this would be a problem in practice. If this seems like overkill, let me know.

aprudhomme (approved):

src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java
karthikalle: Yeah, I don't think that should be a problem as the idFieldDef change is only going to happen during startup or adding new fields. It will be a while before it's actually used via new indexing requests. If there are incoming requests which don't have the idField values, it will break as expected.

aprudhomme: As long as it doesn't get into the thread cache somehow before being set, that should be ok.

karthikalle: Since there is one index state per-process per-index and we create new callables while adding new documents, I think it should be okay for now.

umeshdangat (approved):

Issue #158, opened 2020-09-14
Add custom similarity support


aprudhomme:
Currently, the only supported lucene similarity implementations for a field are classic and BM25. We should create a plugin interface to allow registration of custom similarity implementations.

PR #160, opened 2020-09-15
Add custom similarity support

aprudhomme:
Issue https://github.com/Yelp/nrtsearch/issues/158

Adds support for custom similarity implementations. Fields may specify one of the built in types (BM25 (default), classic, boolean) or a Similarity registered by a SimilarityPlugin.


src/main/java/com/yelp/nrtsearch/server/luceneserver/field/IndexableFieldDef.java
```diff
@@ -86,21 +86,11 @@ protected IndexableFieldDef(String name, Field requestField) {
...
-        // TODO support CustomSimilarity
-        assert false;
-        this.similarity = null;
-      }
-    } else {
-      this.similarity = DEFAULT_SIMILARITY;
-    }
+    Map<String, Object> similarityParams =
+        Maps.transformValues(
+            requestField.getSimilarityParams().getFieldsMap(), StructValueTransformer.INSTANCE);
```
sarthakn7, line 91: It looks like we have a util class for this, should we use it instead ?
https://github.com/Yelp/nrtsearch/blob/master/src/main/java/com/yelp/nrtsearch/server/utils/StructValueTransformer.java#L44

sarthakn7 (commented):
Looks good, left a small comment. Also might have to regenerate grpc-gateway after merging #154 .

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/IndexableFieldDef.java
aprudhomme: Yep, forgot I added that

sarthakn7 (approved):

PR #161, opened 2020-09-22
Bump to v0.1.8

aprudhomme:
Bump server version to expose similarity plugin.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,7 +20,7 @@ sourceCompatibility = 1.14
 targetCompatibility = 1.14
 
 allprojects {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -20,7 +20,7 @@ sourceCompatibility = 1.14
 targetCompatibility = 1.14
 
 allprojects {
-    version = '0.1.7'
+    version = '0.1.8'
```
sarthakn7, line 23: should this be a bump of minor version instead of patch since there is new functionality ?

sarthakn7 (approved):

PR #164, opened 2020-10-05
Make searcher pruning thread run on replicas

aprudhomme:
The current version of the searcher pruning thread exits after the first iteration if the index writer is null. Since this is always true on the replica nodes, there is no searcher pruning.

This branch removes the null check and instead make the thread Closable. The thread is closed when the ShardState closes or if there is a start error.


umeshdangat (approved):
lg2m

sarthakn7 (approved):

Issue #166, opened 2020-10-07
Support multi-valued latlon fields


aprudhomme:
LatLon fields do not currently support storing multiple values. We should make this option available.

PR #169, opened 2020-10-08
Add multi-value support to LAT_LON field type

aprudhomme:
Extend LAT_LON field type to support multiple point values. Points are indexed by specifying a sequence of lat lon values.

Script and Query tests updated to verify a multi-value point field.

https://github.com/Yelp/nrtsearch/issues/166


umeshdangat (approved):
lg2m

PR #170, opened 2020-10-08
Add multi value field support and length/empty properties

aprudhomme:
Adds support for multi valued fields in expression bindings. The default value used for these fields will be the first (minimum) value.

Fields may now be specified in the form "doc['field_name'].property"
For numeric fields the 'property' value may be:
- value : field value
- length : number of values in this field
- empty : 1 if the field has no values, 0 otherwise


umeshdangat (approved):
lgtm

PR #176, opened 2020-10-15
Add endpoint to clean up old index backups

jedrazb:
# Change summary

Added endpoint `delete_index_backup` to clean up old nrtsearch indexes backups. The endpoint can be called on a regular basis to keep only backups from the last `n` days.  Changes include:
* Added 2 methods to `Archiver`:  `getVersionedResource` which gets all the version hashes for a particular resource, `deleteVersion` takes  a resource and a version hash and deletes that version
* Defined rpc for `delete_index_backup` endpoint
* Created `DeleteIndexBackupHandler` which decides what versions should be deleted basing on `nDays` parameter in the request

# Testing

## Manual testing procedure

Start (and restore) an index

```
curl -XPOST localhost:10003/v1/start_index -d '{
  "indexName" : "talk_v1",
   "mode": "PRIMARY",
   "restore" : {
     "serviceName" : "nrtsearch-generic",
     "resourceName": "talk_v1"
   }
}'
```

Create a few backups

```
curl -XPOST localhost:10003/v1/backup_index -d '{
	"indexName": "talk_v1", 
	"serviceName": "nrtsearch-generic-jblaszyk",
	"resourceName": "talk_v1"
}'
```

Call the new endpoint. Note: for the sake of **manual testing** I **changed days to minutes**. I was deleting versions older than 5 mins. 

```
curl -XPOST localhost:10003/v1/delete_index_backup -d '{
	"indexName": "talk_v1", 
	"serviceName": "nrtsearch-generic-jblaszyk", 
	"resourceName": "talk_v1", 
	"nDays": 5
}'
```

Manually verified on s3 that backups which should be deleted were deleted. 🎉

## Unit testing

Added unit tests for `Archiver` methods, and the `DeleteBackupIndexHandler`. 

Unit tests are ✅.


src/main/java/com/yelp/nrtsearch/server/luceneserver/DeleteIndexBackupHandler.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.ArrayList;
+import java.util.Calendar;
+import java.util.Date;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DeleteIndexBackupHandler
```
sarthakn7, line 32: You can add a test for this class as well. It would involve creating a test object of this class by passing a mocked `Archiver` object and then testing the `handle` method (and optionally `olderThanNDays` - but you should be able to test it by testing just the handle method)

src/main/java/com/yelp/nrtsearch/server/utils/VersionedResourceObject.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.Date;
+
+public class VersionedResourceObject {
+
+  private final String serviceName;
+  private final String resourceName;
+  private final String versionHash;
+  private final Date creationTimestamp;
+
+  public VersionedResourceObject(
```
sarthakn7, line 27: Can implement `toString` method at least, if not `equals` and `hashcode` as well. These can be autogenerated by IDEs.

sarthakn7 (commented):
Feature-wise looks good. I see that some of my comments refer to existing code that you used, but can always do things in a better way :)

src/main/java/com/yelp/nrtsearch/server/luceneserver/DeleteIndexBackupHandler.java
```diff
@@ -0,0 +1,104 @@
...
+              protoRequest.getResourceName(),
+              protoRequest.getNDays()),
+          e);
+      return deleteIndexBackupResponseBuilder.build();
+    }
+
+    return deleteIndexBackupResponseBuilder.build();
+  }
+
+  public static boolean olderThanNDays(
```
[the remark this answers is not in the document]
jedrazb: I do have tests for it so I'll keep it `public`

src/main/java/com/yelp/nrtsearch/server/luceneserver/DeleteIndexBackupHandler.java
jedrazb: I added the test of `DeleteIndexBackupHandler` with mocked `Archiver`. In that test I only verify that a single one old resource is deleted, appropriate methods are called and the correct response is returned. Also I have a separate test for `isOlderThanNDays` function. 

In order to make a better test of handler I would need to mock the return value of `Instant.now()` called inside the handler. Basing on `Instant.now()` the handler decides which resources are older than `n` days. As far as I know it's possible to mock `Instant.now()` with `PowerMock` library. Are you ok with adding this to the dependency list for this project? Or do you know better ways of mocking system time with Mockito (I couldn't find a solution with that lib)?

src/main/java/com/yelp/nrtsearch/server/luceneserver/DeleteIndexBackupHandler.java
```diff
@@ -0,0 +1,104 @@
...
+      }
+
+      for (VersionedResourceObject objectToDelete : objectsOlderThanNDays) {
+        archiver.deleteVersion(
+            objectToDelete.getServiceName(),
+            objectToDelete.getResourceName(),
+            objectToDelete.getVersionHash());
+      }
+    } catch (IOException e) {
+      logger.error(
```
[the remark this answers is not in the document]
sarthakn7: You can put exactly `{}` as the placeholder for any variables in the message - %s, %d, etc. don't work I think

src/main/java/com/yelp/nrtsearch/server/utils/VersionManager.java
```diff
@@ -114,4 +115,19 @@ public boolean blessVersion(String serviceName, String resourceName, String reso
 
     return true;
   }
+
+  /* Deletes a particular version of a resource. */
+  public boolean deleteVersion(String serviceName, String resourceName, String resourceHash) {
+    final String resourceKey = String.format("%s/%s/%s", serviceName, resourceName, resourceHash);
+    if (!s3.doesObjectExist(bucketName, resourceKey)) {
+      logger.error(
```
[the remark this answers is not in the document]
sarthakn7: Same as the previous one - you can put `{}` as the placeholder for variables

src/main/java/com/yelp/nrtsearch/server/luceneserver/DeleteIndexBackupHandler.java
jedrazb: Instead of mocking system clock, the tests set the creation timestamps of object in a following fashion 
```Instant date1 = Instant.now().minus(20, ChronoUnit.DAYS);```

sarthakn7 (approved):

PR #178, opened 2020-10-15
Add geo_bounding_box query for LatLonFields

### LINKED ISSUE #167: Add geobox query
We already index point data as LatLonPoint. Let's add support for geo box queries.

karthikalle:
Fixes #167

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
@@ -21,6 +21,7 @@ import com.yelp.nrtsearch.server.grpc.*;
 import com.yelp.nrtsearch.server.luceneserver.analysis.AnalyzerCreator;
 import com.yelp.nrtsearch.server.luceneserver.field.FieldDef;
 import com.yelp.nrtsearch.server.luceneserver.field.LatLonFieldDef;
+import com.yelp.nrtsearch.server.luceneserver.field.properties.GeoQueryable;
 import com.yelp.nrtsearch.server.luceneserver.field.properties.RangeQueryable;
 import com.yelp.nrtsearch.server.luceneserver.field.properties.TermQueryable;
 import com.yelp.nrtsearch.server.luceneserver.script.ScoreScript;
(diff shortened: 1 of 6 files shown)


src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
```diff
@@ -275,6 +278,18 @@ private Query getRangeQuery(RangeQuery rangeQuery, IndexState state) {
...
   }
 
+  private Query getGeoBoundingBoxQuery(GeoBoundingBoxQuery geoBoundingBoxQuery, IndexState state) {
+    String fieldName = geoBoundingBoxQuery.getField();
+    FieldDef field = state.getField(fieldName);
+
+    if (!(field instanceof LatLonFieldDef)) {
+      throw new IllegalArgumentException(
+          "Field: " + fieldName + " does not support GeoBoundingBoxQuery");
+    }
```
aprudhomme, line 288: Could we maybe do this by creating a GeoQueryable field property instead. With the pluggable nature of field types, it is better to not tie ourselves to specific class implementations. Also, I imagine we will want to add geo distance and polygon query support in the future.

karthikalle: I thought about doing it initially but thought it might be early to add it. But it does make it more cleaner.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/LatLonFieldDef.java
```diff
@@ -122,4 +125,14 @@ public SortField getSortField(SortType type) {
     return LatLonDocValuesField.newDistanceSort(
         getName(), origin.getLatitude(), origin.getLongitude());
   }
+
+  @Override
+  public Query getGeoBoundingBoxQuery(GeoBoundingBoxQuery geoBoundingBoxQuery) {
```
aprudhomme, line 130: nit: can this throw an exception if the field is not searchable.

aprudhomme (approved):

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/LatLonFieldDef.java
karthikalle: Done!

PR #179, opened 2020-10-19
Implement CompletionInfixSuggester

jinxihexi0411:
CompletionInfixSuggester is implemented for suggestion infix matching. This suggester is based on the implementation details in `AnalyzingInfixSuggester`. Suggestions are looked up by wfst algorithm and context is used to filter out unexpected documents. 

- Index is created by build method, where the `InputIterator` should be `FromFileSuggestItemIterator` type, where `text`, set of `search text`, set of `context`, and `payload` can be parsed.
- Each Document includes two stored field (`text` field and `payload` field) and multiple `ContextSuggestField`
- `lookup` method is supported by `SuggestIndexSearcher`, where `nrtsuggester` is used under the hood.
- codec is customized. `Completion84PostingsFormat` is used to index suggest_text field.

In the next PR, a `FuzzyInfixSuggester` will be implemented by extending `CompletionInfixSuggester`.

Test done:

test cases pass

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/suggest/CompletionInfixSuggester.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/suggest/CompletionInfixSuggester.java
@@ -46,26 +46,27 @@ import org.apache.lucene.search.suggest.document.TopSuggestDocs;
 import org.apache.lucene.store.Directory;
 import org.apache.lucene.util.BytesRef;
 
-/** Analyzes the input text and suggests matched suggest items
- * based on the prefix matches to any pre-analyzed suffix-token gram in the indexed text.
+/**
+ * Analyzes the input text and suggests matched suggest items based on the prefix matches to any
+ * pre-analyzed suffix-token gram in the indexed text.
  *
- * <p> This suggester relies on a customized InputIterator FromFileSuggestItemIterator to build suggestion index.
- * Each suggest text is pre-analyzed as a list of suffix-token grams.
+ * <p>This suggester relies on a customized InputIterator FromFileSuggestItemIterator to build
+ * suggestion index. Each suggest text is pre-analyzed as a list of suffix-token grams.
  *
- * <p> This suggester requires payload defined for each suggest item. The payload persist necessary information
- * for the downstream services.
+ * <p>This suggester requires payload defined for each suggest item. The payload persist necessary
+ * information for the downstream services.
  *
- * <p> This suggester supports multiple contexts. The suggested results are returned only when any of its contexts
- * matches any context in the lookup query. Context match is not considered if there is no context specified in the
- * lookup query.
+ * <p>This suggester supports multiple contexts. The suggested results are returned only when any of
+ * its contexts matches any context in the lookup query. Context match is not considered if there is
+ * no context specified in the lookup query.
  *
- * <p> The same as AnalyzingInfixSuggester, results are sorted by descending weight values.
+ * <p>The same as AnalyzingInfixSuggester, results are sorted by descending weight values.
  *
- * Example usage of this suggester:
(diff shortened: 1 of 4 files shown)


src/main/java/com/yelp/nrtsearch/server/luceneserver/suggest/CompletionInfixSuggester.java
```diff
@@ -0,0 +1,291 @@
...
+import org.apache.lucene.search.suggest.document.CompletionQuery;
+import org.apache.lucene.search.suggest.document.ContextQuery;
+import org.apache.lucene.search.suggest.document.ContextSuggestField;
+import org.apache.lucene.search.suggest.document.PrefixCompletionQuery;
+import org.apache.lucene.search.suggest.document.SuggestIndexSearcher;
+import org.apache.lucene.search.suggest.document.TopSuggestDocs;
+import org.apache.lucene.store.Directory;
+import org.apache.lucene.util.BytesRef;
+
+public class CompletionInfixSuggester extends AnalyzingInfixSuggester {
```
umeshdangat, line 49: Maybe add a comment giving example usage. (I know we have tests explaining what this class does, but might be nice to show a quick usage via build and then lookup). Also could make this class final so as to restrict/caution against further inheritance since it works on a specific iterator that we have right now ?

src/main/java/com/yelp/nrtsearch/server/luceneserver/suggest/CompletionInfixSuggester.java
```diff
@@ -0,0 +1,291 @@
...
+      CharSequence key, Set<BytesRef> contexts, boolean onlyMorePopular, int num)
+      throws IOException {
+    ensureOpen();
+    IndexSearcher searcher;
+    SearcherManager mgr;
+    List<LookupResult> results;
+
+    synchronized (searcherMgrLock) {
+      mgr = searcherMgr; // acquire & release on same SearcherManager, via local reference
+      searcher = searcherMgr.acquire();
```
umeshdangat, line 75: if we issue `lookup` before issuing `build` this would throw a `NullPointerException` ? If so we could probably throw a more appropriate RuntimeException here instead ?

umeshdangat (commented):
Nice ! Seems pretty good to me, left minor comments.

src/main/java/com/yelp/nrtsearch/server/luceneserver/suggest/CompletionInfixSuggester.java
jinxihexi0411: ++. Added a check of `searcher manager` at start of `lookup` function. `RuntimeException` will be throw if searchermgr is null.

src/main/java/com/yelp/nrtsearch/server/luceneserver/suggest/CompletionInfixSuggester.java
jinxihexi0411: Sure thing. Detailed comment will be added.

This class can not be marked as final, since one more suggester `FuzzyInfixSuggester` will be implemented by inheriting this one for fuzzy match. For the pending suggester, the only difference is the way to construct lookup query. `FuzzyInfixSuggester` will be made to final.

umeshdangat:
@jinxihexi0411 seems like some formatting issues with spotless: https://travis-ci.com/github/Yelp/nrtsearch/builds/191106000#L16895.

jinxihexi0411:
> @jinxihexi0411 seems like some formatting issues with spotless: https://travis-ci.com/github/Yelp/nrtsearch/builds/191106000#L16895.

Yeah, did not realize that the comment can also affect format. Correct this already with spotlessApply.

umeshdangat (approved):

PR #181, opened 2020-10-20
Add support for child fields

aprudhomme:
Add support for child fields. An indexable field can have an optional list of child fields, which also must be indexable. This supports nesting.

Child fields index the same data as their parent.

Child fields can be retrieved/queried by specifying their path in dot notation.


sarthakn7 (approved):
Lgtm but you might have to merge master and build grpc-gateway

PR #186, opened 2020-10-26
move suggest proto to a separate file

jinxihexi0411:
Suggest-related defination is moved to a separate file for better readability.

In the following PRs, this proto file will be updated for adding support for CompletionInfixSuggester and FuzzyInfixSuggester

- Test passed with
```
./gradlew clean && ./gradlew installDist && ./gradlew spotlessApply && ./gradlew test
```


clientlib/src/main/proto/yelp/nrtsearch/luceneserver.proto
```diff
@@ -3,6 +3,7 @@ syntax = "proto3";
 
 import "yelp/nrtsearch/search.proto";
 import "yelp/nrtsearch/analysis.proto";
+import "yelp/nrtsearch/suggest.proto";
```
umeshdangat, line 6: @jinxihexi0411 typically when we change API in proto files (since you are just moving proto files might not be applicable here), we run this command https://github.com/Yelp/nrtsearch/blob/e2a8e9cd4af22505b459c36007696707b9938c20/build.gradle#L117 .

This allows for the generation of the new swagger spec and latest builds of rest-grpc Gateway. We use the rest-grpc gateway so that clients can also communicate with nrtsearch using REST and swagger spec allows us to use yelp auto-generated swagger cleint libs. 

So you might want to try running this command and checking if the files produced change. For example see this PR: https://github.com/Yelp/nrtsearch/pull/189/files which changes the proto files.

jinxihexi0411: After running `buildGrpcGateway` tasks. The swagger json file remains the same, but `grpc-gateway/luceneserver.pb.go` got updated. 

I posted a new commit based on ^

umeshdangat (approved):

PR #187, opened 2020-10-26
Build search request into SearchContext

aprudhomme:
An extracted piece of my larger attempt to refactor search logic and simplify SearchHandler https://github.com/Yelp/nrtsearch/pull/137

All processing of the SearchRequest message is offloaded to the SearchRequestProcessor, which produces a SearchContext containing all the information needed to execute the query.

Having a single object for search info will allow for cleaner interfaces, such as a custom fetch operation for score_components. It may also be a helpful start for making the search path changes needed for nested fields.

Other changes
- Sort parsing has been separated into a static class
- Sort values are now filled based on the data type, instead of toString
- Search collectors for the primary query ranking (relevance/sort/large hits) are abstracted by a DocCollector interface


umeshdangat (approved):
lg2m

PR #189, opened 2020-10-27
Add facet to sample doc values from the top ranked docs

aprudhomme:
Poc for a facet that samples over the top documents. The search operation collects enough docs to satisfy the top docs and any sampler facets. The facet values are counted in a similar way to the script facet, with the values source being the field doc values.


karthikalle:
I think it is good to go with this approach. I have tried to look at custom FacetsCollector or change our DrillSidewaysImpl, but couldn't find an efficient way to preserve the order and collect only for top k documents.

src/main/java/com/yelp/nrtsearch/server/luceneserver/SearchHandler.java
```diff
@@ -965,6 +977,17 @@ else if (fd instanceof IndexableFieldDef && ((IndexableFieldDef) fd).isStored())
     return sortedFields;
   }
 
+  private int getNumHitsToCollect(int topHits, SearchRequest request) {
```
karthikalle, line 980: Would be good to add some comments on why we are retrieving more than topHits for sample docs.

karthikalle (approved):

PR #190, opened 2020-10-29
Added suggest.proto to grpc-gateway generation

sarthakn7:
suggest.proto was added in #186 , adding it to grpc-gateway generation as well. Also removed unused import from suggest.proto.


umeshdangat (approved):

sarthakn7:
closing in favor of #191 which has the updated generated files as well.

PR #192, opened 2020-10-30
Add Polygon type & query to nrtsearch

taoyyu:
Add polygon type and query to nrtsearch.

The data will be just geojson string (json dump of the ES format).

doc_value not allowed yet(same as ES) since we usually don't want to retrieve poly. However, it should not be too hard to support it.

suggest proto generated file diff included.


src/main/java/com/yelp/nrtsearch/server/luceneserver/field/PolygofieldDef.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.List;
+import org.apache.lucene.document.Document;
+import org.apache.lucene.document.LatLonShape;
+import org.apache.lucene.document.ShapeField;
+import org.apache.lucene.geo.Polygon;
+import org.apache.lucene.index.DocValuesType;
+import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.search.Query;
+
+public class PolygofieldDef extends IndexableFieldDef implements PolygonQueryable {
```
sarthakn7, line 36: Nit: typo, name should probably be `PolygonFieldDef`

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/PolygofieldDef.java
```diff
@@ -0,0 +1,127 @@
...
+  public LoadedDocValues<?> getDocValues(LeafReaderContext context) throws IOException {
+    // TODO: figure out later
+    return null;
+  }
+
+  @Override
+  public Query getGeoPolygoQuery(GeoPolygonQuery geoPolygonQuery) {
+    return LatLonShape.newBoxQuery(
+        geoPolygonQuery.getField(),
+        ShapeField.QueryRelation.CONTAINS,
```
sarthakn7, line 121: Should we let users specify the `QueryRelation` for this query ?

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/PolygofieldDef.java
```diff
@@ -0,0 +1,127 @@
...
+    if (requestField.getHighlight()) {
+      throw new IllegalArgumentException(
+          String.format(
+              "field: %s cannot have highlight=true. only type=text or type=atom fields can have highlight=true",
+              getName()));
+    }
+
+    if (requestField.getStore()) {
+      throw new IllegalArgumentException("polygon fields cannot be stored");
+    }
```
sarthakn7, line 55: I don't understand this, aren't the polygons always going to be stored fields only ?

src/test/java/com/yelp/nrtsearch/server/luceneserver/field/PolygonFieldDefTest.java
```diff
@@ -0,0 +1,110 @@
...
+import com.yelp.nrtsearch.server.grpc.*;
+import com.yelp.nrtsearch.server.grpc.AddDocumentRequest.MultiValuedField;
+import com.yelp.nrtsearch.server.luceneserver.ServerTestCase;
+import io.grpc.testing.GrpcCleanupRule;
+import java.io.IOException;
+import java.util.*;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+public class PolygonFieldDefTest extends ServerTestCase {
```
sarthakn7, line 32: Should we also add a test to retrieve a value stored in the field ?

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/PolygofieldDef.java
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  @Override
+  public LoadedDocValues<?> getDocValues(LeafReaderContext context) throws IOException {
+    // TODO: figure out later
+    return null;
+  }
+
+  @Override
+  public Query getGeoPolygoQuery(GeoPolygonQuery geoPolygonQuery) {
```
sarthakn7, line 118: Should we call this query `BoxQuery` to match the Lucene equivalent ? Otherwise the query names would be confusing if we add a `PolygonQuery` later. The [geo-polygon query in ES](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-geo-polygon-query.html) seems to do a `CONTAINS` with a polygon and not a bounding box too, not sure what exactly we need.

taoyyu: the box is just the point. Here the polygon query is whether the query point is inside the indexed polygon. In lucene 8.4, we can only use the box to represent the point. In lucene 8.5, there is a dedicated method to for point. I can rename it to geoshape query to match ES. Is this ok?

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/PolygofieldDef.java
taoyyu: This is same as ES https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/index/mapper/GeoShapeFieldMapper.java#L188.
In ES polygon does not support stored fields or doc value.(Only able to retrieve from _source)

Also we never retrieve this field in use cases and this field is pretty large. So I decide to make it not retrievable for now. I can change this later in a future pr

src/test/java/com/yelp/nrtsearch/server/luceneserver/field/PolygonFieldDefTest.java
taoyyu: same as above

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/PolygofieldDef.java
taoyyu: We only have polygon contains point use case. I think it is enough to make it simple like this

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/PolygofieldDef.java
sarthakn7: I see that we don't actually add [StoredField](https://lucene.apache.org/core/8_4_0/core/org/apache/lucene/document/StoredField.html) for any type. Can leave it as is.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/PolygofieldDef.java
sarthakn7: The geoshape query also seems to do something different, how about `PointQuery` (or `GeoPointQuery`) so that it will match with Lucene 8.5 ?

sarthakn7 (approved):
Lgtm, would need to merge master and re-generate grpc-gateway

sarthakn7 (approved):

PR #195, opened 2020-11-06
Implement a new iterator to read suggest item file based on protobuf

jinxihexi0411:
In this PR, a new iterator is implemented to parse the suggest item from a protobuf-based binary file. It would be a lot easier to revise any field in the field in the future, compared to the [FromFileSuggestItemIterator](https://github.com/Yelp/nrtsearch/blob/6b2166948728e0a2c87b3c227b401781e0be38b6/src/main/java/com/yelp/nrtsearch/server/luceneserver/suggest/FromFileSuggestItemIterator.java#L38) , which replies on the adhoc logic to parse each line. 

Similarly, five fields `searchTexts`, `contexts`, `text`, `weight`, and `payload` are defined in the `NrtsearchIndex` proto schema. The batches in `search_suggest` repo also follow the same schema to generate the binary field. 

Test done:
- `make test` pass. The dummy binary index file is correctly iterated by the new iterator.


clientlib/src/main/proto/yelp/nrtsearch/suggest.proto
```diff
@@ -147,4 +147,13 @@ message FuzzyInfixSuggester {
...
   bool unicodeAware = 8; //True if all edits are measured in unicode characters, not UTF-8 bytes
+}
+
+/* The schema of suggest item to build suggest index*/
+message NrtsearchIndex {
+  int64 uniqueId = 1;
+  repeated string searchTexts = 2;
+  int64 score = 3;
+  repeated string contexts = 4;
+  bytes payload = 5;
```
sarthakn7, line 158: Is this the schema for a specific index or is this something generic ?

jinxihexi0411: This is generic for all suggest indices, which include category type, textual query type, and business type.

umeshdangat (approved):

PR #196, opened 2020-11-06
Add facet for field doc values count

aprudhomme:
Allows for a count aggregation over a field's doc values. Performed for fields that do not have any facet config.

This is intended for use with fields that have lower cardinality and do not use ordinals, such as int or long.


karthikalle (approved):
This is useful to be able to get facets on any int or long fields without having to reindex all the data.

PR #201, opened 2020-11-13
Add eager global ordinal option for SORTED_SET_DOC_VALUES facet

aprudhomme:
Add eager global ordinals option for sorted set doc values facet. Populates global ordinal cache when each new IndexSearcher is created.

This option is only active when the server is started as standalone or replica. The primary will never do eager loading.


src/main/java/com/yelp/nrtsearch/server/luceneserver/ShardState.java
```diff
@@ -426,6 +427,44 @@ public void startSearcherPruningThread(CountDownLatch shutdownNow) {
...
+      if (loadEagerOrdinals) {
+        loadEagerGlobalOrdinals(reader);
+      }
+      return searcher;
+    }
+
+    private void loadEagerGlobalOrdinals(IndexReader reader) throws IOException {
+      for (Map.Entry<String, FieldDef> entry : indexState.eagerGlobalOrdinalFields.entrySet()) {
+        // only sorted set doc values facet currently supported
+        if (entry.getValue().getFacetValueType() == FacetValueType.SORTED_SET_DOC_VALUES) {
```
umeshdangat, line 460: If `eagerGlobalOrdinals` flag is set for a field that does not have `SORTED_SET_DOC_VALUES`, should we check and disallow that at mapping creation time or at least log a warning in here?

umeshdangat (approved):
lg2m, left a minor comment.

umeshdangat (approved):

PR #202, opened 2020-11-13
FieldDefs only accept their intended type, Validation for termQuery and termInSetQuery for all FieldDefs

erikyangs:
IntFieldDef converts TextValue to IntValue in TermQuery and TermInSetQuery.

Similar to [BooleanFieldDef](https://github.com/Yelp/nrtsearch/blob/750f62d1ce6d9d6cf74276181573e5cb5d053f8c/src/main/java/com/yelp/nrtsearch/server/luceneserver/field/BooleanFieldDef.java#L150).

---
**Side note**
We may also want to add validation in the other field types for `getTermQuery`. Currently if the wrong type is inputted (e.g. a String for LongFieldDef), it uses the default value (usually `0`).


src/main/java/com/yelp/nrtsearch/server/luceneserver/field/IntFieldDef.java
```diff
@@ -107,11 +110,40 @@ public Query getRangeQuery(RangeQuery rangeQuery) {
...
+    return IntPoint.newExactQuery(getName(), intValue);
   }
 
   @Override
   public Query getTermInSetQuery(TermInSetQuery termInSetQuery) {
-    return IntPoint.newSetQuery(getName(), termInSetQuery.getIntTerms().getTermsList());
+    List<Integer> intValues;
+    switch (termInSetQuery.getTermTypesCase()) {
+      case TEXTTERMS:
+        List<Integer> intTermsList = new ArrayList<>();
```
aprudhomme, line 133: nit: initialize the ArrayList to the size of the text term list

sarthakn7 (commented):
I think this introduces a lot of inconsistency in the queries. Just boolean wasn't too bad, but now we would support text type for int but not for other numeric types like long ? Or do we add support for text type to all term queries for all types ?
I am still okay with adding this if it is really needed, but in that case we should also explore if we need similar support for other field types and other queries.
Also more code does mean more maintenance too - another reason to avoid this if we have another way.

sarthakn7:
++ about validation btw, I think this was manually added to one of the queries but we should ideally have it in a higher-level interface to apply to all queries.

erikyangs:
@sarthakn7 fair point, depends what approach we want. I think either (1) all fielddefs should support TextType, or (2) no fielddefs should support other types.

**(1) All fielddefs support TextType**
Matches ES behavior, which converts String values into appropriate types for term queries for int and bool fields.

**(2) No fielddefs support other types**
Less complexity. Validation should catch when the wrong types are given.
Only caveat is that some of our query-building makes the assumption that `String` lists will be converted into `Integer` lists in ES, which won't work for NRTSearch. Will need to rework code there to make sure correct types are used.

**Changes**
* I think (2) makes more sense - it's more strict and less conversion magic. I'll revert these changes to IntFieldDef and remove the conversion in BoolFieldDef to make things consistent.
* I'll add to this PR to do the validation for all field types.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/IntFieldDef.java
umeshdangat: nit: feel free to ignore.

Are we sure initializing with correct size is more efficient compared to initializing with empty array? I remember reading this a while back: https://shipilev.net/blog/2016/arrays-wisdom-ancients/#_empty_array_instantiation

Not sure if its relevant for Lists or only for primitive arrays..

aprudhomme: I don't think this applies to ArrayList because of how they grow. Without a capacity, it looks like primitive array allocation is deferred until the first insertion. The size is still dynamic based on the number of elements added, so neither case will have static initialization.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/BooleanFieldDef.java
```diff
@@ -140,25 +140,19 @@ public String getType() {
...
-        termValue = parseBooleanOrThrow(termQuery.getTextValue());
-        break;
-      case BOOLEANVALUE:
-        termValue = termQuery.getBooleanValue();
-        break;
-      default:
-        throw new IllegalArgumentException(
-            "BOOLEAN field does not support term type: " + termQuery.getTermTypesCase());
-    }
-    String indexTermValue = termValue ? "1" : "0";
```
erikyangs, line 161: FYI - BREAKING CHANGE
Removing support for TEXTVALUE in BooleanFieldDef means existing use cases will throw errors when inputting strings `"false"` or `"true"`.

src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
```diff
@@ -145,12 +149,33 @@ private Query getTermQuery(com.yelp.nrtsearch.server.grpc.TermQuery termQuery, I
...
+      throw new IllegalArgumentException(
+          String.format(
+              "%s field does not support term type: %s",
+              fieldDef.getName(), termQuery.getTermTypesCase()));
+    }
+  }
+
+  private void validateTermQueryIsSearchable(FieldDef fieldDef) {
+    if (fieldDef instanceof IndexableFieldDef
+        && !(fieldDef instanceof IdFieldDef)
```
erikyangs, line 165: Should IdFieldDefs always be searchable? Was failing tests since the registerFields schema in the test didn't have it as searchable.

erikyangs: I have made IdFieldDefs always required to be searchable.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/TermQueryableIndexableFieldDef.java
```diff
@@ -0,0 +1,154 @@
...
+package com.yelp.nrtsearch.server.luceneserver.field;
+
+import com.yelp.nrtsearch.server.grpc.Field;
+import com.yelp.nrtsearch.server.grpc.TermInSetQuery;
+import com.yelp.nrtsearch.server.grpc.TermQuery;
+import com.yelp.nrtsearch.server.luceneserver.field.properties.TermQueryable;
+import org.apache.lucene.search.Query;
+
+public abstract class TermQueryableIndexableFieldDef extends IndexableFieldDef
+    implements TermQueryable {
```
aprudhomme, line 25: I am not sure what being indexable adds to this class. Perhaps this could be added to the TermQueryable interface, with a default method implementations provided.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/TermQueryableIndexableFieldDef.java
```diff
@@ -0,0 +1,154 @@
...
+        return getTermQueryFromTextValue(termQuery);
+      default:
+        throw new IllegalArgumentException(
+            String.format(
+                "%s field does not support term type: %s",
+                termQuery.getField(), termQuery.getTermTypesCase()));
+    }
+  }
+
+  Query getTermQueryFromBooleanValue(TermQuery termQuery) {
```
aprudhomme, line 63: Since you already know which value of the query message is set, you could probably just pass that as the function parameter, i.e. the bool/int/float/double/text value. This would also allow the name to be shortened, as the parameter would define the type.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/TermQueryableIndexableFieldDef.java
```diff
@@ -0,0 +1,154 @@
...
+    throwIllegalArgumentException(termQuery);
+    return null;
+  }
+
+  Query getTermQueryFromDoubleValue(TermQuery termQuery) {
+    throwIllegalArgumentException(termQuery);
+    return null;
```
aprudhomme, line 70: It might be better to throw an UnsupportedOperationException, or an IllegalStateException. The other option would be for these methods to return null, and do the exception construction in the main query building method.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/TermQueryableIndexableFieldDef.java
erikyangs: Originally all the fielddefs (e.g. `BooleanFieldDef`, `IntFieldDef`, etc.) were all `<FieldDef> extends IndexableFieldDef implements TermQueryable, <other_interfaces>`. To be able to keep this structure while providing default method implementations / validation, I created this class. Is there a better way to do this?

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/TermQueryableIndexableFieldDef.java
erikyangs: Changed to pass in the expected value type.

Can't shorten function names to `getTermInSetQuery(<type>)` because List<type> can't be overloaded (https://stackoverflow.com/questions/2241514/overloading-java-function-with-list-parameter). Kept `getTermQueryFrom<Type>Value(<type>)` for consistency with this.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/TermQueryableIndexableFieldDef.java
aprudhomme: Would this be doable using interface default methods in TermQueryable, such as https://www.tutorialspoint.com/java8/java8_default_methods.htm ?

erikyangs: Ah, didn't know default methods were possible in Java interfaces. This should work, will make the changes.

erikyangs:
**Summary**
Based on many passes of reviews, this PR now does the following:

**(1) All fielddefs support TextType**
Matches ES behavior, which converts String values into appropriate types for term queries for int and bool fields.

This option was chosen to match Elasticsearch behavior.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/properties/TermQueryable.java
```diff
@@ -25,18 +26,130 @@
...
+      throw new UnsupportedOperationException(
+          String.format(
+              "%s field does not support term type: %s",
+              termQuery.getField(), termQuery.getTermTypesCase()));
+    } else {
+      return query;
+    }
+  }
+
+  default Query getTermQueryFromBooleanValue(boolean booleanValue) {
```
aprudhomme, line 72: Nit: these new interface methods should also have a java docs comment

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/properties/TermQueryable.java
```diff
@@ -25,18 +26,130 @@
...
+        break;
+      case TEXTVALUE:
+        query = getTermQueryFromTextValue(termQuery.getTextValue());
+        break;
+      default:
+        query = null;
+        break;
+    }
+
+    if (query.equals(null)) {
```
aprudhomme, line 62: This will throw a NullPointerException if query is null. You will want to use reference equality, i.e. `query == null`

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/properties/TermQueryable.java
```diff
@@ -25,18 +26,130 @@
...
+        query = getTermInSetQueryFromLongValues(termInSetQuery.getLongTerms().getTermsList());
+        break;
+      case TEXTTERMS:
+        query = getTermInSetQueryFromTextValues(termInSetQuery.getTextTerms().getTermsList());
+        break;
+      default:
+        query = null;
+        break;
+    }
+    if (query.equals(null)) {
```
aprudhomme, line 126: Same as above

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/DoubleFieldDef.java
```diff
@@ -123,12 +123,24 @@ private void ensureUpperIsMoreThanLower(RangeQuery rangeQuery, double lower, dou
...
+  }
+
+  @Override
+  public Query getTermQueryFromTextValue(String textValue) {
+    return DoublePoint.newExactQuery(getName(), Double.parseDouble(textValue));
+  }
+
+  @Override
+  public Query getTermInSetQueryFromTextValues(List<String> textValues) {
+    List<Double> doubleTerms = new ArrayList();
```
aprudhomme, line 142: Nit: ArrayList can be created with the needed size

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/FloatFieldDef.java
```diff
@@ -119,12 +119,24 @@ private void ensureUpperIsMoreThanLower(RangeQuery rangeQuery, float lower, floa
...
+  }
+
+  @Override
+  public Query getTermQueryFromTextValue(String textValue) {
+    return FloatPoint.newExactQuery(getName(), Float.parseFloat(textValue));
+  }
+
+  @Override
+  public Query getTermInSetQueryFromTextValues(List<String> textValues) {
+    List<Float> floatTerms = new ArrayList();
```
aprudhomme, line 138: Same as above

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/IntFieldDef.java
```diff
@@ -106,12 +106,24 @@ public Query getRangeQuery(RangeQuery rangeQuery) {
...
+  }
+
+  @Override
+  public Query getTermQueryFromTextValue(String textValue) {
+    return IntPoint.newExactQuery(getName(), Integer.parseInt(textValue));
+  }
+
+  @Override
+  public Query getTermInSetQueryFromTextValues(List<String> textValues) {
+    List<Integer> intTerms = new ArrayList();
```
aprudhomme, line 125: Same as above

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/LongFieldDef.java
```diff
@@ -110,12 +110,24 @@ private void ensureUpperIsMoreThanLower(RangeQuery rangeQuery, long lower, long
...
+  }
+
+  @Override
+  public Query getTermQueryFromTextValue(String textValue) {
+    return LongPoint.newExactQuery(getName(), Long.parseLong(textValue));
+  }
+
+  @Override
+  public Query getTermInSetQueryFromTextValues(List<String> textValues) {
+    List<Long> longTerms = new ArrayList();
```
aprudhomme, line 129: Same as above

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/properties/TermQueryable.java
```diff
@@ -25,18 +26,130 @@
  */
 public interface TermQueryable {
   /**
-   * Build a term query for this field type with the given configuration.
+   * Performs type validation. Do not @Override in subclasses.
+   *
+   * <p>Build a term query for this field type with the given configuration.
```
aprudhomme, line 31: Nit: can you switch the order of these lines. The first sentence becomes the summary line for generated java docs, so it should describe the functionality

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/properties/TermQueryable.java
```diff
@@ -25,18 +26,130 @@
...
+
+  default Query getTermQueryFromTextValue(String textValue) {
+    return null;
+  }
 
   /**
-   * Build a term in set query for this field type with the given configuration.
+   * Performs type validation. Do not @Override in subclasses.
+   *
+   * <p>Build a term in set query for this field type with the given configuration.
```
aprudhomme, line 99: Same as above

aprudhomme (approved):
Looks good to me

PR #214, opened 2020-12-04
Add search timeout support

aprudhomme:
Make search timeout functional. Clients may specify a timeout in seconds in the search request, which applies to the collection phase.

Currently check for timeout before processing each segment. This has lower overhead, but will allow request to go a bit longer if they timed out in the middle of a particularly large segment. If this does not end up being sufficient, we could try to add finer grain detection or implement something like terminate_after for better bounding.

Requests also have the option of specifying if partial results are ok. If not, the request will throw an exception.

Other changes:
- DocCollector is now an abstract class instead of an interface, to better share some common implementation
- ServerTestCase implementations can now inject additional configuration to be appended to the config file


src/test/java/com/yelp/nrtsearch/server/grpc/TimeoutTest.java
```diff
@@ -0,0 +1,399 @@
...
+                assertTrue(hasTimeout[0]);
+                return topDocs;
+              } catch (IOException e) {
+                throw new RuntimeException(e);
+              }
+            });
+    assertEquals(20, hits.totalHits.value);
+  }
+
+  @Test(expected = CollectionTimeoutException.class)
```
sarthakn7, line 264: Will this exception be thrown even though `disallowPartialResults` is not set in the request ? Same for the next test too

umeshdangat (approved):
looks good to me !

src/test/java/com/yelp/nrtsearch/server/grpc/TimeoutTest.java
aprudhomme: The test is using its own version of the timeout wrapper, so it doesn't need to configure the one added by the request processor.

PR #220, opened 2020-12-08
Fix direct memory leaks when transferring copy job

aprudhomme:
The newCopyOneFile method is called when transferring the current file being copied to a new copy job. Since the file copy is not tied to the job at all, it can be moved as is. The previous code made a new file copy request and didn't drain the old iterator.

A copy job can be closed before completion if the replica no longer needs the file. The chunk iterator should be drained in the close.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/SimpleCopyJob.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/SimpleCopyJob.java
@@ -32,7 +32,6 @@ import org.apache.lucene.replicator.nrt.NodeCommunicationException;
 import org.apache.lucene.replicator.nrt.ReplicaNode;
 
 public class SimpleCopyJob extends CopyJob {
-  final byte[] copyBuffer = new byte[65536];
   private final CopyState copyState;
   private final ReplicationServerClient primaryAddres;
   private final String indexName;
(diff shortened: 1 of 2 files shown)


src/main/java/com/yelp/nrtsearch/server/luceneserver/SimpleCopyJob.java
```diff
@@ -56,19 +55,8 @@ public SimpleCopyJob(
...
-        cancel("exc during start", t);
-      } catch (IOException e) {
-        throw new NodeCommunicationException("cancel IOException during newCopyOneFile", e);
-      }
-      throw new NodeCommunicationException("exc during start", t);
-    }
-
-    return new CopyOneFile(prev, rawFileChunkIterator);
+    // no state needs to be changed when transferring to a new job
+    return prev;
```
sarthakn7, line 59: Should we check if `prev` is closed, and if it is we create a new `CopyOneFile` ? Just wondering if Lucene can change the code to close `current` before trying to create a new `CopyOneFile`

aprudhomme: The main purpose of this method is to transfer the in progress copy state, in place of closing the file copy. If prev was closed, even the existing code would fail since the index file would also be closed.

sarthakn7 (approved):

src/main/java/com/yelp/nrtsearch/server/luceneserver/SimpleCopyJob.java
umeshdangat: I think this change should work.

[This](https://github.com/apache/lucene-solr/blob/1d85cd783863f75cea133fb9c452302214165a4d/lucene/replicator/src/test/org/apache/lucene/replicator/nrt/SimpleCopyJob.java#L192) is the lucene test code I based this code off. 

Seems like the original test code simply made a new `CopyOneFile` object but re-used the `DataInput` i.e. made a shallow copy only.

src/main/java/org/apache/lucene/replicator/nrt/CopyOneFile.java
```diff
@@ -98,6 +81,11 @@ public CopyOneFile(CopyOneFile other, Iterator<RawFileChunk> rawFileChunkIterato
   public void close() throws IOException {
     out.close();
     dest.finishCopyFile(name);
+    // This job may have been canceled before being completed, meaning the replica no longer needs
+    // it. Drain the iterator to not leak direct memory.
+    while (rawFileChunkIterator.hasNext()) {
```
umeshdangat, line 86: ++

umeshdangat (approved):

PR #224, opened 2020-12-10
Add arcDistance for Location doc values

desiam:
Uses Haversin distance:
`SingleLocation`: Calculates distance between input point, and the doc value GeoPoint
`Locations`: Throws `IllegalArgumentException`, because distance is between two points, and we only want one output distance.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/doc/LoadedDocValues.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/doc/LoadedDocValues.java
@@ -296,6 +296,10 @@ public abstract class LoadedDocValues<T> extends AbstractList<T> {
           LatLng.newBuilder().setLatitude(point.getLat()).setLongitude(point.getLon()).build();
       return SearchResponse.Hit.FieldValue.newBuilder().setLatLngValue(latLon).build();
     }
+
+    public double arcDistance(double lat, double lon) {
+      return getValue().arcDistance(lat, lon);
+    }
   }
 
   public static final class Locations extends SortedNumericValues<GeoPoint> {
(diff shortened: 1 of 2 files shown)


src/main/java/com/yelp/nrtsearch/server/luceneserver/doc/LoadedDocValues.java
```diff
@@ -296,6 +296,10 @@ public GeoPoint getValue() {
           LatLng.newBuilder().setLatitude(point.getLat()).setLongitude(point.getLon()).build();
       return SearchResponse.Hit.FieldValue.newBuilder().setLatLngValue(latLon).build();
     }
+
+    public double arcDistance(double lat, double lon) {
```
sarthakn7, line 300: Where will this method be called from?

taoyyu: ads uses this in their painless.

desiam: It would be used in a painless script for SingleLocation doc values (Ex. `return doc['location'].arcDistance(params.lat, params.lon)`)

sarthakn7: you can try it on a local instance before merging

desiam: Tested locally with painless, which worked as expected
```
    @Test
    public void testLocationArcDistance() {
        double lat = 0.0;
        double lon = 0.0;
        assertScript(String.format("return doc['lat_lon'].arcDistance(%f, %f)", lat, lon),
                1.27933529230076E7, 1.2769084708787214E7);
    }
```

taoyyu (approved):

PR #226, opened 2020-12-12
Reduce scope of synchronization in merge segment warming

aprudhomme:
Instead of waiting for merge warming to complete under synchronization of preCopy.connections, create a copy of the current connections, process them, and remove them. This should remove the possibility of merge warming when adding replicas deadlocking.


sarthakn7 (approved):
Lgtm, would be great if @umeshdangat can take a look as well

umeshdangat (approved):
lg2m !

PR #236, opened 2021-01-06
nested query support

taoyyu:
inner hit not supported yet. ads does not use it yet, but search will need that in the future for some experiences.


src/test/resources/field/registerFieldsObject.json
```diff
@@ -1,12 +1,23 @@
...
+      "search": true,
+      "store": true
+    },
     {
       "name": "doc_id",
       "type": "ATOM",
       "search": true,
       "storeDocValues": true
     },
+    {
```
taoyyu, line 16: currently I require the user to add these 2 fields in the schema explicitly. I can hide this to the user. Wondering which approach is preferred. I feel some query like empty query may end up return child docs. The user can use _nested_type to indicates they only want parent docs

src/test/resources/field/registerFieldsObject.json
```diff
@@ -1,12 +1,23 @@
...
+      "name": "_nested_type",
```
taoyyu, line 17: currently I require the user to add these 2 fields(_nested_type, _nested_path) in the schema explicitly. I can hide this to the user. Wondering which approach is preferred. I feel some query like empty query may end up return child docs. The user can use _nested_type to indicates they only want parent docs

aprudhomme: I think ES automatically adds a filter to the base query so that only root documents are considered, with child document fields returned in inner hits.

If we decide we want to add the ability to query child fields directly, we would probably want it to be a request level option to specify the nested path to filter by. This would avoid needing to expose these metafields in the index schema.

aprudhomme (commented):
It might be possible to just use _nested_path for the root doc too and avoid using _nested_type all together.

Another possible way to do doc filtering is with a DocValuesFieldExistsQuery. I'm not sure if this is more efficient than a term query.

taoyyu:
addressed some comments. I add the code to register _nested_path field for every new index so I don't need to worry about user register schema not good now. I removed _nested_type and use constant query as filters to avoid messing up with scoring for hiddend fields as suggested in reviews.
Also noticed nrtsearch does not support exists query yet, which is widely used for ads... Going to send a separate pr for that. ES have a hidden field which contains all non-null fields for each document. Going to do something similar

taoyyu:
update pr again support to retrieve object and polygon fields as struct/json(ads use it a lot). Store them as storedField.

aprudhomme (commented):
Stored fields are being used to make polygon/objects retrievable, but I thought we usually prefer doc values when possible. Doc values could also be supported by adding the json strings as sorted set doc values and creating a LoadedDocValues implementation that decodes them to maps. https://github.com/Yelp/nrtsearch/blob/master/src/main/java/com/yelp/nrtsearch/server/luceneserver/doc/LoadedDocValues.java#L399

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/SearchRequestProcessor.java
```diff
@@ -245,6 +245,11 @@ private static Query extractQuery(IndexState state, SearchRequest searchRequest)
     } else {
       q = new MatchAllDocsQuery();
     }
+
+    if (searchRequest.getQueryNestedPath() != null
+        && searchRequest.getQueryNestedPath().length() > 0) {
+      return QUERY_NODE_MAPPER.applyQueryNestedPath(q, searchRequest.getQueryNestedPath());
+    }
```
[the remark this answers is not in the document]
taoyyu: I am concerned about backward compatibility here. For old index, there is no _nested_path field indexed. They will have to backfill all the data after bumping the docker image version

src/main/java/com/yelp/nrtsearch/server/luceneserver/RegisterFieldsHandler.java
```diff
@@ -132,6 +133,27 @@ public FieldDefResponse handle(final IndexState indexState, FieldDefRequest fiel
         indexState.addField(ent.getValue(), fieldAsJsonObject);
       }
     }
+
+    if (!indexState.getAllFields().containsKey("_nested_path")) {
+      // register _nested_path field to every index to differentiate root
+      // and child documents
```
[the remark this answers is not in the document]
taoyyu: I can only add a registered field. If _nested_path is not nested. I am not able to document add it.

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/SearchRequestProcessor.java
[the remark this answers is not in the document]
aprudhomme: You could not add any filter if the index has no nested fields.

src/main/java/com/yelp/nrtsearch/server/luceneserver/RegisterFieldsHandler.java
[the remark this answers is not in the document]
aprudhomme: You are only adding the the lucene Document, right? This does not do any check against the nrtsearch schema.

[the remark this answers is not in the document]
taoyyu: I got this exception
`Caused by: java.lang.IllegalArgumentException: field "_nested_path" is unknown: it was not registered with registerField
	at com.yelp.nrtsearch.server.luceneserver.IndexState.getField(IndexState.java:671)
	at com.yelp.nrtsearch.server.luceneserver.ServerCodec.getPostingsFormatForField(ServerCodec.java:43)
	at org.apache.lucene.codecs.lucene84.Lucene84Codec$1.getPostingsFormatForField(Lucene84Codec.java:67)
	at org.apache.lucene.codecs.perfield.PerFieldPostingsFormat$FieldsWriter.buildFieldsGroupMapping(PerFieldPostingsFormat.java:219)
	at org.apache.lucene.codecs.perfield.PerFieldPostingsFormat$FieldsWriter.write(PerFieldPostingsFormat.java:151)
	at org.apache.lucene.index.FreqProxTermsWriter.flush(FreqProxTermsWriter.java:97)
	at org.apache.lucene.index.DefaultIndexingChain.flush(DefaultIndexingChain.java:177)
	at org.apache.lucene.index.DocumentsWriterPerThread.flush(DocumentsWriterPerThread.java:468)
	at org.apache.lucene.index.DocumentsWriter.doFlush(DocumentsWriter.java:555)
	at org.apache.lucene.index.DocumentsWriter.flushAllThreads(DocumentsWriter.java:722)
	at org.apache.lucene.index.IndexWriter.getReader(IndexWriter.java:494)
	at org.apache.lucene.index.StandardDirectoryReader.doOpenFromWriter(StandardDirectoryReader.java:297)
	at org.apache.lucene.index.StandardDirectoryReader.doOpenIfChanged(StandardDirectoryReader.java:272)
	at org.apache.lucene.index.StandardDirectoryReader.doOpenIfChanged(StandardDirectoryReader.java:262)
	at org.apache.lucene.index.DirectoryReader.openIfChanged(DirectoryReader.java:165)
	at org.apache.lucene.facet.taxonomy.SearcherTaxonomyManager.refreshIfNeeded(SearcherTaxonomyManager.java:147)
	at org.apache.lucene.facet.taxonomy.SearcherTaxonomyManager.refreshIfNeeded(SearcherTaxonomyManager.java:41)
	at org.apache.lucene.search.ReferenceManager.doMaybeRefresh(ReferenceManager.java:176)
	at org.apache.lucene.search.ReferenceManager.maybeRefreshBlocking(ReferenceManager.java:253)
	at com.yelp.nrtsearch.server.luceneserver.ShardState.maybeRefreshBlocking(ShardState.java:982)
	at com.yelp.nrtsearch.server.grpc.LuceneServer$LuceneServerImpl.refresh(LuceneServer.java:672)
	at com.yelp.nrtsearch.server.grpc.LuceneServerGrpc$MethodHandlers.invoke(LuceneServerGrpc.java:2664)
	at io.grpc.stub.ServerCalls$UnaryServerCallHandler$UnaryServerCallListener.onHalfClose(ServerCalls.java:172)
	at io.grpc.internal.ServerCallImpl$ServerStreamListenerImpl.halfClosed(ServerCallImpl.java:331)
	at io.grpc.internal.ServerImpl$JumpToApplicationThreadServerStreamListener$1HalfClosed.runInContext(ServerImpl.java:820)
	at io.grpc.internal.ContextRunnable.run(ContextRunnable.java:37)
	at io.grpc.internal.SerializingExecutor.run(SerializingExecutor.java:123)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630)
	... 1 more`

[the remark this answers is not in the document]
aprudhomme: I see, it needs the postings format. In that case, maybe add a static collection of meta FieldDefs to IndexState and change getField to check the meta fields first. This should keep it out of the schema, but still make it available.

You would also want to use this FieldDef to add fields to the document. i.e. metaFieldDef.parseDocumentField(...) instead of document.add(new StringField(...))

src/main/java/com/yelp/nrtsearch/server/luceneserver/doc/LoadedDocValues.java
```diff
@@ -333,6 +341,44 @@ public SortedDateTimes(SortedNumericDocValues docValues) {
...
+        value = null;
+      }
+    }
+
+    @Override
+    public Map<String, Object> get(int index) {
+      if (value == null) {
+        throw new IllegalStateException("No doc values for document");
+      }
+      return value.get(index);
```
aprudhomme, line 367: It would be good to also check the index range and throw IndexOutOfBoundsException

src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java
```diff
@@ -441,6 +449,9 @@ public IndexState(
       initSaveLoadState();
     }
     searchThreadPoolExecutor = globalState.getSearchThreadPoolExecutor();
+
+    // add meta data fields
+    metaFields = getPredefinedMetaFields();
```
aprudhomme, line 454: Can this initialization be done statically. I think the meta field defs should be the same for all indices.

src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java
```diff
@@ -530,6 +541,9 @@ public synchronized void load(JsonObject jsonObject) throws IOException {
   /** The field definitions (registerField) */
   private final Map<String, FieldDef> fields = new ConcurrentHashMap<String, FieldDef>();
 
+  /** The meta field definitions */
+  public static Map<String, FieldDef> metaFields;
```
aprudhomme, line 545: It would be good to change IndexState::PerFieldSimilarityWrapper to remove `if (fields.containsKey(name))`. It would also be good to have IndexState.addField() reject fields with the same name as a meta field.

aprudhomme (commented):
Looks good, just a few minor comments.

src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java
taoyyu: nope. It will throw NPE if I make it static. SimilarityCreator.getInstance() and FieldDefCreator.getInstance() are both null otherwise. It seems we have a lot of static instance. But they are all initialized in constructors... Don't want to change all of these code here.

aprudhomme (approved):

PR #240, opened 2021-01-15
Implement a new dictionary to store index for suggest item

jinxihexi0411:
Upstream [branch](https://github.com/Yelp/nrtsearch/pull/239).

`search text`s are required for suggesting exact and fuzzy match of infix suggestion. However, the native `DocumentDictionary` does not support to read such info from the index. In this regard, a new dictionary, named as `SuggestDocumentDictionary`, is implemented to extend the support on `search text` field.

Test done:

- all tests pass
- added a test to make sure `SuggestDocumentInputIterator` is able to read out `field`, `payload`, `contexts`, and `search_texts` fields from index


umeshdangat (approved):
lg2m

Issue #243, opened 2021-01-20
Log search request in json in case of an error


sarthakn7:
Currently if a search fails the request is logged with a toString so it looks like:
```
topHits: 20
retrieveFields: "field1"
retrieveFields: "field2"
query {
  functionScoreQuery {
    query {
      booleanQuery {
....
```
This makes it tough to execute these logged queries as directly. Instead let's log these as json.

PR #244, opened 2021-01-20
Logging search request as json in case of an error

sarthakn7:
Logging search request as json in case of an error so the log now looks like this:

```
WARNING: error while trying to execute search for index test_index: request: {"indexName":"test_index","topHits":10,"retrieveFields":["doc_id","license_no"],"query":{"boost":-1.0,"booleanQuery":{"clauses":[{"query":{"phraseQuery":{"field":"vendor_name","terms":["first","again"]}},"occur":"MUST"}]}}}
```
Removed the newline so the complete error is logged in one line and also removed logging the query when the boost is negative as the request will be logged anyway.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
@@ -57,7 +57,7 @@ public class QueryNodeMapper {
     Query queryNode = getQueryNode(query, state);
 
     if (query.getBoost() < 0) {
-      throw new IllegalArgumentException("Boost must be a positive number, query: " + query);
+      throw new IllegalArgumentException("Boost must be a positive number");
     }
 
     if (query.getBoost() > 0) {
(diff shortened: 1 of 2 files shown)


sarthakn7:
Fixes #243 .

src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
```diff
@@ -57,7 +57,7 @@ public Query getQuery(com.yelp.nrtsearch.server.grpc.Query query, IndexState sta
     Query queryNode = getQueryNode(query, state);
 
     if (query.getBoost() < 0) {
-      throw new IllegalArgumentException("Boost must be a positive number, query: " + query);
```
erikyangs, line 60: This could throw the same errors as in https://github.com/Yelp/nrtsearch/pull/242 if the query is too large.
```
Jan 19, 2021 5:33:01 PM io.grpc.netty.shaded.io.grpc.netty.NettyServerHandler onStreamError
WARNING: Stream Error
io.grpc.netty.shaded.io.netty.handler.codec.http2.Http2Exception$HeaderListSizeException: Header size exceeded max allowed size (8192)
...
```

erikyangs (approved):

src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
erikyangs: oops, this is a removal. all good!

PR #247, opened 2021-01-21
Add exists query

desiam:
Based on https://www.elastic.co/guide/en/elasticsearch/reference/7.2/query-dsl-exists-query.html
* Use a hidden field: `ATOM` `_field_names` to track field names with values
* Fields with `search: false` will return values (may not be the case in the latest ES version)
* Uses a simple `TermQuery` wrapped in a `ConstantScoreQuery`

In theory `null` and `[]` wouldn't result in added, so they should be accounted for. Added a couple tests for top level, nested query exists, and an a flattened object field.

Notes
* **Can move test into `nrtsearch/grpc/QueryTest.java` or elsewhere if that's more appropriate**
* **Added a couple simple tests with `null` to prove that we don't need to account for `null` values for now, but can remove**

Left out:
* Exists query on nested object or object name (Ex. `pickup_partners`)


src/test/java/com/yelp/nrtsearch/server/ExistsQueryTest.java
```diff
@@ -0,0 +1,164 @@
...
+                            .setExistsQuery(
+                                ExistsQuery.newBuilder().setField("pickup_partners.hours").build()))
+                    .build())
+            .build();
+
+    queryAndVerifyIds(nestedQuery, "2");
+  }
+
+  // TODO:(samir|2020-01-21): Test exists query on object itself
+  //  (Ex. delivery_areas or delivery_areas.partner)
```
taoyyu, line 130: currently exists query may not work for object field. If object field is not stored or store dco value, document.getFields will not have it. We could handle they separately. later.

src/main/java/com/yelp/nrtsearch/server/luceneserver/AddDocumentHandler.java
```diff
@@ -92,9 +92,34 @@ public static DocumentsContext getDocumentsContext(
...
+    /** Extract all field names in the document and stores it into a hidden field */
+    private static void extractFieldNamesForDocument(Document document) {
+      IndexableFieldDef fieldNamesFieldDef =
+          (IndexableFieldDef) IndexState.getMetaField(IndexState.FIELD_NAMES);
+
+      List<String> fieldNames =
+          document.getFields().stream()
+              .map(IndexableField::name)
+              .distinct()
+              .collect(Collectors.toList());
```
taoyyu, line 118: maybe filter out the predefined meta fields like _nested_path

taoyyu (commented):
You need to run ./gradlew buildGrpcGateway to update the generated files.

src/test/java/com/yelp/nrtsearch/server/ExistsQueryTest.java
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  @Test
+  public void testTopLevelField() {
+    // Empty strings become valid lucene fields
+    ExistsQuery existsQuery = ExistsQuery.newBuilder().setField("name").build();
+    queryAndVerifyIds(existsQuery, "1", "2");
+  }
+
+  // TODO:(samir|2020-01-21): Test exists query on nested object itself (Ex. pickup_partners)
```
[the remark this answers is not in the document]
desiam: Yeah, let's drop support for this. We will only be querying nested fields in a nested context anyways.

src/test/java/com/yelp/nrtsearch/server/ExistsQueryTest.java
desiam: Yeah I'm leaving the TODO here to test exists on the object itself. I'm not certain if ES even supports it, and I don't believe this will be a blocker for ads_search.

src/main/java/com/yelp/nrtsearch/server/luceneserver/AddDocumentHandler.java
desiam: This is based on https://github.com/elastic/elasticsearch/blob/e144471b3e607cc8b6b3c7a0e7f4706a4bdb9381/server/src/main/java/org/elasticsearch/indices/IndicesModule.java#L184, which explicitly `//_field_names must be added last so that it has a chance to see all the other mappers`. I don't believe adding the meta-fields as terms will significantly affect performance, but we can remove them if needed.

I also tested exists queries on `_id` and `_version` and it works as intended. It could be useful for meta fields we add in the future.

src/test/java/com/yelp/nrtsearch/server/grpc/LuceneServerTest.java
```diff
@@ -432,7 +432,7 @@ public void testStats() throws IOException, InterruptedException {
     assertEquals(0, statsResponse.getOrd());
     assertEquals(2, statsResponse.getCurrentSearcher().getNumDocs());
     assertEquals(1, statsResponse.getCurrentSearcher().getNumSegments());
-    assertEquals(6610, statsResponse.getDirSize(), 1000);
+    assertEquals(6610, statsResponse.getDirSize(), 1500);
```
taoyyu, line 435: any reason about this change? Did you change anything related?

desiam: My changes did end up causing this issue. However, this assert might end up being flaky, because it's calculating a directory size which can vary based on a number of factors.

taoyyu (approved):

PR #253, opened 2021-01-27
Support epoch_millis in DateTimeField

jedrazb:
### Changes 
Support `epoch_millis` as DATE_TIME field format

### Testing
Unit tests:
* range query for epoch_millis DATE_TIME
* adding docs with epoch_millis DATE_TIME field

Manual testing:
* Registered following fields
```
'{                                                      
   "indexName":"testIdx",
   "field":[
      {
         "name":"id",
         "type":"ATOM",
         "storeDocValues":true
      },
      {
         "name":"title",
         "type":"TEXT",
         "search":true,
         "store":true,
         "tokenize":true
      },
                        {
        "name": "timestamp",
        "type": "DATE_TIME",
        "storeDocValues": true,
        "dateTimeFormat": "epoch_millis"
      }
   ]
}
```
* Index following csv document
```
id,title,timestamp
0,first vendor,1611678698815
1,second vendor,1611678827275
```
* Query the index, timestamp is added as expected
```
{"diagnostics":{"parsedQuery":"*:*","rewrittenQuery":"*:*","firstPassSearchTimeMs":3.678646,"getFieldsTimeMs":15.7659},"totalHits":{"value":"2"},"hits":[{"score":1,"fields":{"id":{"fieldValue":[{"t$xtValue":"0"}]},"timestamp":{"fieldValue":[{"longValue":"1611678698815"}]},"title":{"fieldValue":[{"textValue":"first vendor"}]}}},{"luceneDocId":1,"score":1,"fields":{"id":{"fieldValue":[{"textValue":"1"}]},"timestamp":{"fieldValue":[{"longValue":"1611678827275"}]},"title":{"fieldValue":[{"textValue":"second vendor"}]}}}],"searchState":{}}%                                                     
```


src/main/java/com/yelp/nrtsearch/server/luceneserver/field/DateTimeFieldDef.java
```diff
@@ -97,11 +101,16 @@ public Query getRangeQuery(RangeQuery rangeQuery) {
...
   }
 
-  private static long convertDateStringToMillis(
-      String dateString, DateTimeFormatter dateTimeFormatter) {
-    return LocalDateTime.parse(dateString, dateTimeFormatter)
-        .toInstant(ZoneOffset.UTC)
-        .toEpochMilli();
+  private static long convertDateStringToMillis(String dateString, String dateTimeFormat) {
+    switch (dateTimeFormat) {
+      case EPOCH_MILLIS:
```
sarthakn7, line 106: I see switch-case used here and elsewhere for only 2 and sometimes 1 case including the default case. Are we planning to add more conditions? If not we can use if-else

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/DateTimeFieldDef.java
```diff
@@ -243,6 +268,20 @@ private long getTimeToIndex(String dateString) {
...
   }
 
+  private long getTimeFromEpochMillisString(String epochMillisString) {
+    String format =
+        String.format(
+            "%s could not parse %s as date_time with format %s",
+            getName(), epochMillisString, dateTimeFormat);
+    try {
+      long epochMillisLong = Long.parseLong(epochMillisString);
+      Instant epochMillisInstant = Instant.ofEpochMilli(epochMillisLong);
```
sarthakn7, line 278: Instead of converting to an Instant we can probably just validate this by checking that epochMillisLong is positive. If epochMillisString is parsed to a long without an error I think the number being negative is the only invalid state possible

src/test/java/com/yelp/nrtsearch/server/luceneserver/field/DateTimeFieldDefTest.java
```diff
@@ -0,0 +1,145 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+public class DateTimeFieldDefTest extends ServerTestCase {
```
sarthakn7, line 39: Should we add tests for invalid values throwing errors?

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/DateTimeFieldDef.java
aprudhomme: What if it is a time before the epoch, wouldn't that be negative?

sarthakn7: @aprudhomme good point. Then we don't need any additional validation at all

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/DateTimeFieldDef.java
jedrazb: Sure, changed it to if-else

src/test/java/com/yelp/nrtsearch/server/luceneserver/field/DateTimeFieldDefTest.java
jedrazb: Added more test for indexing and querying with invalid formats

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/DateTimeFieldDef.java
jedrazb: Ok, in this case the only validation is parsing input as to `long`

jedrazb:
- Added more tests for `DATE_TIME` field. New tests check if errors are thrown if we pass datetime value formatted in the incorrect way

src/test/java/com/yelp/nrtsearch/server/luceneserver/field/DateTimeFieldDefTest.java
```diff
@@ -0,0 +1,200 @@
...
+                        .setField("timestamp_string_format")
+                        .setLower("2021-01-27 10:05:40")
+                        .setUpper("2021-01-27 10:07:40")
+                        .build())
+                .build(),
+            List.of("doc_id"));
+    assertFields(response, "1");
+  }
+
+  @Test(expected = RuntimeException.class)
```
sarthakn7, line 113: Is there a less general exception we can test? Every unchecked exception is going to be a RuntimeException. You can also catch the exception and assert the message like in https://github.com/Yelp/nrtsearch/blob/master/src/test/java/com/yelp/nrtsearch/server/grpc/LuceneServerTest.java#L952 which is better when an exception specific to your error is not thrown

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/DateTimeFieldDef.java
```diff
@@ -55,6 +55,8 @@
 public class DateTimeFieldDef extends IndexableFieldDef implements Sortable, RangeQueryable {
   private static final long MIN_DATE_TIME = 0;
 
+  private static final String EPOCH_MILLIS = "epoch_millis";
```
sarthakn7, line 58: It would be helpful to have it mentioned in https://github.com/Yelp/nrtsearch/blob/master/clientlib/src/main/proto/yelp/nrtsearch/luceneserver.proto#L419 that the format can be "epoch_millis" too apart from formats like "dd-mm-yyyy"

src/test/java/com/yelp/nrtsearch/server/luceneserver/field/DateTimeFieldDefTest.java
jedrazb: Decided to assert for the exception message

sarthakn7 (approved):

PR #258, opened 2021-01-29
Enabled specifying exclusive bounds for range queries

sarthakn7:
With these changes we can now set `lowerExclusive` or `upperExclusive` or both parameters to true in `RangeQuery` to make the lower or upper or both bounds exclusive. The `ensureUpperIsMoreThanLower` check is being called after the bound has been changed.
Additional changes:
1. Added `ensureUpperIsMoreThanLower` check for `IntFieldDef`, other field definitions already have this
2. Using `Float.NEGATIVE_INFINITY`, `Float.POSITIVE_INFINITY`, `Double.NEGATIVE_INFINITY` and `Double.POSITIVE_INFINITY` for range queries where the respective bounds are not provided instead of `Float.MIN_VALUE`, `Float.MAX_VALUE`, `Double.MIN_VALUE`, `Double.MAX_VALUE` since the infinity values are recommended by Lucene ([FloatPoint#newRangeQuery](https://lucene.apache.org/core/8_4_0/core/org/apache/lucene/document/FloatPoint.html#newRangeQuery-java.lang.String-float-float-), [DoublePoint#newRangeQuery](https://lucene.apache.org/core/8_4_0/core/org/apache/lucene/document/DoublePoint.html#newRangeQuery-java.lang.String-double-double-))
3. Bumped version to create a new version of the clientlib with the new parameters


sarthakn7:
Also I haven't added tests for `DateTimeFieldDef` yet to reduce conflicts with #253 .

erikyangs (approved):

PR #261, opened 2021-02-01
add GeoRadiusQuery support to nrtsearch

taoyyu:
Support query by distance in nrtsearch with lucene LatLonPoint.newDistanceQuery.

This implementation is not exactly the same from ES. ES support 2 distance type: arc and plane(arc is default). LatLonPoint.newDistanceQuery is just the arc implementation and plane is not accurate. We just use arc distance for all use cases at Yelp so I think think this should be enough.


src/main/java/com/yelp/nrtsearch/server/luceneserver/geo/GeoUtils.java
```diff
@@ -0,0 +1,51 @@
...
+      if (distanceList.length == 1) {
+        return distanceNumber;
+      }
+      String distanceUnit = distanceList[1];
+      if (distanceUnit.equals("m")) {
+        return distanceNumber;
+      } else if (distanceUnit.equals("km")) {
+        return distanceNumber * 1000;
+      } else if (distanceUnit.equals("mi")) {
+        return distanceNumber * 1609.344;
```
aprudhomme, line 43: nit: could the constants be named. something like KM_TO_M/MI_TO_M

aprudhomme (approved):

PR #265, opened 2021-02-10
Add scripted filter query

jedrazb:
### Support for scripted filter

Add support of [Elasticsearch-like FilterScript](https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/script/FilterScript.java).

### Solution 
Use [Lucene's FunctionMatchQuery ](https://lucene.apache.org/core/7_4_0/queries/org/apache/lucene/queries/function/FunctionMatchQuery.html) with filter script which calculates the document's score, then by default we only select the candidates with positive score.

### Testing
Unit tests pass


src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
```diff
@@ -185,6 +188,15 @@ private FunctionScoreQuery getFunctionScoreQuery(
...
 
+  private FunctionMatchQuery getFunctionFilterQuery(
+      FunctionFilterQuery functionFilterQuery, IndexState state) {
+    ScoreScript.Factory scriptFactory =
+        ScriptService.getInstance().compile(functionFilterQuery.getScript(), ScoreScript.CONTEXT);
+    Map<String, Object> params =
+        ScriptParamsUtils.decodeParams(functionFilterQuery.getScript().getParamsMap());
+    return new FunctionMatchQuery(scriptFactory.newFactory(params, state.docLookup), x -> x > 0);
+  }
+
```
taoyyu, line 199: seems FunctionFilterQuery.getQuery is not used? We can delete query field if it is not used

src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
```diff
@@ -185,6 +188,15 @@ private FunctionScoreQuery getFunctionScoreQuery(
         scriptFactory.newFactory(params, state.docLookup));
   }
 
+  private FunctionMatchQuery getFunctionFilterQuery(
+      FunctionFilterQuery functionFilterQuery, IndexState state) {
+    ScoreScript.Factory scriptFactory =
+        ScriptService.getInstance().compile(functionFilterQuery.getScript(), ScoreScript.CONTEXT);
+    Map<String, Object> params =
+        ScriptParamsUtils.decodeParams(functionFilterQuery.getScript().getParamsMap());
+    return new FunctionMatchQuery(scriptFactory.newFactory(params, state.docLookup), x -> x > 0);
```
sarthakn7, line 197: Nit: Instead of `x` can use a more meaningful variable name like `score`

taoyyu:
lgtm. not sure about the failed check though. Does make test pass locally?\?

taoyyu (approved):
Not very sure about the log in the check failure.
If the tests pass locally, ship

PR #266, opened 2021-02-12
Support rescore

jedrazb:
### Changes
- Add two grpc structs, `QueryRescorer` and `PluginRescorer`
- `QueryRescorer` is a wrapper a top of original Lucene's QueryRescorer class. The final score when using query rescorer is:
```
queryWeight * firstPassScore + secondQueryWeight * secondPassScore
```
- `PluginRescorer` allows for using external rescore plugins. Rescore plugins must extend from Lucene's `Rescorer` class

### Validation
- unit tests for QueryRescorer and RescorerProvider (plugin)


clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -251,6 +251,19 @@ message Query {
     }
 }
 
+message PluginRescorer {
+    int32 windowSize = 1;
+    Script script = 2;
```
aprudhomme, line 256: I would suggest just adding the needed fields here instead of using the Script message, since this isn't really a script. You could probably get away with using a google.protobuf.Struct for parameters, since typing shouldn't be a big issue.

clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -251,6 +251,19 @@ message Query {
...
+}
+
+message QueryRescorer {
+    int32 windowSize = 1;
```
aprudhomme, line 260: If all Rescorer definitions will include the windowSize, it could be moved to the main message.

src/main/java/com/yelp/nrtsearch/server/luceneserver/rescore/RescoreManager.java
```diff
@@ -0,0 +1,68 @@
...
+        Script plugin = pluginRescorer.getScript();
+
+        org.apache.lucene.search.Rescorer rescorerFromPlugin =
+            RescorerCreator.getInstance()
+                .createRescorer(
+                    plugin.getLang(), ScriptParamsUtils.decodeParams(plugin.getParamsMap()));
+
+        hits = rescorerFromPlugin.rescore(searcher, hits, pluginRescorer.getWindowSize());
+      }
+    }
```
aprudhomme, line 64: I would suggest building the Rescorers as part of the request processing https://github.com/Yelp/nrtsearch/blob/master/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/SearchRequestProcessor.java#L75 and adding the final list of built `org.apache.lucene.search.Rescorer`s to the SearchContext. This way requests will fail fast on an invalid rescorer definition, rather than after the first search pass.

src/main/java/com/yelp/nrtsearch/server/luceneserver/rescore/RescoreManager.java
```diff
@@ -0,0 +1,68 @@
...
+      TopDocs firstPassTopDocs,
+      IndexState indexState)
+      throws IOException {
+
+    TopDocs hits = firstPassTopDocs;
+
+    for (Rescorer rescorer : rescorers) {
+      if (rescorer.hasQueryRescorer()) {
+        QueryRescorer queryRescorer = rescorer.getQueryRescorer();
+        Query query = queryNodeMapper.getQuery(queryRescorer.getRescoreQuery(), indexState);
```
aprudhomme, line 47: I think you will want to call https://lucene.apache.org/core/8_0_0/core/org/apache/lucene/search/IndexSearcher.html#rewrite-org.apache.lucene.search.Query- on the built query.

src/main/java/com/yelp/nrtsearch/server/luceneserver/rescore/RescoreManager.java
jedrazb: Sure! I ended up using a wrapper component `RescoreTask` which wraps around `org.apache.lucene.search.Rescorer` and `int windowSize`. It makes the rescoring easier since we need to know the `windowSize` argument when calling `rescore()`

clientlib/src/main/proto/yelp/nrtsearch/search.proto
jedrazb: 👍

clientlib/src/main/proto/yelp/nrtsearch/search.proto
jedrazb: Okok, I used `google.protobuf.Struct` for params when defining PluginRescorer

src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
```diff
@@ -303,6 +304,7 @@ private void initExtendableComponents(
       FieldDefCreator.initialize(configuration, plugins);
       ScriptService.initialize(configuration, plugins);
       SimilarityCreator.initialize(configuration, plugins);
+      RescorerCreator.initialize(configuration, plugins);
```
aprudhomme, line 307: nit: alphabetize

src/main/java/com/yelp/nrtsearch/server/luceneserver/SearchHandler.java
```diff
@@ -146,7 +147,16 @@ public SearchResponse handle(IndexState indexState, SearchRequest searchRequest)
...
-      diagnostics.setFirstPassSearchTimeMs(((System.nanoTime() - searchStartTime) / 1000000.0));
+      long firstPassSearchTime = System.nanoTime() - searchStartTime;
+
+      diagnostics.setFirstPassSearchTimeMs((firstPassSearchTime / 1000000.0));
+
+      if (!searchContext.getRescorers().isEmpty()) {
+        for (RescoreTask rescorer : searchContext.getRescorers()) {
+          hits = rescorer.rescore(s.searcher, hits);
+        }
+        diagnostics.setRescoreTimeMs(((System.nanoTime() - firstPassSearchTime) / 1000000.0));
```
aprudhomme, line 158: firstPassSearchTime is the first pass search duration in ns. You will want to add a `long rescoreStartTime = System.nanoTime();` before you start the rescore and use that to compute the rescore time.

aprudhomme (commented):
Looks good. Would you mind adding some javadoc strings for the public classes/interfaces/methods.

jedrazb:
Sure, added the javadocs

aprudhomme (approved):

PR #268, opened 2021-02-24
Make index data preload a config option

aprudhomme:
Allows specifying the `preloadIndexData` config value to disable preloading of index data into memory on startup. This could be useful for primaries, which would not be serving search queries.

For now this only applies to regular indices, not suggest (since I'm not sure if this kind of behavior is desired).

Preloading defaults to true to maintain the status quo.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/DirectoryFactory.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/DirectoryFactory.java
@@ -34,8 +34,13 @@ public abstract class DirectoryFactory {
   /** Sole constructor. */
   public DirectoryFactory() {}
 
-  /** Open a new {@link Directory} at the specified path. */
-  public abstract Directory open(Path path) throws IOException;
+  /**
+   * Open a new {@link Directory} at the specified path.
+   *
+   * @param path directory path
+   * @param preload if index data should be preloaded into memory (only for MMapDirectory type)
+   */
+  public abstract Directory open(Path path, boolean preload) throws IOException;
 
   /**
    * Returns an instance, using the specified implementation {FSDirectory, MMapDirectory,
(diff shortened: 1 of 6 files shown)


src/main/java/com/yelp/nrtsearch/server/luceneserver/ShardState.java
```diff
@@ -489,7 +490,7 @@ public synchronized void start() throws Exception {
     }
 
     boolean success = false;
-
+    boolean preloadData = indexState.globalState.configReader.getBoolean(PRELOAD_CONFIG_KEY, true);
```
sarthakn7, line 493: Should we read and store this config parameter in `LuceneServerConfiguration` only? That would make sure all the config options are in one place. We can expose either just preloadData or entire luceneserverconfiguration to ShardState via GlobalState

src/main/java/com/yelp/nrtsearch/server/luceneserver/DirectoryFactory.java
```diff
@@ -87,19 +92,32 @@ public Directory open(Path path) throws IOException {
...
       try {
         ctor = dirClass.getConstructor(Path.class);
-      } catch (NoSuchMethodException nsme) {
+      } catch (NoSuchMethodException ignored) {
+      }
+      try {
+        ctor = dirClass.getConstructor(Path.class, boolean.class);
+      } catch (NoSuchMethodException ignored) {
+      }
+      if (ctor == null) {
```
sarthakn7, line 104: I'm wondering if we should do away with this whole `else` block. I think it would be used only if someone decides to create a custom directory plugin. (Not a blocker for this PR)

aprudhomme: Yeah, if we really want to support this, we should probably add a plugin interface at some point.

sarthakn7 (approved):

PR #274, opened 2021-03-03
Force merge deletes

sarthakn7:
Exposed the [forceMergeDeletes method on IndexWriter](https://lucene.apache.org/core/8_4_0/core/org/apache/lucene/index/IndexWriter.html#forceMergeDeletes--) via an rpc. Also added cli commands for both forceMerge and forceMergeDeletes.


aprudhomme (approved):
LGTM, just need to rebuild the gateway.

PR #276, opened 2021-03-03
Add option to stream objects from s3 during download

aprudhomme:
Add option to stream objects from s3, instead of downloading them to a file first. Allows for concurrent downloading and decompressing/extracting. The download is still backed by a threadpool to allow parallel downloading of file parts.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/utils/ArchiverImpl.java
+++ b/src/main/java/com/yelp/nrtsearch/server/utils/ArchiverImpl.java
@@ -101,6 +101,11 @@ public class ArchiverImpl implements Archiver {
     this.downloadAsStream = downloadAsStream;
   }
 
+  public ArchiverImpl(
+      final AmazonS3 s3, final String bucketName, final Path archiverDirectory, final Tar tar) {
+    this(s3, bucketName, archiverDirectory, tar, false);
+  }
+
   @Override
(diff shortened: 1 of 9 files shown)


src/main/java/com/yelp/nrtsearch/server/utils/ArchiverImpl.java
```diff
@@ -69,20 +76,29 @@
...
+  private final boolean downloadAsStream;
 
   @Inject
   public ArchiverImpl(
-      final AmazonS3 s3, final String bucketName, final Path archiverDirectory, final Tar tar) {
+      final AmazonS3 s3,
+      final String bucketName,
+      final Path archiverDirectory,
+      final Tar tar,
+      final boolean downloadAsStream) {
```
umeshdangat, line 89: nit: you could make a new constructor and use chaining. 

`this(s3, bucketName, archiverDirectory, tar, false)`

That way the test code could still call the existing constructor and not change code in tests.

sarthakn7 (approved):

src/main/java/com/yelp/nrtsearch/server/utils/ArchiverImpl.java
aprudhomme: done

umeshdangat (approved):

PR #277, opened 2021-03-04
Add MatchAllDocsQuery to FunctionScoreQuery if its inner query isn't set

vim345:
I'm not too sure how an inner query cannot be set in practice to do an end-to-end test. This fixed is purely based on unit tests.
### Testing
```
./gradlew clean && ./gradlew installDist && ./gradlew test
```
Passed ✅

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
@@ -58,10 +58,6 @@ public class QueryNodeMapper {
               MatchOperator.MUST, BooleanClause.Occur.MUST));
 
   public Query getQuery(com.yelp.nrtsearch.server.grpc.Query query, IndexState state) {
-    if (query.getQueryNodeCase()
-        == com.yelp.nrtsearch.server.grpc.Query.QueryNodeCase.QUERYNODE_NOT_SET) {
-      return new MatchAllDocsQuery();
-    }
     Query queryNode = getQueryNode(query, state);
 
     if (query.getBoost() < 0) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
```diff
@@ -181,11 +181,17 @@ private FunctionScoreQuery getFunctionScoreQuery(
...
-    return new FunctionScoreQuery(
-        getQuery(functionScoreQuery.getQuery(), state),
-        scriptFactory.newFactory(params, state.docLookup));
+    com.yelp.nrtsearch.server.grpc.Query q = functionScoreQuery.getQuery();
+    Query query;
+    if (!q.toString().isEmpty()) {
+      query = getQuery(q, state);
+    } else {
+      query = new MatchAllDocsQuery();
+    }
```
aprudhomme, line 193: We have some similar logic done for the root query https://github.com/Yelp/nrtsearch/blob/master/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/SearchRequestProcessor.java#L254 
@sarthakn7 do you think maybe this should be the default behavior of the QueryNodeMapper whenever there is an UNSET query node?

sarthakn7: +1 Yeah we can make it the default behavior. That way this would be handled consistently for any other query that nests a query.

src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
```diff
@@ -58,6 +58,10 @@
               MatchOperator.MUST, BooleanClause.Occur.MUST));
 
   public Query getQuery(com.yelp.nrtsearch.server.grpc.Query query, IndexState state) {
+    if (query.getQueryNodeCase()
+        == com.yelp.nrtsearch.server.grpc.Query.QueryNodeCase.QUERYNODE_NOT_SET) {
```
aprudhomme, line 62: You could probably just add this case to the big switch statement in getQueryNode.

vim345: Thanks for the suggestion. Moved the logic.

aprudhomme (approved):

PR #279, opened 2021-03-05
Update documentation for topHits param

jedrazb:
Explain how to use `topHits` for pagination

### DIFF
--- a/clientlib/src/main/proto/yelp/nrtsearch/search.proto
+++ b/clientlib/src/main/proto/yelp/nrtsearch/search.proto
@@ -254,7 +254,7 @@ message Query {
 message SearchRequest {
     string indexName = 1; //index to search against
     int32 startHit = 2; // Which hit to start from (for pagination); default: 0
-    int32 topHits = 3; // How many top hits to retrieve; default: 10. For pagination: set it to startHit + window_size. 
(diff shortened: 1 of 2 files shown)


clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -254,7 +254,7 @@ message Query {
 message SearchRequest {
     string indexName = 1; //index to search against
     int32 startHit = 2; // Which hit to start from (for pagination); default: 0
-    int32 topHits = 3; // How many top hits to retrieve; default: 10
+    int32 topHits = 3; // How many top hits to retrieve; default: 10. For pagination: set it to startHit + window_size.
```
umeshdangat, line 257: I think you might want to rebuild the swagger spec to account for the updated comment? It will show up in swagger_ui

umeshdangat, line 257: Hmm, although I think its already broken since only the comments that go on a line before the field itself show up in swagger_UI
Screen Shot 2021-03-05 at 10 14 01 AM

jedrazb: ++ Regenerated grpcGateway (is it correct way to update the swagger schema)?

umeshdangat (approved):

PR #282, opened 2021-03-18
File-based custom grpc NameResolver

sarthakn7:
Created a grpc NameResolver which maps a file URI to Nrtsearch addresses. The file must contain the hosts and ports in json format. E.g. file contents:
```
[
  {"host":"10.20.31.28","port":8080},
  {"host":"10.20.41.29","port":8081},
  {"host":"10.20.51.30","port":8082}
]
```

The tests verify that the requests are load balanced across all servers, there are no errors when a server is closed and that updates to the file are applied to the channel.


umeshdangat (approved):
looks good !

PR #283, opened 2021-03-19
Added go package to proto files and changed grpc-gateway to fail build if the docker build fails

sarthakn7:
We were seeing build failures where the grpc-gateway build failed with:
```
protoc-gen-go: unable to determine Go import path for "yelp/nrtsearch/analysis.proto"

Please specify either:
        • a "go_package" option in the .proto source file, or
        • a "M" argument on the command line.

See https://developers.google.com/protocol-buffers/docs/reference/go-generated#package for more information.

--go_out: protoc-gen-go: Plugin failed with status code 1.
```

Added go package for all proto files to fix this. Not sure why it suddenly started failing, but I'm guessing that we haven't pinned a version somewhere. Also added a check to fail build if the docker build for grpc-gateway throws an error.

### DIFF
--- a/build_grpc_gateway.sh
+++ b/build_grpc_gateway.sh
@@ -1,10 +1,5 @@
+set -euxo pipefail
 docker build -t grpc-gateway -f grpc-gateway/Dockerfile .
-status=$?
-if [ $status -ne 0 ];
-then
-  echo "Unable to build grpc-gateway";
-  exit $status;
-fi
 docker run grpc-gateway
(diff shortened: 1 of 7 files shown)


taoyyu (approved):

build_grpc_gateway.sh
```diff
@@ -1,4 +1,10 @@
 docker build -t grpc-gateway -f grpc-gateway/Dockerfile .
```
aprudhomme, line 1: Could we just do something like `set -euxo pipefail` at the top to make any failed command fail the script?

aprudhomme (approved):

PR #284, opened 2021-03-23
Make random file send delay optional

aprudhomme:
Adds a config option to disable the random delay added between sending file chunks. This should hopefully speed up larger file transfers.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
+++ b/src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
@@ -1545,7 +1545,9 @@ public class LuceneServer {
                     .build();
             rawFileChunkStreamObserver.onNext(rawFileChunk);
             totalRead += chunkSize;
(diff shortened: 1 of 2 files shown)


sarthakn7 (approved):

src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
```diff
@@ -1545,7 +1545,9 @@ public void recvRawFile(
                     .build();
             rawFileChunkStreamObserver.onNext(rawFileChunk);
             totalRead += chunkSize;
-            randomDelay(random);
+            if (globalState.configuration.getFileSendDelay()) {
```
umeshdangat, line 1548: we should probably do a backfill of YelpReviewsDataset (or some other dataset) with the delay turned off. IIRC I saw this issue when backfilling an entire index with high throughput and primary buffer OOMed since replica could not keep up with primary producing data.

PR #285, opened 2021-03-23
Executing onCompleted for addDocuments and commit on separate threadpool

sarthakn7:
With this change the luceneserver grpc threads won't be blocked on long-running onCompleted (of addDocuments) and commit calls when there are merges happening. Reusing the indexing threadpool for this since addDocuments and commit are parts of indexing only. We can separately increase the size of this threadpool if needed.


umeshdangat (approved):
looks good to me . We will probably need to increase the number of threads for indexService given we moved work to that threadpool

PR #290, opened 2021-03-25
Made onCompletedForIndex in addDocuments serial and collecting futures for each index separately

sarthakn7:
Made onCompletedForIndex in addDocuments serial so that we don't deadlock in case we have lesser threads than the maximum parallel addDocuments calls. 
Also noticed that we were collecting the futures for all index in the same list and then blocking for the result from all indices in the first onCompletedForIndex call. Fixed this by collecting the futures separately for each index.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
+++ b/src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
@@ -19,6 +19,8 @@ import static com.yelp.nrtsearch.server.grpc.ReplicationServerClient.MAX_MESSAGE
 
 import com.google.api.HttpBody;
 import com.google.common.base.Splitter;
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.Multimap;
 import com.google.common.collect.Sets;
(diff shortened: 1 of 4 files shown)


src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
```diff
@@ -633,22 +635,27 @@ private String onCompletedForIndex(String indexName) {
...
+              // for remaining documents directly. This serializes indexing remaining documents for
+              // multiple indices but avoids deadlocking if there aren't more threads than the
+              // maximum
+              // number of parallel addDocuments calls.
+              long gen = new DocumentIndexer(globalState, addDocRequestList).runIndexingJob();
+              gens.add(gen);
             }
             // collect futures, block if needed
             PriorityQueue<Long> pq = new PriorityQueue<>(Collections.reverseOrder());
+            gens.forEach(pq::offer);
```
aprudhomme, line 655: Do we even need to use a priority queue here? All we care about is the highest gen, we could keep track of that directly.

aprudhomme (approved):

PR #291, opened 2021-03-26
Don't throw an error if there are no docs to index in the request

jedrazb:


src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
```diff
@@ -679,7 +679,7 @@ public void onCompleted() {
                 () -> {
                   try {
                     // TODO: this should return a map on index to genId in the response
-                    String genId = null;
+                    String genId = "";
```
aprudhomme, line 682: The genId is a stringified long, maybe default to "-1"?

sarthakn7 (approved):

Issue #232, opened 2020-12-23
Leftover directory after deleting index


sarthakn7:
Deleting an index and then recreating an index with the same name as the deleted index throws the following exception:
```
java.lang.IllegalArgumentException: rootDir "/nail/nrtsearch/primary_index/<index_name>" already exists
	at com.yelp.nrtsearch.server.luceneserver.GlobalState.createIndex(GlobalState.java:219)
	at com.yelp.nrtsearch.server.grpc.LuceneServer$LuceneServerImpl.createIndex(LuceneServer.java:303)
	at com.yelp.nrtsearch.server.grpc.LuceneServerGrpc$MethodHandlers.invoke(LuceneServerGrpc.java:2636)...
```
We need to make sure that the index directory is deleted when the index is deleted.

PR #298, opened 2021-04-20
Index's Root Directory removal on deleteIndex operation

fidato13:
Relates : https://github.com/Yelp/nrtsearch/issues/232

Changes included:
* New helper class (FileUtil) for filesystem operations
* ShardState utilizes FileUtil for deleting Shard dirs
* IndexState deletes the root index directory using FileUtil
* Assert for verifying the index root directory being deleted
* log message update for DeleteIndexHandler's reply
* Fix for Removing the old gens files (`indices.*`)
* Flush the updated globalState to disk once an index is deleted

### DIFF
--- a/src/test/java/com/yelp/nrtsearch/server/grpc/LuceneServerTest.java
+++ b/src/test/java/com/yelp/nrtsearch/server/grpc/LuceneServerTest.java
@@ -19,6 +19,8 @@ import static com.yelp.nrtsearch.server.grpc.GrpcServer.rmDir;
 import static org.junit.Assert.*;
 
 import com.google.api.HttpBody;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
 import com.google.protobuf.Empty;
 import com.yelp.nrtsearch.server.LuceneServerTestConfigurationFactory;
 import com.yelp.nrtsearch.server.config.LuceneServerConfiguration;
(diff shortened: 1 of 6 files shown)


fidato13:
@sarthakn7 
1. Noticed that the following code for removing the old global states (`indices.*`) doesn't work as intended:
https://github.com/Yelp/nrtsearch/blob/3b7864156cc3efd6eed6831f2a1c2ade8e6cbda1/src/main/java/com/yelp/nrtsearch/server/luceneserver/GlobalState.java#L179-L188

As `Path sub` is an absolute path and hence startsWith("indices.") check fails. Fixing this will sort out the issue regarding previous `indices.` files not being removed after a new index is created.

2. Also, it seems after the deletion of the index happens, the globalstate is not flushed to disk (?) . I have made a change to call saveIndexNames() after the deletion of the index. This should make sure that the latest `index.*` file always has the updated list of active indexes.

Do you think we can address these globalstate issue in this PR ?, else I can create a separate issue and PR to fix the above.

jedrazb (approved):

jedrazb:
Let's wait for the ship-it from @sarthakn7

sarthakn7:
> Do you think we can address this globalstate issue in this PR ?, else I can create a separate issue (if there isn't already) and a PR to fix the above.

Sure it is good to address this issue as well. Thanks for looking into this!

src/test/java/com/yelp/nrtsearch/server/grpc/LuceneServerTest.java
```diff
@@ -604,15 +608,93 @@ public void testDeleteIndex() throws IOException, InterruptedException {
...
+    assertEquals(
+        "indices.0",
+        Files.list(globalState.getStateDir()).findAny().get().getFileName().toString());
+
+    Path primaryStateIndexPath = globalState.getStateDir().resolve("indices.0");
+    JsonObject persistedIndexNames =
+        JsonParser.parseString(Files.readString(primaryStateIndexPath)).getAsJsonObject();
+
+    assertEquals(
+        globalState.getIndexNames(), Collections.unmodifiableSet(persistedIndexNames.keySet()));
```
sarthakn7, line 656: Do we need to call `Collections.unmodifiableSet` on the set before asserting?

fidato13: Good catch, we don't, above statements are now updated.

sarthakn7 (approved):
Lgtm. Please try to avoid force-pushing after making review changes as that makes it harder for reviewers to see the changes.

fidato13:
> Lgtm. Please try to avoid force-pushing after making review changes as that makes it harder for reviewers to see the changes.

Thanks, makes sense, will make sure of that in future.

PR #306, opened 2021-04-29
Call fetch task for hits grouped by leaves and in order of doc ids

sarthakn7:
Jprofiling shows that creating a scorer for every hit in the fetch task is expensive especially when we have a large number of top hits. This change should allow the fetch task to create one scorer per leaf instead of creating one per hit.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,7 +20,7 @@ sourceCompatibility = 1.14
 targetCompatibility = 1.14
 
 allprojects {
-    version = '0.6.12'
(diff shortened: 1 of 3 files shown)


sarthakn7:
I'm wondering if it would be cleaner to have a `processHitsInLeaf` method in `FetchTask`? Or would it be sufficient to do this and just add in the comment on the `processHit` method that this would be called grouped by leaves and in order of the doc ids?

aprudhomme (approved):

PR #312, opened 2021-05-05
optimize fields retrieval

taoyyu:
This code has some placeholder code for highlights logic in the future. I didn't remove them. Let me know if you want me to delete these code.

There is no test for this code. But this code is actually used by every tests which I guess it is enough.

I'll profile it to check perf locally again

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/SearchHandler.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/SearchHandler.java
@@ -182,12 +182,12 @@ public class SearchHandler implements Handler<SearchRequest, SearchResponse> {
       hitBuilders.sort(Comparator.comparing(Hit.Builder::getLuceneDocId));
       List<LeafReaderContext> leaves = s.searcher.getIndexReader().leaves();
 
-      Map<Integer, LeafReaderContext> hitIdToLeaves = new HashMap<>();
+      List<LeafReaderContext> hitIdToLeaves = new ArrayList<>();
       for (int hitIndex = 0; hitIndex < hitBuilders.size(); ++hitIndex) {
         var hitResponse = hitBuilders.get(hitIndex);
         LeafReaderContext leaf =
             leaves.get(ReaderUtil.subIndex(hitResponse.getLuceneDocId(), leaves));
-        hitIdToLeaves.put(hitIndex, leaf);
+        hitIdToLeaves.add(hitIndex, leaf);
       }
 
       for (String field : searchContext.getRetrieveFields().keySet()) {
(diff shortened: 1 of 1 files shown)


taoyyu:
hm, test failed but passed locally. I'll pull master and fix it

src/main/java/com/yelp/nrtsearch/server/luceneserver/doc/LoadedDocValues.java
```diff
@@ -354,14 +355,19 @@ public ObjectJsonDocValues(BinaryDocValues docValues) {
...
+                .map(e -> e.getStructValue())
+                .collect(Collectors.toList());
       } else {
         value = null;
       }
     }
 
     @Override
-    public Map<String, Object> get(int index) {
+    public Struct get(int index) {
```
aprudhomme, line 370: Are the custom scorers using this going to be ok with using a `Struct`?

src/main/java/com/yelp/nrtsearch/server/luceneserver/SearchHandler.java
```diff
@@ -191,27 +190,54 @@ public SearchResponse handle(IndexState indexState, SearchRequest searchRequest)
...
 
-        if (!searchContext.getRetrieveFields().isEmpty()) {
-          var fieldValueMap =
-              fillFields(
+      for (String field : searchContext.getRetrieveFields().keySet()) {
+        for (int hitIndex = 0; hitIndex < hitBuilders.size(); ++hitIndex) {
+          var hitResponse = hitBuilders.get(hitIndex);
+          LeafReaderContext leaf = hitIdToLeaves.get(hitIndex);
+          CompositeFieldValue v =
+              getFieldForHit(
```
aprudhomme, line 206: I guess you can try this approach to see if it gives you enough perf improvement. This could be made better by the reuse of per segment resources.

src/main/java/com/yelp/nrtsearch/server/luceneserver/doc/LoadedDocValues.java
taoyyu: This is object field. It should not be used by any scorer. For scoring, we will replace with maptype. Object is only used for retrieval

src/main/java/com/yelp/nrtsearch/server/luceneserver/SearchHandler.java
taoyyu: I am not sure how to reuse the segment here.
The doc value field is column based. So it is recommended to iterate by fields then id. but segment is by id. I checked leafreader source code. There are a ton of implementations and leafreader instance in leafreader classes. not sure if that will help

aprudhomme: The idea would be to create the per segment object (ScoreScript, LoadedDocValues<>, etc) once, then advance the doc id for subsequent segment documents.

If you want, you can try your current approach first, since it is a more straightforward change.

taoyyu: ack, let's try this first. I'll have follow-up prs since this optimization actually make stored fields slower since that is row based. I'll have follow-up prs to treat them differently

aprudhomme (approved):

PR #313, opened 2021-05-11
parallel fill fields for nrtsearch

taoyyu:
use multiple threads to do doc values lookup here to see perf. This feature is disabled by default. And we only use threadpool when both number of fields to retrieve and number of hits to retrieve pass the minimum requirements(default minimum is 20 fields, 50hits)

Currently this thread pool only does fill fields. Hopefully it can be extended do all fetch phase tasks like score components.

We may even need consider thread pool for rescore once we have ads l2...


aprudhomme (approved):
Looks ok to try.

PR #316, opened 2021-05-19
fix parallelism for fields retrieval

taoyyu:
This is a bug. 

The second parameter for Lists.partition should be size, not number of lists


umeshdangat:
lgtm !

aprudhomme (approved):

PR #317, opened 2021-05-20
Add reflection API

jedrazb:
Enabling reflection API on the server, steps taken from [here](https://github.com/grpc/grpc-java/blob/master/documentation/server-reflection-tutorial.md#:~:text=To%20enable%20server%20reflection%2C%20you,%2D%2D%2D%20a%2Fexamples%2Fbuild.)

Reflection API is pretty useful for tooling, since it allows for discovery of endpoints and schemas on the server without the need for proto files.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -81,6 +81,8 @@ dependencies {
     implementation 'info.picocli:picocli:4.0.4'
     implementation 'org.apache.commons:commons-csv:1.7'
 
+    implementation "io.grpc:grpc-services:1.30.0"
+
     testImplementation "junit:junit:4.12"
     testImplementation "org.mockito:mockito-core:2.25.1"
     testImplementation "org.apache.lucene:lucene-test-framework:${luceneVersion}"
--- a/build.gradle
+++ b/build.gradle
@@ -31,6 +31,7 @@ System.setProperty("org.gradle.internal.publish.checksums.insecure", "True")
 
 def luceneVersion = '8.4.0'
 project.ext.slf4jVersion = '2.0.0-alpha0'
+project.ext.grpcVersion = '1.30.0'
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -81,6 +81,8 @@ dependencies {
     implementation 'info.picocli:picocli:4.0.4'
     implementation 'org.apache.commons:commons-csv:1.7'
 
+    implementation "io.grpc:grpc-services:1.30.0"
```
sarthakn7, line 84: Can we somehow use the same version variable here and in https://github.com/Yelp/nrtsearch/blob/master/clientlib/build.gradle#L30 ? I think you can do it by making it a project level variable

aprudhomme, line 84: I assume this should be the same grpc version used in the clientlib. Can we make the version a property in the root project and reference it in both places?

jedrazb:
Added projectl-level grpc version

clientlib/build.gradle
```diff
@@ -27,15 +27,14 @@ startScripts.enabled = false
 def _artifactId = 'clientlib'
 
 // Dependency versions
-def grpcVersion = '1.30.0' // CURRENT_GRPC_VERSION
```
sarthakn7, line 30: There are 2 dependencies which are still using this variable

jedrazb: Good catch, no idea how could I miss it

sarthakn7 (approved):

jedrazb:
Build failing due to docker pull rate limit: 
```
Step 1/30 : FROM ubuntu:20.04
toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
```

PR #319, opened 2021-05-26
add new endpoint that returns Any

taoyyu:
add searchV2 that return Any


clientlib/src/main/proto/yelp/nrtsearch/luceneserver.proto
```diff
@@ -128,6 +129,15 @@ service LuceneServer {
       body: "*"
     };
     }
+
+    /* Search */
+    rpc searchV2 (SearchRequest) returns (google.protobuf.Any) {
+        option (google.api.http) = {
+      post: "/v1/search"
```
aprudhomme, line 136: I suppose the post endpoint should also be v2.

aprudhomme (approved):

PR #326, opened 2021-06-02
Add option to sync nrt point on index start

aprudhomme:
Add the option to sync an initial nrt point while starting the index. This is a best effort operation, if the primary is unavailable for too long (30s), the sync will be skipped.

Once the copy job begins, the sync operation will wait until either there is a new index version, or there is no longer an active copy job.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/ShardState.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/ShardState.java
@@ -173,6 +173,7 @@ public class ShardState implements Closeable {
   private KeepAlive keepAlive;
   // is this shard restored
   private boolean restored;
+  private volatile boolean started = false;
 
   /** Restarts the reopen thread (called when the live settings have changed). */
   public void restartReopenThread() {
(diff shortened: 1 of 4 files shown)


src/main/java/com/yelp/nrtsearch/server/luceneserver/ShardState.java
```diff
@@ -933,6 +934,10 @@ public synchronized void startReplica(
               verbose ? System.out : new PrintStream(OutputStream.nullOutputStream()),
               primaryGen);
 
+      if (indexState.globalState.configuration.getSyncInitialNrtPoint()) {
+        nrtReplicaNode.syncFromCurrentPrimary(INITIAL_SYNC_PRIMARY_WAIT_MS);
+      }
```
sarthakn7, line 939: In `ShardState.isStarted()` which is used by `IndexState.isStarted()` we consider the shard to be ready if `nrtReplicaNode != null`: https://github.com/Yelp/nrtsearch/blob/master/src/main/java/com/yelp/nrtsearch/server/luceneserver/ShardState.java#L209
What do you think about assigning the `NRTReplicaNode` to a local variable, syncing the nrt point and then assigning it to the global variable so that we consider a shard ready after the sync is complete? Not sure if syncing without that global assignment causes any other issues though

aprudhomme: I added an explicit `started` flag, so it is no longer a surprise.

sarthakn7 (approved):

PR #330, opened 2021-06-08
Allow environment vars in host name config

aprudhomme:
Allow use of environment variables in the host name config. Substrings of the form `${VAR}` will be replace with their value from the environment. If the variable is not set, it is replaced with an empty string.

### DIFF
--- a/src/test/java/com/yelp/nrtsearch/server/config/LuceneServerConfigurationTest.java
+++ b/src/test/java/com/yelp/nrtsearch/server/config/LuceneServerConfigurationTest.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2020 Yelp Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


src/test/java/com/yelp/nrtsearch/server/config/LuceneServerConfigurationTest.java
```diff
@@ -0,0 +1,81 @@
+/*
+ * Copyright 2020 Yelp Inc.
```
sarthakn7, line 2: I noticed that spotless changes the year to 2020. Is that fine?

sarthakn7 (approved):

src/test/java/com/yelp/nrtsearch/server/config/LuceneServerConfigurationTest.java
aprudhomme: Would be good to create an issue for this. I think updating the template will change all of the header. Maybe that is what we want, but I'm not sure.

PR #340, opened 2021-07-05
Fix deadlock conditions between indexState and shardState

jedrazb:
## Problem

This aims to fix deadlock conditions between `indexState` and `shardState`. We encountered a deadlock between the
call of `synchronized commit` (called from indexState) and `synchronized startPrimary` (called from shard state). The deadlock happened because when calling startPrimary from shard state we were also making call to:
- `synchronized getBooleanSetting` - a method of IndexState

Since synchronized methods have one monitor per object instance it can lead to deadlock like this:

Screenshot 2021-07-05 at 16 45 08

## Solution

I verified that `getBooleanSetting` is the only synchronized method of IndexState which is called by instance of ShardState.

I moved away the save settings to separate object with granular read/write locks to avoid the need to use synchronized methods. 

It should fix the deadlock scenario arising from coupling indexstate and shardstate objects with synchronized methods.


src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java
```diff
@@ -824,9 +814,9 @@ public static boolean isChildName(String name) {
    * Live setting: set the mininum refresh time (seconds), which is the longest amount of time a
    * client may wait for a searcher to reopen.
    */
-  public synchronized void setMinRefreshSec(double min) {
```
aprudhomme, line 827: The set methods will still need to be synchronized. Otherwise, there is a race condition and the local class members may have a different value than what is in the json state.

src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java
```diff
@@ -1252,19 +1254,21 @@ public synchronized MergeThread getMergeThread(
   }
 
   synchronized boolean getBooleanSetting(String name, boolean val) {
```
aprudhomme, line 1256: The synchronization on these get methods will need to be removed.

The locking order is still
commit: IndexState -> ShardState
startPrimary: ShardState -> IndexState
It is the reversal of lock ordering that leaves the system open to deadlock.

jedrazb: Alright, removed the synchronization of the "get setting" methods

src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java
```diff
@@ -1251,20 +1253,22 @@ public synchronized MergeThread getMergeThread(
     return iwc;
   }
 
-  synchronized boolean getBooleanSetting(String name, boolean val) {
-    return settingsSaveState.get(name) == null ? val : settingsSaveState.get(name).getAsBoolean();
+  boolean getBooleanSetting(String name, boolean val) {
+    ThreadSafeJSONObject settings = saveState.getSettings();
+    return settings.get(name) == null ? val : settings.get(name).getAsBoolean();
```
aprudhomme, line 1258: You will only want to get() the value once and use it twice, otherwise it may change between accesses.

src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java
```diff
@@ -1251,20 +1253,22 @@ public synchronized MergeThread getMergeThread(
...
   }
 
-  synchronized double getDoubleSetting(String name, double val) {
-    return settingsSaveState.get(name) == null ? val : settingsSaveState.get(name).getAsDouble();
+  double getDoubleSetting(String name, double val) {
+    ThreadSafeJSONObject settings = saveState.getSettings();
+    return settings.get(name) == null ? val : settings.get(name).getAsDouble();
```
aprudhomme, line 1263: Same here.

aprudhomme (approved):

PR #344, opened 2021-07-13
Add periodic commits 

jedrazb:
Add periodic commits for all shards.

This can be configured in lucene server configuration with `periodicCommitDelaySec`

## Validation

Set `periodicCommitDelaySec` to `2` and could see commit taking place every 2 seconds after starting the index. Stopping the index cancelled the commit task gracefully.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java
@@ -154,7 +154,7 @@ public class IndexState implements Closeable, Restorable {
   private ExecutorService fetchThreadPoolExecutor;
   private IdFieldDef idFieldDef = null;
   private Warmer warmer = null;
-  private final PeriodicCommit periodicCommit;
+  private PeriodicCommit periodicCommit = null;
 
   public ShardState addShard(int shardOrd, boolean doCreate) {
     if (shards.containsKey(shardOrd)) {
@@ -513,8 +513,10 @@ public class IndexState implements Closeable, Restorable {
     }
     searchThreadPoolExecutor = globalState.getSearchThreadPoolExecutor();
     fetchThreadPoolExecutor = globalState.getFetchService();
-    periodicCommit = new PeriodicCommit(this);
-    periodicCommit.schedule();
+    if (isPeriodicCommitEnabled()) {
+      periodicCommit = new PeriodicCommit(this);
+      periodicCommit.schedule(globalState.configuration.getPeriodicCommitDelaySec());
+    }
   }
 
   void initSaveLoadState() throws IOException {
@@ -644,7 +646,9 @@ public class IndexState implements Closeable, Restorable {
   @Override
   public void close() throws IOException {
     logger.info(String.format("IndexState.close name= %s", name));
-    periodicCommit.cancel();
+    if (isPeriodicCommitEnabled()) {
+      periodicCommit.cancel();
+    }
     List<Closeable> closeables = new ArrayList<>();
     closeables.addAll(shards.values());
     closeables.addAll(fields.values());
@@ -1420,6 +1424,11 @@ public class IndexState implements Closeable, Restorable {
     }
 
     return result;
+  }
+
+  protected boolean isPeriodicCommitEnabled() {
+    return globalState.configuration.getPeriodicCommitDelaySec()
+        != LuceneServerConfiguration.PERIODIC_COMMIT_OFF;
   }
 
   /**
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java
@@ -160,6 +160,7 @@ public class IndexState implements Closeable, Restorable {
   private ExecutorService fetchThreadPoolExecutor;
   private IdFieldDef idFieldDef = null;
   private Warmer warmer = null;
+  private final PeriodicCommit periodicCommit;
 
   public ShardState addShard(int shardOrd, boolean doCreate) {
     if (shards.containsKey(shardOrd)) {
@@ -518,6 +519,8 @@ public class IndexState implements Closeable, Restorable {
     }
     searchThreadPoolExecutor = globalState.getSearchThreadPoolExecutor();
     fetchThreadPoolExecutor = globalState.getFetchService();
+    periodicCommit = new PeriodicCommit(this);
+    periodicCommit.schedule();
   }
 
   void initSaveLoadState() throws IOException {
@@ -647,6 +650,7 @@ public class IndexState implements Closeable, Restorable {
   @Override
   public void close() throws IOException {
     logger.info(String.format("IndexState.close name= %s", name));
+    periodicCommit.cancel();
     List<Closeable> closeables = new ArrayList<>();
     closeables.addAll(shards.values());
     closeables.addAll(fields.values());
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java
@@ -125,7 +125,7 @@ public class IndexState implements Closeable, Restorable {
   public static final int DEFAULT_SLICE_MAX_DOCS = 250_000;
   public static final int DEFAULT_SLICE_MAX_SEGMENTS = 5;
 
-  Logger logger = LoggerFactory.getLogger(IndexState.class);
+  private static final Logger logger = LoggerFactory.getLogger(IndexState.class);
   public final GlobalState globalState;
 
   /** Which norms format to use for all indexed fields. */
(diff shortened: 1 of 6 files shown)


src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java
```diff
@@ -518,6 +519,8 @@ public IndexState(
     }
     searchThreadPoolExecutor = globalState.getSearchThreadPoolExecutor();
     fetchThreadPoolExecutor = globalState.getFetchService();
+    periodicCommit = new PeriodicCommit(this);
```
sarthakn7, line 522: Can we just set this to null and not create any objects if the `periodicCommitDelaySec` is set to -1?

sarthakn7 (commented):
It would be good if we can add tests for this as well

src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java
```diff
@@ -641,6 +646,9 @@ public synchronized void load(JsonObject jsonObject) throws IOException {
   @Override
   public void close() throws IOException {
     logger.info(String.format("IndexState.close name= %s", name));
+    if (isPeriodicCommitEnabled()) {
+      periodicCommit.cancel();
+    }
```
aprudhomme, line 651: Might be better to just make the class extend closable and add it to the closables list.

src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java
```diff
@@ -641,6 +646,9 @@ public synchronized void load(JsonObject jsonObject) throws IOException {
   @Override
   public void close() throws IOException {
     logger.info(String.format("IndexState.close name= %s", name));
+    if (isPeriodicCommitEnabled()) {
```
aprudhomme, line 649: I would probably just do a null check, since that is what really matters.

src/main/java/com/yelp/nrtsearch/server/luceneserver/PeriodicCommit.java
```diff
@@ -0,0 +1,59 @@
...
+
+  public PeriodicCommit(IndexState indexState) {
+    this.indexState = indexState;
+    timer = new Timer();
+    periodicCommitTask =
+        new TimerTask() {
+          @Override
+          public void run() {
+            try {
+              indexState.commit();
```
aprudhomme, line 41: The ShardState `commit()` is currently a noop on replicas. It will need to be modified to invoke `nrtReplicaNode.commit()` for this to have the desired effect.

src/main/java/com/yelp/nrtsearch/server/luceneserver/PeriodicCommit.java
```diff
@@ -0,0 +1,59 @@
...
+import com.yelp.nrtsearch.server.config.LuceneServerConfiguration;
+import java.io.IOException;
+import java.util.Timer;
+import java.util.TimerTask;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class PeriodicCommit {
+
+  Logger logger = LoggerFactory.getLogger(IndexState.class);
```
aprudhomme, line 27: This should be `private static final` and reference the current class. Feel free to fix the one in `IndexState` too.

src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java
sarthakn7: I think we need to do this since we can have the periodic commit disabled.

src/main/java/com/yelp/nrtsearch/server/luceneserver/PeriodicCommit.java
sarthakn7: ++ for catching this!

src/main/java/com/yelp/nrtsearch/server/luceneserver/IndexState.java
sarthakn7: Might be tricky since we want to stop the thread before the shard is closed (to not call commit after the shard is closed).

aprudhomme: The closables list is processed in order, so this could be done first. But there should be a comment explaining the ordering dependency.

src/main/java/com/yelp/nrtsearch/server/luceneserver/PeriodicCommit.java
jedrazb: ++, I have some issues with failing tests after adding the commit to the replica:

I'm defining commit in `ShardState` in the following way:
```
  public synchronized long commit() throws IOException {

    long gen;

    if (writer != null) {
      // nocommit: two phase commit?
      if (taxoWriter != null) {
        taxoWriter.commit();
      }
      gen = writer.commit();
    } else {
      if (nrtReplicaNode != null) {
        nrtReplicaNode.commit();
      }
      gen = -1;
    }

    return gen;
  }
```

When running tests I have some issues with `ReplicationTestFailureScenarios` test cases, the searcher version is expired:
```
com.yelp.nrtsearch.server.grpc.ReplicationTestFailureScenarios > primaryDurabilitySyncWithReplica FAILED                                                                                             
    io.grpc.StatusRuntimeException: UNKNOWN: error while trying to execute search for index test_index. check logs for full searchRequest.                                                           
    searcher: This searcher has expired version=6 vs currentVersion=10                                                                                                                               
        at io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:244)                                                                                                                   
        at io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:225)                                                                                                                               
        at io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:142)                                                                                                                          
        at com.yelp.nrtsearch.server.grpc.LuceneServerGrpc$LuceneServerBlockingStub.search(LuceneServerGrpc.java:2239)                                                                               
        at com.yelp.nrtsearch.server.grpc.ReplicationTestFailureScenarios.publishNRTAndValidateSearchResults(ReplicationTestFailureScenarios.java:453)                                               
        at com.yelp.nrtsearch.server.grpc.ReplicationTestFailureScenarios.primaryDurabilitySyncWithReplica(ReplicationTestFailureScenarios.java:354)                                                 

com.yelp.nrtsearch.server.grpc.ReplicationTestFailureScenarios > primaryStoppedAndRestartedWithNoPreviousIndex FAILED
    io.grpc.StatusRuntimeException: UNKNOWN: error while trying to execute search for index test_index. check logs for full searchRequest.
    searcher: This searcher has expired version=3 vs currentVersion=5
        at io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:244)
        at io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:225)
        at io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:142)
        at com.yelp.nrtsearch.server.grpc.LuceneServerGrpc$LuceneServerBlockingStub.search(LuceneServerGrpc.java:2239)
        at com.yelp.nrtsearch.server.grpc.ReplicationTestFailureScenarios.publishNRTAndValidateSearchResults(ReplicationTestFailureScenarios.java:453)
        at com.yelp.nrtsearch.server.grpc.ReplicationTestFailureScenarios.primaryStoppedAndRestartedWithNoPreviousIndex(ReplicationTestFailureScenarios.java:243)
 ```
 
It is probably cause because now `IndexState.commit()` also commits on replicas.

I can't tell myself if this is a desired behaviour, any thoughts on that @sarthakn7 @aprudhomme?

aprudhomme: hmm, that is interesting. Those test are around recovery when switching to a new primary generation. You may need to turn on the nrt debug messages and try to figure out what is going on for those test. It might be that there are issues with committing index data on the replica that is not (necessarily) durable on the primary, or maybe there is something else we should be doing when syncing an nrtpoint.

jedrazb: Getting this output with debug messages: https://pastebin.com/STX8QncY

aprudhomme: I think the issue is that replica can commit an index version not yet committed on the primary. The thread approach probably won't work for this situation.

This could maybe be done by having the primary communicate its last committed version as part of the nrtpoint info, so the replica would know when it was safe to commit its local index. It might roughly look something like this https://github.com/aprudhomme/nrtsearch/commit/e30f5798c597a2ae7d6043408c6e67220f2a822e

sarthakn7:
This won't work since `replica can commit an index version not yet committed on the primary`. Closing this PR, and the underlying issue of replicas retaining segments from the backup will be handled differently.

PR #345, opened 2021-07-14
Backup warming queries num queries threshold

erikyangs:


clientlib/src/main/proto/yelp/nrtsearch/luceneserver.proto
```diff
@@ -716,6 +716,7 @@ message BackupIndexResponse {
 message BackupWarmingQueriesRequest {
     string index = 1; // Index whose warming queries to backup
     string serviceName = 2; // remote storage namespace qualifier for service
+    int32 numQueriesThreshold = 3; // optional; minimum # of queries required to backup warming queries
```
sarthakn7, line 719: Need to regenerate the rest gateway after adding this

src/main/java/com/yelp/nrtsearch/server/cli/BackupWarmingQueriesCommand.java
```diff
@@ -46,11 +46,20 @@ public String getServiceName() {
...
+  public int getNumQueriesThreshold() {
+    return numQueriesThreshold;
+  }
+
   @Override
   public Integer call() throws Exception {
     LuceneServerClient client = baseCmd.getClient();
     try {
-      client.backupWarmingQueries(getIndex(), getServiceName());
+      client.backupWarmingQueries(getIndex(), getServiceName(), getNumQueriesThreshold());
```
sarthakn7, line 62: Can we add a uptime threshold as well? So we can fail if say the node hasn't been running for at least 3 hours

src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
```diff
@@ -1405,6 +1406,41 @@ public void backupWarmingQueries(
...
+        if (uptimeMinutesThreshold > 0 && currUptimeMinutes < uptimeMinutesThreshold) {
+          logger.warn(
+              "Unable to backup warming queries since uptime is {} minutes, which is less than threshold {}",
+              currUptimeMinutes,
+              uptimeMinutesThreshold);
+          responseObserver.onError(
+              Status.UNKNOWN
+                  .withDescription(
+                      String.format(
+                          "Unable to backup warming queries since uptime is {} minutes, which is less than threshold {}",
```
sarthakn7, line 1438: You need to provide `%s` as the the placeholder instead of `{}` in `String.format`

sarthakn7 (commented):
Can we also add tests? Just need to add 2 failing cases based on `testAddSearchRequest_backupWarmingQueriesToS3`

sarthakn7 (approved):

PR #349, opened 2021-07-23
Added option to delete existing data when restoring an index from backup

sarthakn7:
If the option is enabled we won't throw an error if index data was present when restarting a crashed node.


src/main/java/com/yelp/nrtsearch/server/luceneserver/StartIndexHandler.java
```diff
@@ -65,6 +74,9 @@ public StartIndexResponse handle(IndexState indexState, StartIndexRequest startI
           } else {
             throw new IllegalStateException("Index " + indexState.name + " already restored");
           }
+        } catch (IOException e) {
+          logger.info("Unable to delete existing index data", e);
+          throw new StartIndexHandlerException(e);
         } finally {
           shardState.setRestored(true);
         }
```
aprudhomme, line 82: Do we still want this finally block to run if an exception is thrown?

sarthakn7: @aprudhomme I am actually not sure what purpose that line is serving. It essentially blocks an index from being restored until it is deleted and recreated if there is any error. Should we just remove it?
Also cc: @umeshdangat for any additional context here

aprudhomme: I don't have as much context on this code, but I would think you would want to remove the finally block and set restored true right after the artifact download above.

sarthakn7: Made the change, can revisit removing or modifying this check later

aprudhomme (approved):

PR #350, opened 2021-07-30
Fix delete index data restore

sarthakn7:
#349 enabled deleting existing index data before doing a restore, but it didn't work correctly if the existing index was started from a backup. This was because an index started from backup doesn't have the files in the root directory, but the root directory is pointed to the backup. So the backup directory was not deleted, and the Archiver determined that it didn't need to download the backup since the directory was already present.
This PR fixes the issue by deleting the data directory for the index in Archiver directory.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/StartIndexHandler.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/StartIndexHandler.java
@@ -23,6 +23,9 @@ import com.yelp.nrtsearch.server.grpc.StartIndexResponse;
 import com.yelp.nrtsearch.server.utils.Archiver;
 import java.io.IOException;
 import java.nio.file.Path;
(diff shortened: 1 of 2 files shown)


src/main/java/com/yelp/nrtsearch/server/luceneserver/StartIndexHandler.java
```diff
@@ -61,11 +63,12 @@ public StartIndexResponse handle(IndexState indexState, StartIndexRequest startI
       synchronized (shardState) {
         try {
           if (!shardState.isRestored()) {
+            RestoreIndex restoreIndex = startIndexRequest.getRestore();
             if (startIndexRequest.getRestore().getDeleteExistingData()) {
```
aprudhomme, line 67: nit: if you are going to move the local variable, you might as well use it on the next line.

aprudhomme (approved):

PR #352, opened 2021-08-03
Adding rescore time per rescorer to diagnostics

desiam:
Adding `rescorersTimeMs` to the diagnostics section of the `SearchResponse`. This enables a breakdown of times for each individual rescorer rather than a single duration for all rescorers.

Notes
* Using default `rescorer_%d` if rescorer `name` is not passed into the request
* Assumes that each rescorer name is unique
* **Deprecates `rescoreTimeMs`**, but can undo if this field is useful even with the introduction of `rescorersTimeMs`

### Example query/response
Query
```
{
        "indexName": "testIdx",
        "startHit": 0,
        "topHits": 100,
        "retrieveFields": ["vendor_name"],
        "rescorers": [{
                "name": "first_rescorer",
                "queryRescorer": {
                        "rescoreQuery": {}
                }
        },
        {
                "name": "second_rescorer",
                "queryRescorer": {
                        "rescoreQuery": {}
                }

        }]
}
```
Response
```
[INFO ] 2021-08-05 12:35:13.349 [main] LuceneServerClient - Server returned : diagnostics {
  firstPassSearchTimeMs: 0.224499
  getFieldsTimeMs: 0.040539
  rescoreTimeMs: 1.331563
  rescorersTimeMs {
    key: "first_rescorer"
    value: 1.273288
  }
  rescorersTimeMs {
    key: "second_rescorer"
    value: 0.00945
  }
}
totalHits {
  value: 2
}
searchState {
}
```


clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -392,7 +392,8 @@ message SearchResponse {
         double newSnapshotSearcherOpenMs = 7;
         double nrtWaitTimeMs = 8;
         map<string, double> facetTimeMs = 9;
-        double rescoreTimeMs = 10;
+        double rescoreTimeMs = 10 [deprecated = true];
```
taoyyu, line 395: not sure about deprecated = true here. Does this affect anything? We are still returning rescoreTimeMs right?

taoyyu (commented):
looks sane to me. Wondering if you have tested locally. Can you add an example response here?

clientlib/src/main/proto/yelp/nrtsearch/search.proto
desiam: Reverted change. We will support both `rescoreTimeMs` and `rescorersTimeMs` for now.

clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -492,7 +493,7 @@ message QueryRescorer {
     Query rescoreQuery = 1;
     double queryWeight = 2;
     double rescoreQueryWeight = 3;
-
+    string name = 4;
```
aprudhomme, line 496: Maybe the name could be added to the `Rescorer` message. This would avoid collisions if the same plugin rescorer gets used multiple times.

desiam: Fields has been moved to the `Rescorer` message

aprudhomme (approved):

taoyyu (approved):

PR #372, opened 2021-09-03
Fix race condition in gateway docker build script

aprudhomme:
The grpc-gateway build script runs a series of docker commands to copy files out of the gateway container. These commands get the container id of the last container executed. The is a race condition, and will fail if another container is run between the command executions.

Fixed script to save container id from the run command, and use that id for all subsequent commands.

Also now removes the container when the command finishes.


sarthakn7 (approved):
++ for fixing this!

PR #377, opened 2021-09-23
include building grpc gateway step on readme

lmenezes:
Hi there,

following the steps on the readme file to run the example I realized that building the gRPC gateway was not included. Just added that.

Thanks for sharing :)


sarthakn7 (approved):
Thanks for adding this!

PR #379, opened 2021-09-29
Added some query documentation

sarthakn7:
Adding some documentation for all types of queries in Nrtsearch.


docs/analysis.rst
```diff
@@ -0,0 +1,38 @@
...
+      repeated NameAndParams tokenFilters = 2;
+  }
+
+  // Used to be able to check if a value was set
+  message IntObject {
+      int32 int = 1;
+  }
+
+  message CustomAnalyzer {
+      repeated NameAndParams charFilters = 1; // Available char filters as of Lucene 8.2.0: htmlstrip, mapping, persian, patternreplace
```
vim345, line 24: +1 on persian 😄

docs/queries/disjunction_max.rst
```diff
@@ -0,0 +1,13 @@
+Disjunction Max Query
+==========================
+
+A query that generates the union of documents produced by its subqueries, and that scores each document with the maximum score for that document as produced by any subquery, plus a tie breaking increment for any additional matching subqueries. This is useful when searching for a word in multiple fields with different boost factors (so that the fields cannot be combined equivalently into a single search field). This maps to DisjunctionMaxQuery in Lucene.
```
vim345, line 4: I think creating an example of putting data in Nrtserach and then searching them using this query would be easier to understand than just the pure description. This can be done for the queries that are as straightforward as some others such as `exists`. We can do that in a separate PR.

docs/queries/function_filter.rst
```diff
@@ -0,0 +1,49 @@
...
+Proto definition:
+
+.. code-block::
+
+   message FunctionFilterQuery {
+       Script script = 1; // script definition to compute a custom document score
+   }
+
+   message Script {
+       string lang = 1; // script language
```
vim345, line 15: If I'm not mistaken core nrtsearch doesn't support any scripting language out of the box and these languages can be supported using custom plugins. Maybe we can give some examples such as `painless`, `groovy` and mention that they should be enabled separately.

vim345 (approved):

docs/queries/disjunction_max.rst
sarthakn7: Yes we should do that, I'm only making it easier to understand query syntax with this PR.

docs/queries/function_filter.rst
sarthakn7: Nrtsearch does support Javascript expressions https://github.com/Yelp/nrtsearch/blob/master/src/main/java/com/yelp/nrtsearch/server/luceneserver/script/js/JsScriptEngine.java
Will make it clearer as we improve the documentation

Issue #382, opened 2021-10-16
Tests Fail because of BackupWarmingQueries


jpizagno:
While following the README.md section [Build Server and Client](https://github.com/Yelp/nrtsearch#build-server-and-client) the build fails with the error message below.

For more information, I tried this in a Dockerized version of nrtsearch with a base image of `FROM openjdk:14` , and also tried running on the host OS (Ubuntu 20, oracle jdk 14).  Both environment produced the same error.

To help out, I copied the full stacktrace from the gradle output into this error message below.

Does this test need to be fixed, or is this a problem with the environment?

```
~/repository/nrtsearch$ ./gradlew clean installDist test
...
> Task :test

com.yelp.nrtsearch.server.grpc.LuceneServerTest > testBackupWarmingQueries FAILED

org.junit.ComparisonFailure: expected:<...ies since uptime is [0] minutes, which is l...> but was:<...ies since uptime is [5] minutes, which is l...>
	at org.junit.Assert.assertEquals(Assert.java:117)
	at org.junit.Assert.assertEquals(Assert.java:146)
	at com.yelp.nrtsearch.server.grpc.LuceneServerTest.testBackupWarmingQueries(LuceneServerTest.java:993)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:564)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.rules.ExternalResource$1.evaluate(ExternalResource.java:54)
[...]

PR #383, opened 2021-10-16
This generalizes test LuceneServerTest.testBackupWarmingQueries()

jpizagno:
## Motivation 
It seems that when running tests locally the test LuceneServerTest.testBackupWarmingQueries() fails. This was reported in [issue-382](https://github.com/Yelp/nrtsearch/issues/382).  

## Solution
One solution would be to generalize the test. I noticed that when running the test in different environments the server uptime varied.  So this change is one way of fixing this. 

## Testing
Testing via `./gradlew clean installDist test` is green.

## Note to Reviewers
This may not be the best way of solving this problem. Is there a way of getting the gRPC server uptime?

For example, it would be better to rewrite the test to look like the example below, but I do not see a method for getting the gRPC server uptime.

```
assertEquals(
          "UNKNOWN: Unable to backup warming queries since uptime is "+replicaGrpcServer.getUptime()+" minutes, which is less than threshold 1000",
          e.getMessage());
```

### DIFF
--- a/src/test/java/com/yelp/nrtsearch/server/grpc/LuceneServerTest.java
+++ b/src/test/java/com/yelp/nrtsearch/server/grpc/LuceneServerTest.java
@@ -990,9 +990,7 @@ public class LuceneServerTest {
                   .build());
       fail("Expecting exception on the previous line");
     } catch (StatusRuntimeException e) {
-      assertEquals(
-          "UNKNOWN: Unable to backup warming queries since uptime is 0 minutes, which is less than threshold 1000",
-          e.getMessage());
+      assertEquals("UNKNOWN: Unable to backup warming queries", e.getMessage().substring(0, 41));
     }
 
     // Should fail; does not meet NumQueriesThreshold
--- a/src/test/java/com/yelp/nrtsearch/server/grpc/LuceneServerTest.java
+++ b/src/test/java/com/yelp/nrtsearch/server/grpc/LuceneServerTest.java
@@ -51,6 +51,8 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


src/test/java/com/yelp/nrtsearch/server/grpc/LuceneServerTest.java
```diff
@@ -990,9 +990,7 @@ public void testBackupWarmingQueries() throws IOException, InterruptedException
                   .build());
       fail("Expecting exception on the previous line");
     } catch (StatusRuntimeException e) {
-      assertEquals(
-          "UNKNOWN: Unable to backup warming queries since uptime is 0 minutes, which is less than threshold 1000",
-          e.getMessage());
+      assertEquals("UNKNOWN: Unable to backup warming queries", e.getMessage().substring(0, 41));
```
sarthakn7, line 993: We do want to check that the error was because of less uptime. Can we create a regex pattern which accepts the error message but any number of minutes and do a pattern match instead of exact comparison? This would be something like:
```
Pattern pattern = Pattern.compile("UNKNOWN: Unable to backup warming queries since uptime is [0-9] minutes, which is less than threshold 1000");
boolean matchFound = pattern.matcher("UNKNOWN: Unable to backup warming queries").matches();
assertTrue(matchFound);
```

jpizagno: @sarthakn7  great idea! I updated the PR to use pattern matching.  Please review again!  what are your thoughts?

sarthakn7 (approved):
Lgtm!

jpizagno:
@sarthakn7 CICD tests are green. please merge. :grinning:

PR #398, opened 2021-11-08
Add documentation todo's

frotsch:
Update TODO to include some missing doc tasks because currently it's hard to get started with nrtsearch because basic informations and settings are not documented.

### DIFF
--- a/TODO
+++ b/TODO
@@ -11,6 +11,17 @@
 - Call commit after registerFields? Else the Field states is not durable
 - Separate out committing state (for registerFields, settings, liveSettings) from being done only upon committing an index
 
+*Documentation*
+- Add docs for setting up more than one node and how to wire them together
+- Document all possible settings (valid for lucene_server_configuration.yaml or "lucene_server_default_configuration.yaml") and their respective default values 
+- Provide accessible formatted API documentation based on 
+  - https://github.com/Yelp/nrtsearch/issues/16
+  - https://github.com/Yelp/nrtsearch/pull/283
+  and strike all references to protoc-gen-doc because it's no longer used
+- Add docs about how to configure remote storage and backup/restore
+- Add docs about how to configure persistent storage
+- Add docs how nrtsearch can be monitored
+
 *Enhancements*
 - Support highlights at search time
 - Support facets at search time


TODO
```diff
@@ -11,6 +11,17 @@
...
 
+*Documentation*
+- Add docs for setting up more than one node and how to wire them together
+- Document all possible settings (valid for lucene_server_configuration.yaml or "lucene_server_default_configuration.yaml") and their respective default values 
+- Provide accessible formatted API documentation based on 
+  - https://github.com/Yelp/nrtsearch/issues/16
+  - https://github.com/Yelp/nrtsearch/pull/283
+  and strike all references to protoc-gen-doc because it's no longer used
+- Add docs about how to configure remote storage and backup/restore
+- Add docs about how to configure persistent storage
```
sarthakn7, line 22: Can you elaborate a bit about this? Nrtsearch doesn't care about storage type, so I think this would be better suited for an operator.

sarthakn7 (commented):
These are good points overall and we should definitely work on them 👍

TODO
```diff
@@ -11,6 +11,17 @@
 - Call commit after registerFields? Else the Field states is not durable
 - Separate out committing state (for registerFields, settings, liveSettings) from being done only upon committing an index
 
+*Documentation*
+- Add docs for setting up more than one node and how to wire them together
+- Document all possible settings (valid for lucene_server_configuration.yaml or "lucene_server_default_configuration.yaml") and their respective default values 
+- Provide accessible formatted API documentation based on
```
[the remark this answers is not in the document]
frotsch: There should be link from the readme to https://nrtsearch.readthedocs.io

TODO
frotsch: Where can the storage location (path) be configured
Is it ok (or not) to use NFS/NAS for persisting taxonomy or state (or do we need SSD speed for everything)
FS tuning
...

sarthakn7:
@frotsch can you actually create Github issues for these so that it is easier to track? That TODO file was created at the very beginning when @umeshdangat was working on it alone and we don't really track it anymore. A lot of the points mentioned in it are already completed. I'm going to send out a PR to delete that file soon.

PR #413, opened 2021-11-24
Add search response compression option and lz4 codec

aprudhomme:
Adds option to specify a gRPC compression codec to use for search response message. Since the only provided codec is gzip (pretty slow), an lz4 codec implementation has also been added.

Static compressor/decompressor registries have been added to the client stub builder. There are set during client channel building, and added to the main gRPC server.

The compression is set in the manner shown in the upstream example https://github.com/grpc/grpc-java/blob/master/examples/src/main/java/io/grpc/examples/experimental/CompressingHelloWorldServerPerMethod.java#L91

If an invalid compression type is specified, a warning is logged and the request continues without compression.
If the compression type is valid but not supported by the client, gRPC transparently falls back to uncompressed.


src/test/java/com/yelp/nrtsearch/server/grpc/QueryTest.java
```diff
@@ -134,6 +135,61 @@ public void testSearchV2QueryText() throws InvalidProtocolBufferException {
     LuceneServerTest.checkHits(hit);
   }
 
+  @Test
+  public void testSearchQueryResponseCompression() {
+    List<String> compressionTypes = Arrays.asList("", "identity", "gzip", "invalid");
```
sarthakn7, line 140: Should we test lz4 as well?

clientlib/src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServerStubBuilder.java
```diff
@@ -28,6 +31,16 @@
 
 /** Easy entrypoint for clients to create a Lucene Server Stub. */
 public class LuceneServerStubBuilder implements Closeable {
+  // Create registries with LZ4 compression support
+  public static final CompressorRegistry COMPRESSOR_REGISTRY =
+      CompressorRegistry.getDefaultInstance();
```
sarthakn7, line 36: Do we need to register the LZ4 codec for compressor as well?

src/test/java/com/yelp/nrtsearch/server/grpc/QueryTest.java
aprudhomme: The grpc server is build differently in testing and is only using the default registries. Maybe it is a good idea to set the custom ones there as well.

clientlib/src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServerStubBuilder.java
aprudhomme: The compressor registry is directly mutated. The lz4 codec is registered in the static {} block below.

sarthakn7 (approved):

PR #419, opened 2021-12-07
Make script cache configurable

swethakann:


src/main/java/com/yelp/nrtsearch/server/luceneserver/script/ScriptService.java
```diff
@@ -56,12 +57,12 @@ private ScriptService(LuceneServerConfiguration configuration) {
...
-    // TODO make configurable
+    ScriptCacheConfig scriptCacheConfig = configuration.getScriptCacheConfig();
     scriptCache =
         CacheBuilder.newBuilder()
-            .concurrencyLevel(4)
-            .maximumSize(1000)
-            .expireAfterAccess(1, TimeUnit.DAYS)
+            .concurrencyLevel(scriptCacheConfig.getConcurrencyLevel())
+            .maximumSize(scriptCacheConfig.getMaximumSize())
+            .expireAfterAccess(scriptCacheConfig.getExpirationTime(), TimeUnit.DAYS)
```
aprudhomme, line 65: Could we make the time unit also configurable.

src/test/java/com/yelp/nrtsearch/server/config/ScriptCacheConfigTest.java
```diff
@@ -0,0 +1,53 @@
...
+  @Test
+  public void testConfig() {
+    String configFile =
+        String.join(
+            "\n",
+            "nodeName: \"lucene_server_foo\"",
+            "ScriptCacheConfig:",
+            "  concurrencyLevel: 4",
+            "  maximumSize: 1000",
+            "  expirationTime: 1");
```
aprudhomme, line 47: You will probably want to use different values than the default here, to make sure they are really being used.

src/test/java/com/yelp/nrtsearch/server/config/ScriptCacheConfigTest.java
```diff
@@ -0,0 +1,82 @@
...
+        String.join(
+            "\n", "nodeName: \"lucene_server_foo\"", "ScriptCacheConfig:", "  timeUnit: \"ABCD\"");
+    IllegalArgumentException exception =
+        assertThrows(
+            IllegalArgumentException.class,
+            () -> {
+              ScriptCacheConfig config = getConfig(configFile);
+              assertEquals(
+                  TimeUnit.valueOf(ScriptCacheConfig.DEFAULT_TIME_UNIT), config.getTimeUnit());
+            });
```
aprudhomme, line 72: This assert would not actually be reached, right?

src/main/java/com/yelp/nrtsearch/server/config/ScriptCacheConfig.java
```diff
@@ -0,0 +1,97 @@
...
+
+  /**
+   * @param timeUnitStr
+   * @return expiration time unit from config
+   * @throws IllegalArgumentException
+   */
+  public static TimeUnit getTimeUnitFromString(String timeUnitStr) throws IllegalArgumentException {
+    if (timeUnitStr == null) {
+      throw new IllegalArgumentException("script cache expiration time unit cannot be null");
+    }
```
aprudhomme, line 58: nit: a cleaner way to do this would be using `Objects.requireNonNull()`, which will throw a `NullPointerException`. https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#requireNonNull-T-java.lang.String-
Also, since these are `RuntimeException`s, they don't need to be declared on the method signature.

aprudhomme (approved):

PR #425, opened 2022-01-04
fix an issue causing replica cannot startIndex without primary

Ziqi-Wang:
The luecene abstract replica node initiates with pull delta from the primary. We set the primaryGen to be -1 if the primary is down to allow the replica to start alone.

RP-5477

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/NRTReplicaNode.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/NRTReplicaNode.java
@@ -15,11 +15,7 @@
  */
 package com.yelp.nrtsearch.server.luceneserver;
 
-import com.yelp.nrtsearch.server.grpc.FileMetadata;
-import com.yelp.nrtsearch.server.grpc.FilesMetadata;
-import com.yelp.nrtsearch.server.grpc.GetNodesResponse;
-import com.yelp.nrtsearch.server.grpc.NodeInfo;
-import com.yelp.nrtsearch.server.grpc.ReplicationServerClient;
+import com.yelp.nrtsearch.server.grpc.*;
 import com.yelp.nrtsearch.server.monitoring.NrtMetrics;
 import com.yelp.nrtsearch.server.utils.HostPort;
 import java.io.IOException;
@@ -30,11 +26,8 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
-import org.apache.lucene.replicator.nrt.CopyJob;
+import org.apache.lucene.replicator.nrt.*;
 import org.apache.lucene.replicator.nrt.CopyState;
-import org.apache.lucene.replicator.nrt.FileMetaData;
-import org.apache.lucene.replicator.nrt.NodeCommunicationException;
-import org.apache.lucene.replicator.nrt.ReplicaNode;
 import org.apache.lucene.search.SearcherFactory;
 import org.apache.lucene.store.Directory;
 import org.slf4j.Logger;
@@ -75,7 +68,13 @@ public class NRTReplicaNode extends ReplicaNode {
     jobs.setName("R" + id + ".copyJobs");
     jobs.setDaemon(true);
     jobs.start();
-    start(primaryGen);
+    if (!this.primaryAddress.isPrimaryAvailable()) {
+      // HACK: The default replica pull sync will not check the primary's availability.
+      // We set the curPrimaryGen into -1 to skip the sync when the primary is down.
+      start(-1);
+    } else {
+      start(primaryGen);
+    }
   }
 
   @Override
--- a/src/main/java/com/yelp/nrtsearch/server/grpc/ReplicationServerClient.java
+++ b/src/main/java/com/yelp/nrtsearch/server/grpc/ReplicationServerClient.java
@@ -17,13 +17,16 @@ package com.yelp.nrtsearch.server.grpc;
 
 import com.google.protobuf.GeneratedMessageV3;
 import com.yelp.nrtsearch.server.luceneserver.SimpleCopyJob.FileChunkStreamingIterator;
+import io.grpc.ConnectivityState;
 import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
+import io.grpc.StatusRuntimeException;
 import io.grpc.stub.StreamObserver;
 import java.io.Closeable;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 import org.slf4j.Logger;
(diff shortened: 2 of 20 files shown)


src/main/java/com/yelp/nrtsearch/server/luceneserver/NRTReplicaNode.java
```diff
@@ -75,7 +69,13 @@ public NRTReplicaNode(
     jobs.setName("R" + id + ".copyJobs");
     jobs.setDaemon(true);
     jobs.start();
-    start(primaryGen);
+    if(!this.primaryAddress.isPrimaryAvailable()){
```
Ziqi-Wang, line 72: The replica is not registered yet at this point

sarthakn7:
@Ziqi-Wang can you please check the test failure?

aprudhomme:
Can we get some tests added to confirm this fixes the problem. My understanding of the issue is that it happens when trying to start a replica with a different primary generation than the restored index. I'm not sure it passing -1 works in this case https://github.com/apache/lucene/blob/7d5df2d6fee7834967884209cf1668210f25df80/lucene/replicator/src/java/org/apache/lucene/replicator/nrt/ReplicaNode.java#L194

sarthakn7:
Along with tests it would also be good to mention any manual testing that was done.

waziqi89:
> Can we get some tests added to confirm this fixes the problem. My understanding of the issue is that it happens when trying to start a replica with a different primary generation than the restored index. I'm not sure if passing -1 works in this case https://github.com/apache/lucene/blob/7d5df2d6fee7834967884209cf1668210f25df80/lucene/replicator/src/java/org/apache/lucene/replicator/nrt/ReplicaNode.java#L194
I believe are find the right place. Based on the if condition, -1 will negate the expression and stop the replica from pull. Generation comparison is to make sure to pull only if the replica version is behind the primary's ( check the assert in the next, so it's < only ).

waziqi89:
> Along with tests it would also be good to mention any manual testing that was done.
@sarthakn7 
added 2 tests. Not sure how do we fail a primary only for the manual test

waziqi89:
@sarthakn7 
I run the app with your suggestion: create a local replica, and startIndex with primary hostname "badhost". It generates a lot of warning in resolving hostname, and cannot register to primary (in keepAlive), which are as expected. The replica does not crash, and can continue register fields and serve searches.

waziqi89:
some of the logs
```[WARN ] 2022-01-07 04:07:56.847 [KeepAlive] ShardState$KeepAlive - Replica host: ‘', binary port: 8001 cannot reach primary host: badhost replication port: 8000
Jan 07, 2022 4:07:59 AM io.grpc.internal.ManagedChannelImpl$NameResolverListener handleErrorInSyncContext
WARNING: [Channel<9>: (badhost:8000)] Failed to resolve name. status=Status{code=UNAVAILABLE, description=Unable to resolve host badhost, cause=java.lang.RuntimeException: java.net.UnknownHostException: badhost: Name or service not known
        at io.grpc.internal.DnsNameResolver.resolveAddresses(DnsNameResolver.java:223)
        at io.grpc.internal.DnsNameResolver.doResolve(DnsNameResolver.java:282)
        at io.grpc.internal.DnsNameResolver$Resolve.run(DnsNameResolver.java:318)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630)
        at java.base/java.lang.Thread.run(Thread.java:832)
Caused by: java.net.UnknownHostException: badhost: Name or service not known
        at java.base/java.net.Inet6AddressImpl.lookupAllHostAddr(Native Method)
        at java.base/java.net.InetAddress$PlatformNameService.lookupAllHostAddr(InetAddress.java:932)
        at java.base/java.net.InetAddress.getAddressesFromNameService(InetAddress.java:1505)
        at java.base/java.net.InetAddress$NameServiceAddresses.get(InetAddress.java:851)
        at java.base/java.net.InetAddress.getAllByName0(InetAddress.java:1495)
        at java.base/java.net.InetAddress.getAllByName(InetAddress.java:1354)
        at java.base/java.net.InetAddress.getAllByName(InetAddress.java:1288)
        at io.grpc.internal.DnsNameResolver$JdkAddressResolver.resolveAddress(DnsNameResolver.java:631)
        at io.grpc.internal.DnsNameResolver.resolveAddresses(DnsNameResolver.java:219)
        ... 5 more
}
[WARN ] 2022-01-07 04:08:06.848 [KeepAlive] ShardState$KeepAlive - Replica host: ‘', binary port: 8001 cannot reach primary host: badhost replication port: 8000
[WARN ] 2022-01-07 04:08:16.849 [KeepAlive] ShardState$KeepAlive - Replica host: ‘', binary port: 8001 cannot reach primary host: badhost replication port: 8000
[WARN ] 2022-01-07 04:08:26.849 [KeepAlive] ShardState$KeepAlive - Replica host: ‘', binary port: 8001 cannot reach primary host: badhost replication port: 8000
[WARN ] 2022-01-07 04:08:36.850 [KeepAlive] ShardState$KeepAlive - Replica host: ‘', binary port: 8001 cannot reach primary host: badhost replication port: 8000
[INFO ] 2022-01-07 04:08:44.489 [GrpcLuceneServerExecutor-4-thread-8] RegisterFieldsHandler - REGISTER: doc_id -> com.yelp.nrtsearch.server.luceneserver.field.AtomFieldDef@72dad294
[INFO ] 2022-01-07 04:08:44.493 [GrpcLuceneServerExecutor-4-thread-8] RegisterFieldsHandler - REGISTER: vendor_name -> com.yelp.nrtsearch.server.luceneserver.field.TextFieldDef@1913bdf9
[INFO ] 2022-01-07 04:08:44.495 [GrpcLuceneServerExecutor-4-thread-8] RegisterFieldsHandler - REGISTER: license_no -> com.yelp.nrtsearch.server.luceneserver.field.IntFieldDef@77f6a70f
[INFO ] 2022-01-07 04:08:44.498 [GrpcLuceneServerExecutor-4-thread-8] LuceneServer - RegisterFieldsHandler registered fields response: "{\"license_no\":{\"name\":\"license_no\",\"type\":\"INT\",\"storeDocValues\":true,\"multiValued\":true},\"vendor_name\":{\"name\":\"vendor_name\",\"type\":\"TEXT\",\"search\":true,\"store\":true,\"tokenize\":true},\"doc_id\":{\"name\":\"doc_id\",\"storeDocValues\":true}}"
```

sarthakn7 (commented):
@Ziqi-Wang I'm not sure I understand how this works. Can you please explain why setting primaryGen to -1 means replica will not check for availability? I see `sendNewReplica()` being called either way

waziqi89:
> @Ziqi-Wang I'm not sure I understand how this works. Can you please explain why setting primaryGen to -1 means replica will not check for availability? I see `sendNewReplica()` being called either way

Agre, and sendNewReplica will be called after it. However, even though it throws the IoExeception in sendNewReplica's signature, our implementation catch all Exceptions and log them, allowing it to always succeed, even upon primary's failure.
https://github.com/Yelp/nrtsearch/blob/master/src/main/java/com/yelp/nrtsearch/server/luceneserver/NRTReplicaNode.java#L178
Anyways, sendNewReplica is not a blocker for replica to start alone.

PR #430, opened 2022-02-09
Improve global state handling, use configurable backend

aprudhomme:
Adds improvements an configurability to the handling of global cluster state.

For backwards compatibility, the `GlobalState` class is now abstract, allowing for different implementations of state mutating methods.

The config now allows for specifying a state backend type with the `stateConfig.backendType` key. There are three valid values:

### `LEGACY` (default)
Equivalent functionality to before this change. Uses `LegacyGlobalState` implementation, and writes out `indices.x` files as needed. Uses `RestoreStateHandler` to load indices names from index backup metadata.

### `LOCAL`
Uses `BackendGlobalState` implementation with `LocalStateBackend`. Uses local filesystem to store global state in a json file: `<state_dir>/global_state/state.json`.

### `REMOTE`
Uses `BackendGlobalState` implementation with `RemoteStateBackend`. Uses incremental `Archiver` to store current state in s3, with the `global_state` resource name. A local copy is maintained in the local state directory for debug purposes, but the remote copy is the source of truth. The `stateConfig.remote.readOnly` config values controls if the server instance can modify the remote state.

Global state highlights:
- State automatically loaded during startup
- State changes are durably committed to the backend before being visible locally
- State information is built into an immutable object, which is atomically replaced when changed
- Indices can now have a map of properties (`IndexInfo`) in global state, instead of just the name

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/backup/IndexArchiver.java
+++ b/src/main/java/com/yelp/nrtsearch/server/backup/IndexArchiver.java
@@ -32,7 +32,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class IndexArchiver implements Archiver {
-  private static final Logger logger = LoggerFactory.getLogger(BackupDiffManager.class);
+  private static final Logger logger = LoggerFactory.getLogger(IndexArchiver.class);
   public static final String STATE = "state";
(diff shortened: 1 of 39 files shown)


umeshdangat (approved):
looks good to me 👍🏽

PR #437, opened 2022-03-05
Add support for collector plugins

desiam:
### Context
Adding support for collector plugins
* Clients can now supply their own own `AdditionalCollectorManager` through a `Plugin`, and reference them by name
* Additional collectors now have access to the `searchContext`
* Added `Any anyResult` as a `CollectorResult` for flexibility

### Questions
I modified the interface from 
`AdditionalCollectorManager<? extends org.apache.lucene.search.Collector, ? extends CollectorResult>` to
`AdditionalCollectorManager<? extends org.apache.lucene.search.Collector, CollectorResult>`. 
* **Would there be any reason for a collector not to use `CollectorResult` directly for the final result?** 

### Verification
`./gradlew test`

`CollectorTest.java` which scores reviews by `review_count`, collects the top n reviews per business, and returns the documents as hits in the collector result.


src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
```diff
@@ -349,6 +350,7 @@ private void initExtendableComponents(
       RescorerCreator.initialize(configuration, plugins);
       ScriptService.initialize(configuration, plugins);
       SimilarityCreator.initialize(configuration, plugins);
+      CollectorCreator.initialize(configuration, plugins);
```
aprudhomme, line 353: nit: this block should be in alphabetical order

clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -585,6 +593,7 @@ message CollectorResult {
     oneof CollectorResults {
         //Result of collector that produces buckets and counts.
         BucketResult bucketResult = 1;
+        google.protobuf.Any anyResult = 2;
```
aprudhomme, line 596: nit: add doc strings for fields

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/CollectorProvider.java
```diff
@@ -0,0 +1,43 @@
...
+@FunctionalInterface
+public interface CollectorProvider<
+    T extends AdditionalCollectorManager<? extends Collector, CollectorResult>> {
+
+  /**
+   * Get an {@link AdditionalCollectorManager} implementation initialized with the given context and
+   * parameters.
+   *
+   * @param name collection name
+   * @param context search context
```
aprudhomme, line 38: nit: not the search context

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/DocCollector.java
```diff
@@ -47,6 +48,7 @@
   private boolean terminatedEarly = false;
   private SearchStatsWrapper<? extends Collector> statsWrapper = null;
   private List<CollectorStatsWrapper<?, ?>> collectorStatsWrappers = null;
+  private SearchContext searchContext;
```
aprudhomme, line 51: Is there a reason this is being kept as a member field?

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/AdditionalCollectorManager.java
```diff
@@ -28,4 +29,7 @@
 public interface AdditionalCollectorManager<C extends Collector, T> extends CollectorManager<C, T> {
   /** Get collection name as it should appear in the search response. */
   String getName();
+
+  /** Sets the search context. This must be called before collecting */
+  void setSearchContext(SearchContext searchContext);
```
aprudhomme, line 34: This could have a noop default implementation

aprudhomme (commented):
A few nits, but overall looks reasonable. It should be ok to limit the result type for the additional collector interface, I can't think of a reason we would need something else.

aprudhomme (approved):

PR #443, opened 2022-03-28
Update some config defaults

fragosoluana:
## Problem or Feature
Some default values for flags were changed due to their common use in config files. As a result, these flags can be deleted from configs to make them clear.
RP-5650

## Solution
Changed the following values in `LuceneServerConfiguration.java`
1. publishJvmMetrics: true
2. downloadAsStream: true
3. syncInitialNrtPoint: true
4. fileSendDelay: false

## Verification
- `./gradlew clean installDist test  ` successfully builds project
- Run nrtsearch locally and debug to make sure the values are being overwritten when a config file different than the default one is given

### DIFF
--- a/src/test/java/com/yelp/nrtsearch/server/grpc/ReplicationServerTest.java
+++ b/src/test/java/com/yelp/nrtsearch/server/grpc/ReplicationServerTest.java
@@ -102,7 +102,7 @@ public class ReplicationServerTest {
 
   @Test
   public void copyFiles() throws IOException, InterruptedException {
-    initNonSyncedInitialNrtPointServers();
+    initServerSyncInitialNrtPointFalse();
 
     GrpcServer.TestServer testServerPrimary =
         new GrpcServer.TestServer(luceneServerPrimary, true, Mode.PRIMARY);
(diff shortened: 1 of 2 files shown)


src/test/java/com/yelp/nrtsearch/server/grpc/ReplicationServerTest.java
```diff
@@ -372,7 +309,7 @@ public void testSyncOnIndexStart() throws IOException, InterruptedException {
                     .setTopHits(10)
                     .addAllRetrieveFields(LuceneServerTest.RETRIEVED_VALUES)
                     .build());
-    assertEquals(0, searchResponseSecondary.getHitsCount());
+    assertEquals(4, searchResponseSecondary.getHitsCount());
```
sarthakn7, line 312: I think these tests expect syncInitialNrtPoint to be false at startup. It would be good to retain the behavior of the existing tests

src/test/java/com/yelp/nrtsearch/server/grpc/ReplicationServerTest.java
```diff
@@ -511,4 +454,81 @@ public void testAddDocumentsOnReplicaFailure() throws IOException, InterruptedEx
             .getMessage()
             .contains("Adding documents to an index on a replica node is not supported"));
   }
+
+  private void initDefaultLuceneServer() throws IOException {
+    initLuceneServers("");
+  }
+
+  private void initNonSyncedInitialNrtPointServers() throws IOException {
```
sarthakn7, line 462: Nit: rename to `initServerSyncInitialNrtPointFalse`

sarthakn7 (approved):

PR #444, opened 2022-04-04
Added a custom request plugin

sarthakn7:
Plugin interface that allows implementers to receive custom requests over grpc and process and respond to them. Changes from the draft:
1. Changed `getRoutes` from `List<Routes>` to `Map<String, RequestProcessor>`. RequestProcessor is a functional interface that specifies how a request will be processed
2. With the above change it is not possible to define duplicate paths within the same plugin implementation
3. Added tests
4. Regenerated grpc-gateway files


clientlib/src/main/proto/yelp/nrtsearch/luceneserver.proto
```diff
@@ -935,4 +945,14 @@ message ForceMergeDeletesResponse {
         FORCE_MERGE_DELETES_SUBMITTED = 1;
     }
     Status status = 1;
+}
+
+message CustomRequest {
+    string id = 1; // ID defined for custom requests in a plugin
```
aprudhomme, line 951: Just a thought, do we need the `id`? It does make it harder for there to be collisions, but it requires the caller to have more knowledge about the implementation.

sarthakn7: The caller just needs to additionally know the id along with the path. While it does seem like extra information at the moment I think it is easy to have a situation where two plugins share the same path, e.g. two LTR plugins having an endpoint to add new models. Without this id the plugin implementers would have to consciously add some other identifier to the path itself to make sure that there is no collision and also so that users do not get confused.
This would be applicable only if both plugins are installed on the same cluster, which is also certainly possible if there are experiments being done only on the query side without changing the data.

aprudhomme: I guess, though the plugin creators would need to consciously select a unique id.

I was just trying to imagine what this would look like if we wanted to create a curl like utility to use this endpoint. Perhaps the id could be interpreted as a path prefix.

src/main/java/com/yelp/nrtsearch/server/plugins/CustomRequestPlugin.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Plugin interface that allows adding one or more routes and their corresponding processors which
+ * can be called via the `custom` RPC. This can be implemented by plugins to modify behavior without
+ * requiring updating config files and restarting nrtsearch.
+ */
+public interface CustomRequestPlugin {
+
+  @FunctionalInterface
+  interface RequestProcessor {
+    Map<String, String> process(Map<String, String> request);
```
aprudhomme, line 29: Should the path also be passed as a parameter? It would give the plugin more flexibility with its processor implementations.

sarthakn7: I was expecting a RequestProcessor per path, but I guess providing the path is not a huge overhead and it does provide more flexibility. I'll add it.

clientlib/src/main/proto/yelp/nrtsearch/luceneserver.proto
sarthakn7: I added both id and path as url params for the generated grpc-gateway. I was trying to specify params as the body in the request instead of * but that was throwing an error. It would work if params was struct, but the resulting json would be more complicated.

aprudhomme (approved):

PR #447, opened 2022-04-25
Add context suggest field

TheMiker:
PR to add the context suggest field to nrtsearch. This change is intended to add support for the [context suggest field](https://lucene.apache.org/core/6_1_0/suggest/index.html?org/apache/lucene/search/suggest/document/ContextSuggestField.html) in nrtsearch 

field must be stored in order to retrieve the original value, contexts, and weight data


src/main/java/com/yelp/nrtsearch/server/luceneserver/field/ContextSuggestFieldDef.java
```diff
@@ -0,0 +1,68 @@
...
+
+  @Override
+  public String getType() {
+    return "CONTEXT_SUGGEST_FIELD";
+  }
+
+  @Override
+  public void parseDocumentField(
+      Document document, List<String> fieldValues, List<List<String>> facetHierarchyPaths) {
+    if (fieldValues.size() > 1) {
```
jinxihexi0411, line 56: Is it necessary to check the `fieldValues.size() == 0`?

src/main/java/com/yelp/nrtsearch/server/luceneserver/suggest/protocol/ContextSuggestFieldData.java
```diff
@@ -0,0 +1,42 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.yelp.nrtsearch.server.luceneserver.suggest.protocol;
+
+import java.util.List;
+
```
jinxihexi0411, line 19: nit: please add a brief comment for this schema class? 
Some ideas:
- usage of this class
- meaning of each field

jinxihexi0411 (commented):
overall lg2m on the current stage! Will take another look at this after the test cases.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/ContextSuggestFieldDef.java
TheMiker: I'm not sure. I decided to follow how other fieldDef's have checked for this case https://sourcegraph.yelpcorp.com/Yelp/nrtsearch/-/blob/src/main/java/com/yelp/nrtsearch/server/luceneserver/field/VectorFieldDef.java#L87-94

src/main/java/com/yelp/nrtsearch/server/luceneserver/suggest/protocol/ContextSuggestFieldData.java
TheMiker: added!

jinxihexi0411 (approved):
lg2m!

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/ContextSuggestFieldDef.java
```diff
@@ -0,0 +1,91 @@
...
+  protected ContextSuggestFieldDef(String name, Field requestField) {
+    super(name, requestField);
+    this.indexAnalyzer = this.parseIndexAnalyzer(requestField);
+  }
+
+  @Override
+  protected void validateRequest(Field requestField) {
+    if (!requestField.getStore()) {
+      throw new IllegalArgumentException("Context Suggest fields must be stored");
+    }
```
aprudhomme, line 46: Is there a reason this is required. Stored fields take up space in the index. If it is not required to retrieve the field data directly, it might be good to be able to disable this.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/ContextSuggestFieldDef.java
```diff
@@ -0,0 +1,91 @@
...
+  public String getType() {
+    return "CONTEXT_SUGGEST_FIELD";
+  }
+
+  @Override
+  public void parseDocumentField(
+      Document document, List<String> fieldValues, List<List<String>> facetHierarchyPaths) {
+    if (fieldValues.size() > 1) {
+      throw new IllegalArgumentException(
+          "Cannot index multiple values into single value field: " + getName());
```
sarthakn7, line 67: Are multi-values not allowed at all or only when it is not defined in the schema? If they are not allowed at all you can add a check for it in `validateRequest`

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/ContextSuggestFieldDef.java
TheMiker: i've removed the requirement and now only store if the config states `store: true`

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/ContextSuggestFieldDef.java
TheMiker: i've moved this logic to validateRequest

clientlib/src/main/proto/yelp/nrtsearch/luceneserver.proto
```diff
@@ -437,6 +437,7 @@ enum FieldType {
     OBJECT = 14; // JSON
     // List of float values
     VECTOR = 15;
+    CONTEXT_SUGGEST_FIELD = 16; //Field used for contextual suggest
```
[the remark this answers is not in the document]
TheMiker: changed field name

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/ContextSuggestFieldDef.java
```diff
@@ -0,0 +1,90 @@
...
+
+  @Override
+  public String getType() {
+    return "CONTEXT_SUGGEST";
+  }
+
+  @Override
+  public void parseDocumentField(
+      Document document, List<String> fieldValues, List<List<String>> facetHierarchyPaths) {
+    if (fieldValues.size() == 1) {
```
sarthakn7, line 65: We should still throw an error if there are 0 or >1 values in the request

TheMiker: got it! i've added this error case. thanks

sarthakn7 (approved):
lgtm, would be good to have a ship from @aprudhomme as well before merging

aprudhomme (approved):

PR #448, opened 2022-04-26
Add ability to start indices from global state

aprudhomme:
Allow the started state of indices to be stored in cluster global state. Using this and config information, allow indices to be automatically started on server startup.

## Configuration
Adds new config for index start:
- `indexStartConfig.autoStart` - If indices should be automatically started on startup
- `indexStartConfig.mode` - mode to start indices in (PRIMARY, REPLICA, STANDALONE)
- `indexStartConfig.dataLocationType` - Denotes where persistent index data is present. `LOCAL` if the data is on the local disk (or a persistent volume like EBS). `REMOTE` if the data should be pulled from the latest incremental backup in s3.
- `indexStartConfig.primaryDiscovery.host/port/file` - Configuration for how a replica index should connect to the primary. Can be specified as host and port, or a discovery file. The file must contain a json list of objects containing host/port values. If the port is specified in the config, it overrides the value in the discovery file.

## Primary discovery file
The `primaryDiscoveryFile` option can be specified in a start index request, for use when the `primaryAddress` is empty. Uses a modified version of the https://github.com/Yelp/nrtsearch/blob/master/clientlib/src/main/java/com/yelp/nrtsearch/clientlib/NodeAddressFileNameResolver.java already used for load balancing connections. This resolver only allows at most one backend node to be registered at a time (as there can only be one active primary). It also allows for the port value from the discovery file to be overridden, which is useful in situations where the file is advertising the client server port instead of the replica server port.

## Index stop/start
Start and stop requests are routed through the global state, when not using legacy state. As part of processing the request, the persistent global state is updated to reflect the new index started status.

When the server starts up, it loads the global state and determines which indices need to be started. The indices are automatically started based on the server's index start config.

## Primary generation
Previously, the primary generation used by nrt processing was passed as part of the start request. This branch allows the value of `-1` to denote that it should be auto detected. For the primary, this will be the current epoch timestamp (stored on creation of global state). For the replicas, this will be the latest primary generation loaded from the restored index data.

## Ready endpoint
When no indices are listed in a request to the `ready` endpoint, typically all indices are checked to verify they are started. When using the new state management, only those marked as started in the global state are checked.

## Global state grpc message
The global state representation has been changed from a jackson annotated json class, to a protobuf message. This was done for several reasons:
- protobuf messages have built in builders/rebuilders
- protobuf messages are already immutable
- It would be good to add an endpoint to return global state, already being a message make that simple
Since the json representations are the same, this will be a compatible change


src/main/java/com/yelp/nrtsearch/server/config/IndexStartConfig.java
```diff
@@ -0,0 +1,120 @@
...
+  public static IndexStartConfig fromConfig(YamlConfigReader configReader) {
+    Objects.requireNonNull(configReader);
+    boolean autoStart = configReader.getBoolean(CONFIG_PREFIX + "autoStart", false);
+    Mode mode = Mode.valueOf(configReader.getString(CONFIG_PREFIX + "mode", "STANDALONE"));
+    String discoveryHost =
+        configReader.getString(CONFIG_PREFIX + PRIMARY_DISCOVERY_PREFIX + "host", "");
+    Integer discoveryPort =
+        configReader.getInteger(CONFIG_PREFIX + PRIMARY_DISCOVERY_PREFIX + "port", 0);
+    String discoveryFile =
+        configReader.getString(CONFIG_PREFIX + PRIMARY_DISCOVERY_PREFIX + "file", "");
```
taoyyu, line 53: what if all of host, port and file are missing? Do we have any check about this?

src/main/java/com/yelp/nrtsearch/server/config/IndexStartConfig.java
```diff
@@ -0,0 +1,120 @@
...
+  /**
+   * Create instance from provided configuration reader.
+   *
+   * @param configReader config reader
+   * @return class instance
+   */
+  public static IndexStartConfig fromConfig(YamlConfigReader configReader) {
+    Objects.requireNonNull(configReader);
+    boolean autoStart = configReader.getBoolean(CONFIG_PREFIX + "autoStart", false);
+    Mode mode = Mode.valueOf(configReader.getString(CONFIG_PREFIX + "mode", "STANDALONE"));
```
taoyyu, line 47: We also have server mode in our code for PRIMARY, REPLICA, STANDALONE. Should we unify the config?

taoyyu (approved):
lgtm. Just 2 minor questions

src/main/java/com/yelp/nrtsearch/server/config/IndexStartConfig.java
taoyyu: I mean server mode here [https://github.com/Yelp/nrtsearch/blob/master/src/main/java/com/yelp/nrtsearch/server/luceneserver/index/ImmutableIndexState.java#L394]. Just wonder if it is possible to reuse the server mode here instead of adding another srv-configs. But feel free to ignore.

src/main/java/com/yelp/nrtsearch/server/config/IndexStartConfig.java
aprudhomme: This condition is caught during the index start process https://github.com/aprudhomme/nrtsearch/blob/auto_start_indices/src/main/java/com/yelp/nrtsearch/server/luceneserver/StartIndexHandler.java#L175 I wanted to avoid strict validation here since it is still an optional feature.

PR #456, opened 2022-05-19
Add nrt_util with remote state get/put commands

aprudhomme:
Creates an `nrt_utils` command as a catch all place for utility methods. Adds commands to get and put state directly to the s3 remote backend. These are intended to mainly be used for debug/emergencies, where the servers are unable to start and use of the typical api endpoints is not possible.

`getRemoteState` - Fetch the global/index start from s3 backend. This can be printed to stdout, or optionally written to a file.

`putRemoteState` - Write the state json in a specified file into the remote s3 backend, replacing the current global/index state. Validates state data to ensure it is valid utf8 and can be built into the expected state message type. Optionally, the current state data can be written to a backup file before uploading the new version.

This branch also changes the storage format of state data to be utf8 encoded bytes. This was needed since the previous storage format prepended the content length, which would make manual editing extremely difficult. This change is not backwards compatible, but the provided utility commands should make updating old state straightforward.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/tools/nrt_utils/state/StateCommandUtils.java
+++ b/src/main/java/com/yelp/nrtsearch/tools/nrt_utils/state/StateCommandUtils.java
@@ -1,0 +1,370 @@
+/*
+ * Copyright 2022 Yelp Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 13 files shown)


src/main/java/com/yelp/nrtsearch/tools/nrt_utils/state/StateCommandUtils.java
```diff
@@ -0,0 +1,370 @@
...
+    String stateStr =
+        getStateFileContentsFromS3(
+            versionManager.getS3(),
+            versionManager.getBucketName(),
+            absoluteResourcePath,
+            stateFileName);
+    if (stateStr == null) {
+      System.out.println("No state file found in archive");
+    }
+    return stateStr;
```
taoyyu, line 144: One question. So the get command won't affect the running nrtsearch, right?

taoyyu: I mean I want to get the s3 state and replace the local one. Is that supported?

aprudhomme: These commands function out of band of the nrtsearch service. The get command will have no effect, since it is fetching the state directly from s3. The put command changes the state in the s3 backend, which will be picked up when the nrtsearch service restarts.

PR #464, opened 2022-06-02
Add cleanup command for incremental index backups

aprudhomme:
Adds a utility command to clean up unneeded index files produced by using incremental backups. Otherwise, this data accumulates indefinitely.

## Index data structure
`latest_version file` - file containing a number corresponding to the latest index version, updated on each commit
`version files` - files named with a monotonically increasing version number, contains UUID referencing an index manifest file
`manifest file` - located in main index data directory, named by UUID, contains list of all lucene index files needed for an index version
`index files` - located in the main index data directory, individual lucene segment files

## Cleanup params
`deleteAfter` - remove all versions older than this value
`minVersions` - attempt to keep this many version prior to the current version. This is mainly an attempt for safety, having some previous version to roll back to.
`gracePeriod` - do not delete index files until they are at least this interval older than the oldest retained index version. Provides some safety from timestamp skew. Also, if we start publishing merge pre-copy, this will be needed to ensure we don't end up deleting pending merge files.
`dryRun` - do not execute the deletions against s3

## Process
- Delete all `version files` that are older than `deleteAfter` and are not needed to maintain `minVersions`
- Scan through index data files
- Delete manifest files if: they are older than the min version, minus `gracePeriod`
- Delete index files if: they are older than the min version, minus `gracePeriod` AND they are not needed by union(current version index files, min retained version index files). This should protect against corrupting the retained index versions.

S3 file deletions are batched in groups of 1000 to lower total api calls.

## Tests
Some of these tests are slow, to allow for staggered timestamps of the created index files. I have created a long running test grouping so that this test class is only run when the parameter is specified.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/tools/nrt_utils/incremental/IncrementalDataCleanupCommand.java
+++ b/src/main/java/com/yelp/nrtsearch/tools/nrt_utils/incremental/IncrementalDataCleanupCommand.java
@@ -1,0 +1,436 @@
+/*
+ * Copyright 2022 Yelp Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.yelp.nrtsearch.tools.nrt_utils.incremental;
+
+import com.amazonaws.services.s3.AmazonS3;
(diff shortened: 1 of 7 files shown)


src/main/java/com/yelp/nrtsearch/tools/nrt_utils/incremental/IncrementalDataCleanupCommand.java
```diff
@@ -0,0 +1,436 @@
...
+    String trimmed = interval.trim();
+    if (trimmed.length() < 2) {
+      throw new IllegalArgumentException("Invalid time interval: " + trimmed);
+    }
+    char endChar = trimmed.charAt(trimmed.length() - 1);
+    long numberVal = Long.parseLong(trimmed.substring(0, trimmed.length() - 1));
+
+    if (numberVal < 0) {
+      throw new IllegalArgumentException("Time interval must be >= 0");
+    }
```
taoyyu, line 421: I feel it's probably safer to avoid equals to 0 here.

taoyyu (approved):
lgtm.

taoyyu (approved):

PR #465, opened 2022-06-02
add new end point to reload state for replicas

taoyyu:
As we discussed before, we want an approach to make the replicas reload the index state without restarting.
And we want to have a new endpoint to control this approach to make it easy to use in jenkins.

I refactored my branch with the latest changes and found the remote index state backend should support this easily. So not too much change here.

I decided to only support for new state nrtsearch clusters. For Primary or standalone cluster, this endpoint would be noop since update fields/settings already committed the state.

Let me know if I got something wrong with the new state backend. I'll keep adding some tests and check locally as well.


src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
```diff
@@ -1307,6 +1307,29 @@ public void stopIndex(
...
+    public void syncIndexState(
+        SyncIndexStateRequest request, StreamObserver<DummyResponse> responseObserver) {
+      if (globalState.getConfiguration().getStateConfig().useLegacyStateManagement()) {
+        responseObserver.onError(
+            Status.UNAVAILABLE.withDescription("legacy state not supported").asRuntimeException());
+        return;
+      }
+      try {
+        if (globalState.getConfiguration().getIndexStartConfig().getMode().equals(Mode.REPLICA)) {
+          globalState.getIndexStateManager(request.getIndexName()).load();
```
aprudhomme, line 1320: I think this would work for reloading index state. Do we want to support the case where a new index gets added? This would require reloading the global state, and potentially starting/stopping indices.

clientlib/src/main/proto/yelp/nrtsearch/luceneserver.proto
```diff
@@ -113,6 +113,14 @@ service LuceneServer {
       body: "*"
     };
 
+    }
+    /* Sync an index state */
+    rpc syncIndexState (SyncIndexStateRequest) returns (DummyResponse) {
```
aprudhomme, line 118: Whatever this endpoint ends up being, it would be good to have a cli command to make it easy to invoke locally.

src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
```diff
@@ -1307,6 +1307,29 @@ public void stopIndex(
...
+        }
+      } catch (Exception e) {
```
aprudhomme, line 1322: The responseObserver needs needs onNext/onCompleted called when successful.

src/main/java/com/yelp/nrtsearch/server/luceneserver/state/BackendGlobalState.java
```diff
@@ -177,6 +177,11 @@ StateBackend getStateBackend() {
     return stateBackend;
   }
 
+  @Override
+  public void reloadStateFromBackend() throws IOException {
```
aprudhomme, line 181: This method needs to be synchronized, as it modifies the global state.

src/main/java/com/yelp/nrtsearch/server/luceneserver/state/BackendGlobalState.java
```diff
@@ -177,6 +177,11 @@ StateBackend getStateBackend() {
...
+    getStateBackend().loadOrCreateGlobalState();
```
aprudhomme, line 182: This call will only retrieve the new `GlobalStateInfo` from the backend. This needs to be used to build a new `ImmutableState`, recycling the `IndexStateManager`s from the current state. It might also be good to reload the index state for the previously existing indices as well.

The `updateStartedIndices` method can be used to start indices that should be started.

Once this is done, the current immutable state can be replaced by the new one.

src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
```diff
@@ -1307,6 +1307,47 @@ public void stopIndex(
...
+      if (globalState.getConfiguration().getStateConfig().useLegacyStateManagement()) {
+        responseObserver.onError(
+            Status.UNAVAILABLE.withDescription("legacy state not supported").asRuntimeException());
+        return;
+      }
+      try {
+        if (globalState.getConfiguration().getIndexStartConfig().getMode().equals(Mode.REPLICA)) {
+          String indexName = request.getIndexName();
+          try {
+            globalState.getIndexStateManager(indexName).load();
```
aprudhomme, line 1322: The index should be detectable by checking if it exists in `globalState.getIndexNames()`.

I'm not sure what the behavior should be here.
1. Just always resync all global/index state
2. Sync only index state if it exists, otherwise sync all. Should an empty index name also trigger sync all?

taoyyu: What do you think if I remove the indexName param from the request and reload global state every time this endpoint is triggered? Do as long as this endpoint is called, reload global state, reload all index states and start all indexes if they should be started by not started yet.

src/main/java/com/yelp/nrtsearch/server/luceneserver/state/BackendGlobalState.java
```diff
@@ -177,6 +177,25 @@ StateBackend getStateBackend() {
...
 
+  @Override
+  public synchronized void reloadStateFromBackend() throws IOException {
+    GlobalStateInfo newGlobalStateInfo = getStateBackend().loadOrCreateGlobalState();
+    Map<String, IndexStateManager> newManagerMap = new HashMap<>();
+    for (Map.Entry<String, IndexGlobalState> entry :
+        newGlobalStateInfo.getIndicesMap().entrySet()) {
+      String indexName = entry.getKey();
+      IndexStateManager stateManager =
+          createIndexStateManager(indexName, entry.getValue().getId(), stateBackend);
```
aprudhomme, line 188: If the current `ImmutableState` has an `IndexStateManager` for this index and id, it should be reused instead of creating a new one.

src/main/java/com/yelp/nrtsearch/server/luceneserver/state/BackendGlobalState.java
```diff
@@ -177,6 +177,25 @@ StateBackend getStateBackend() {
...
+      stateManager.load();
+      newManagerMap.put(indexName, stateManager);
+    }
+    ImmutableState newImmutableState = new ImmutableState(newGlobalStateInfo, newManagerMap);
+    setImmutableState(newImmutableState);
```
aprudhomme, line 193: Setting the new `ImmutableState` should be done after starting any new indices. If there is a problem for some reason, the valid previous state will remain active.

src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
aprudhomme: Yeah, that sounds good.

clientlib/src/main/proto/yelp/nrtsearch/luceneserver.proto
```diff
@@ -113,6 +113,14 @@ service LuceneServer {
       body: "*"
     };
 
+    }
+    /* Sync an index state */
+    rpc reloadIndexState (ReloadIndexStateRequest) returns (DummyResponse) {
```
aprudhomme, line 118: Since this is no longer reloading just the index state, perhase the endpoint name could be changed. This would also be true for some of the other methods/resources.

Also, it might be good to create a placeholder response message type instead of using `DummyResponse`. I could imagine wanting to return something like the current state generations in the future.

src/main/java/com/yelp/nrtsearch/server/luceneserver/state/BackendGlobalState.java
```diff
@@ -177,6 +177,29 @@ StateBackend getStateBackend() {
...
 
+  @Override
+  public synchronized void reloadStateFromBackend() throws IOException {
+    GlobalStateInfo newGlobalStateInfo = getStateBackend().loadOrCreateGlobalState();
+    Map<String, IndexStateManager> newManagerMap = new HashMap<>();
+    for (Map.Entry<String, IndexGlobalState> entry :
+        newGlobalStateInfo.getIndicesMap().entrySet()) {
+      String indexName = entry.getKey();
+      IndexStateManager stateManager;
+      if (immutableState.indexStateManagerMap.containsKey(indexName)) {
```
aprudhomme, line 188: nit: I would probably just get the value and do a null check.

src/main/java/com/yelp/nrtsearch/server/luceneserver/state/BackendGlobalState.java
```diff
@@ -177,6 +177,29 @@ StateBackend getStateBackend() {
...
+        stateManager = immutableState.indexStateManagerMap.get(indexName);
```
aprudhomme, line 189: It might be good to check if the previous state manager also has the same id as the index entry, otherwise create a new one. You might need to add a method to expose this.

src/main/java/com/yelp/nrtsearch/server/luceneserver/state/BackendGlobalState.java
```diff
@@ -202,6 +225,10 @@ public void setStateDir(Path source) throws IOException {
     throw new UnsupportedOperationException();
   }
 
+  private void setImmutableState(ImmutableState immutableState) {
+    this.immutableState = immutableState;
+  }
```
aprudhomme, line 230: nit: as this is a separate method, it might be good to verify `Thread.holdsLock(this)` before updating the active state.

src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
```diff
@@ -1307,6 +1307,31 @@ public void stopIndex(
...
+        ReloadIndexStateRequest request, StreamObserver<DummyResponse> responseObserver) {
+      if (globalState.getConfiguration().getStateConfig().useLegacyStateManagement()) {
+        responseObserver.onError(
+            Status.UNAVAILABLE.withDescription("legacy state not supported").asRuntimeException());
+        return;
+      }
+      try {
+        if (globalState.getConfiguration().getIndexStartConfig().getMode().equals(Mode.REPLICA)) {
+          globalState.reloadStateFromBackend();
+        }
```
aprudhomme, line 1321: nit: maybe add a log line if this is called by a non replica.

src/main/java/com/yelp/nrtsearch/server/luceneserver/state/BackendGlobalState.java
taoyyu: Should I make this method synchronized or just remove this method but just add this line in the synchronized reloadStateFromBackend method directly?

aprudhomme: I would probably either remove it, or throw an exception if the thread does not hold the object monitor.

clientlib/src/main/proto/yelp/nrtsearch/luceneserver.proto
```diff
@@ -113,6 +113,14 @@ service LuceneServer {
       body: "*"
     };
 
+    }
+    /* Sync an index state */
+    rpc reloadState (ReloadStateRequest) returns (ReloadStateResponse) {
```
aprudhomme, line 118: nit: comment need to be updated

src/main/java/com/yelp/nrtsearch/server/cli/ReloadStateCommand.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.yelp.nrtsearch.server.grpc.LuceneServerClient;
+import java.util.concurrent.Callable;
+import picocli.CommandLine;
+
+@CommandLine.Command(
+    name = ReloadStateCommand.RELOAD_INDEX_STATE,
+    description = "Reloads the replica's index state from backend")
+public class ReloadStateCommand implements Callable<Integer> {
+  public static final String RELOAD_INDEX_STATE = "reloadIndexState";
```
aprudhomme, line 26: There are still references to index state here

src/main/java/com/yelp/nrtsearch/server/cli/ReloadStateCommand.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.yelp.nrtsearch.server.grpc.LuceneServerClient;
+import java.util.concurrent.Callable;
+import picocli.CommandLine;
+
+@CommandLine.Command(
+    name = ReloadStateCommand.RELOAD_INDEX_STATE,
+    description = "Reloads the replica's state from backend")
+public class ReloadStateCommand implements Callable<Integer> {
+  public static final String RELOAD_INDEX_STATE = "reloadState";
```
aprudhomme, line 26: nit: variable name contains `INDEX`

aprudhomme (commented):
Looks good to me, just need to rebuild the grpc gateway to re generate the api files.

aprudhomme (approved):

PR #468, opened 2022-06-15
Update commons-compress to 1.21

aprudhomme:
Update the commons-compress library to 1.21 for security fixes.

I also needed to enable support for big numbers, since this new version uses the user/group ids. These ids can potentially be large (such as for osx). Upstream ticket https://issues.apache.org/jira/browse/COMPRESS-587

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,7 +22,7 @@ sourceCompatibility = 1.14
 targetCompatibility = 1.14
 
 allprojects {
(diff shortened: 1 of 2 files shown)


taoyyu (approved):

waziqi89:
Do we consider merge this PR? I met the BigNumber issue recently on master

PR #469, opened 2022-06-22
change the logging level of failed request

waziqi89:
To reduce the PIIs exposure risks on failed request (especially for lat/long info), reduce the logging level into debug.


sarthakn7:
How will we get the failed search request in the future if this is merged?

PR #480, opened 2022-08-09
Fix issue when reducing nested collectors with different value sets

aprudhomme:
Fixes case during nested collector reduction when the nested value does not exist in all parallel collection slices.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/additional/NestedCollectorManagers.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/additional/NestedCollectorManagers.java
@@ -110,7 +110,7 @@ public class NestedCollectorManagers {
       Collection<Collector> collectors =
           nestedCollectors.stream()
               .map(m -> m.nestedCollectorsByValue.get(value))
-              .filter(c -> c.containsKey(entry.getKey()))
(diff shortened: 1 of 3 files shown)


src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/additional/NestedCollectorManagers.java
```diff
@@ -110,7 +110,7 @@ public Map<String, CollectorResult> reduce(
       Collection<Collector> collectors =
           nestedCollectors.stream()
               .map(m -> m.nestedCollectorsByValue.get(value))
-              .filter(c -> c.containsKey(entry.getKey()))
+              .filter(c -> c != null && c.containsKey(entry.getKey()))
```
swekannan, line 113: Will the python None value from the py grpc client translate to null for checking in this case?

swekannan (approved):

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/additional/NestedCollectorManagers.java
aprudhomme: This value is not user provided, so it shouldn't be an issue.

PR #493, opened 2022-08-30
Dirchev add support for completion queries

dirchev:
# Add Support for `CompletionQuery` in Search

## Overview

This PR adds the capability to execute `CompletionQuery` against an index using a field of type `CONTEXT_SUGGEST`. The search is done using `SuggestIndexSearcher`.

### New `CompletionQuery`

A new type of `query` was added to the `SearchRequest` schema called `CompletionQuery`. It allows client to specify the type of `CompletionQuery` that has to be created, along with `text`, `contexts` and `field`. Based on the `CompletionQuery` definition, an actual Lucene query is constructed in the `QueryMapper`. 

Example search request using prefix completion query definition:

```json
{
    "indexName": "example",
    "startHit": 0,
    "topHits": 5,
    "retrieveFields": ["business_id", "payload"],
    "query": {
        "completionQuery": {
            "field": "name_suggest",
            "queryType": "PREFIX_QUERY",
            "text": "burger",
            "contexts": ["US"]
        }
    }
}
```

### New `MyTopSuggestDocsCollector`

New class `MyTopSuggestDocsCollector` was created to provide a collector manager, which under the hood uses `TopSuggestDocsCollector` to collect search results. Docs collected from all leaves are merged together using `TopSuggestDocs.merge`. 

### Integration within `search`

The rest of the logic, such as getting the `IndexReader` and processing of `Hits` is the same as any other `SearchRequest`. This means that `CompletionQuery` can be executed on the latest index state, or a snapshot; and `retrieveFields` are retrieved the same way on the resulting `hits`.

## Testing

Added tests to `ContextSuggestFieldDefTest` to execute a number of searches testing the `ContextSuggestFieldDef` field and `CompletionQuery` in `SearchRequest`


src/main/java/com/yelp/nrtsearch/server/luceneserver/SearchHandler.java
```diff
@@ -177,9 +184,11 @@ public SearchResponse handle(IndexState indexState, SearchRequest searchRequest)
       }
 
       // add results from any extra collectors
-      searchContext
-          .getResponseBuilder()
-          .putAllCollectorResults(searcherResult.getCollectorResults());
+      if (searcherResult != null) {
```
dirchev, line 187: This is the only difference with the typical search functionality - we do not have a `searcherResult` object

src/main/java/com/yelp/nrtsearch/server/luceneserver/SearchHandler.java
```diff
@@ -169,6 +171,11 @@ public SearchResponse handle(IndexState indexState, SearchRequest searchRequest)
             .addAllFacetResult(
                 FacetTopDocs.facetTopDocsSample(
                     hits, searchRequest.getFacetsList(), indexState, s.searcher, diagnostics));
+      } else if (isCompletionQuery(searchContext)) {
+        SuggestIndexSearcher suggestIndexSearcher =
+            new SuggestIndexSearcher(s.searcher.getIndexReader());
+        CompletionQuery completionQuery = (CompletionQuery) searchContext.getQuery();
+        hits = suggestIndexSearcher.suggest(completionQuery, searchContext.getTopHits(), true);
```
aprudhomme, line 178: This will first order work, but has some down sides. The search will be done single threaded, and you won't have access to the `key` or `context` fields in the `SuggestScoreDoc`.

A cleaner way would be to check the query type here https://github.com/Yelp/nrtsearch/blob/master/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/SearchRequestProcessor.java#L330 and provide a new `DocCollector` implementation for suggest.

Here is a simple example https://github.com/Yelp/nrtsearch/blob/master/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/HitCountCollector.java For your case, `newCollector()` will provide a new `TopSuggestDocsCollector` (See `SuggestIndexSearcher` implementation), and `reduce()` will get the `TopSuggestDocs` from each Collector and use `TopSuggestDocs.merge()` to combine them.

The `ScoreDoc` passed to `fillHitRanking()` will be of type `SuggestScoreDoc`, and can be used to put the key/context fields into the response along with the score if desired. https://github.com/Yelp/nrtsearch/blob/master/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/RelevanceCollector.java#L64

dirchev: Thanks for the guidance!

It turns out there is no need to use `SuggestIndexSearcher` after all using the approach you have suggested here. I have updated the code with your suggestion, using a custom collector (with collector manager) for completion queries.

src/main/java/com/yelp/nrtsearch/server/luceneserver/SearchHandler.java
```diff
@@ -277,6 +280,10 @@ public SearchResponse handle(IndexState indexState, SearchRequest searchRequest)
     return searchContext.getResponseBuilder().build();
   }
 
+  private boolean isCompletionQuery(SearchContext searchContext) {
```
aprudhomme, line 283: I don't think this is used anymore

src/main/java/com/yelp/nrtsearch/server/luceneserver/SearchHandler.java
```diff
@@ -177,9 +178,11 @@ public SearchResponse handle(IndexState indexState, SearchRequest searchRequest)
       }
 
       // add results from any extra collectors
-      searchContext
-          .getResponseBuilder()
-          .putAllCollectorResults(searcherResult.getCollectorResults());
+      if (searcherResult != null) {
```
aprudhomme, line 181: I don't think this can be null anymore.

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/MyTopSuggestDocsCollector.java
```diff
@@ -0,0 +1,82 @@
...
+public class MyTopSuggestDocsCollector extends DocCollector {
+  private final MyTopSuggestDocsCollectorManager manager;
+
+  public MyTopSuggestDocsCollector(
+      CollectorCreatorContext context,
+      List<AdditionalCollectorManager<? extends Collector, ? extends CollectorResult>>
+          additionalCollectors) {
+    super(context, additionalCollectors);
+    manager = new MyTopSuggestDocsCollectorManager(getNumHitsToCollect());
+  }
```
aprudhomme, line 39: It might be good to throw an exception here if `additionalCollectors` is not empty, since they won't have a meaningful result with the suggest search flow.

clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -210,6 +210,19 @@ message ExistsQuery {
...
+    FUZZY_QUERY = 1; // FuzzyCompletionQuery within ContextCompletionQuery
+}
+
+// A query used for suggest completion
+message CompletionQuery {
+    string field = 1; // Field in the document to query, must be a SuggestField
+    CompletionQueryType queryType = 2;
+    string text = 3;
+    repeated string contexts = 4;
+}
```
aprudhomme, line 224: nit: add remain field documentation, also it needs to be on the line above the field to generation swagger docs correctly (yes, it is wrong in a lot of other places)

aprudhomme (commented):
Looks reasonable overall. You will also need to do a `./gradlew buildGrpcGateway` to regenerate some files from the proto changes.

src/main/java/com/yelp/nrtsearch/server/luceneserver/SearchHandler.java
dirchev: resolved

aprudhomme (approved):
Looks good to me.

PR #499, opened 2022-09-08
Add support for multiValued CONTEXT_SUGGEST field

dirchev:
# Add support for multiValued ContextSuggest field

## Overview

This PR adds support for `multiValued = true` option for `CONTEXT_SUGGEST` field and the appropriate parsing of documents containing multiple values for `CONTEXT_SUGGEST` field.

The PR also updates how `MyTopSuggestDocsCollector` collects results and its final `reduce` logic.

### `multiValued = true` for `CONTEXT_SUGGEST` fields

The `ContextSuggestFieldDef` class was updated not to throw and error when `multiValued = true` is used when registering fields. Additionally, the `parseDocumentField` was updated to go through the list of values and process each field individually.

The change allows `CONTEXT_SUGGEST` field to be registered with `multiValued = true`:

```json5
[
  // full field def omitted for clarity
  {
    "name": "my_context_suggest_field",
    "type": "CONTEXT_SUGGEST",
    "multiValued": true // allowed
   }
]
```

Therefore, when adding documents, we can set the value of the field using an array:

```json5
{
  "my_context_suggest_field": [
    "{\"value\":\"Magical Pizza Place\",\"contexts\":[\"a\"],\"weight\":6}",
    "{\"value\":\"Pizza Place\",\"contexts\":[\"b\"],\"weight\":6}",
    "{\"value\":\"Place\",\"contexts\":[\"c\"],\"weight\":5}"
  ]
}
```

Which results in multiple context suggest fields being added to the document while parsing:

```java
document.add(new ContextSuggestField("my_context_suggest_field", "Magical Pizza Place", 7, Set.of("a"));
document.add(new ContextSuggestField("my_context_suggest_field", "Pizza Place", 6, Set.of("b"));
document.add(new ContextSuggestField("my_context_suggest_field", "Place", 5, Set.of("c"));
```

This way the document will be retrieved using a `CompletionQuery` for texts like `Magical`, but also `Pizza` and `Place`.

### Updated `reduce` logic in `MyTopSuggestDocsCollector`

Currently `MyTopSuggestDocsCollector` would create collectors `skipDuplicates = true` resulting in `SuggestScoreDocs` with the same `key` being deduplicated, even if they have a different `docId`. Additionally, the `TopSuggestDocs.merge` would not deduplicate, meaning that `SuggestScoreDocs` for the same document would be persisted.

The logic in `MyTopSuggestDocsCollector` was updated to achieve the following:

1. When collecting documents, each collector will retrieve all `SuggestScoreDocs` that match the completion text (`key`).
2. During `reduce`, when results from each collector are consolidated, we will return the top scoring `SuggestScoreDocs` sorted by highest score first, unique by `docId`, and limited by `numHitsToCollect`.

The test added demonstrates the new behaviour:

https://github.com/Yelp/nrtsearch/blob/367324aacc18d8f54492a5b0ca7534d3c3e441fa/src/test/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/MyTopSuggestDocsCollectorTest.java#L41-L98

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/field/ContextSuggestFieldDef.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/field/ContextSuggestFieldDef.java
@@ -60,6 +60,10 @@ public class ContextSuggestFieldDef extends IndexableFieldDef {
   @Override
   public void parseDocumentField(
       Document document, List<String> fieldValues, List<List<String>> facetHierarchyPaths) {
+    if (!isMultiValue() && fieldValues.size() > 1) {
+      throw new IllegalArgumentException(
+          "Cannot index multiple values into single value field: " + getName());
+    }
     for (String fieldValue : fieldValues) {
       parseFieldValueToDocumentField(document, fieldValue);
     }
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/field/ContextSuggestFieldDef.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/field/ContextSuggestFieldDef.java
@@ -50,10 +50,6 @@ public class ContextSuggestFieldDef extends IndexableFieldDef {
     if (requestField.getSearch()) {
       throw new IllegalArgumentException("Context Suggest fields cannot be searched");
     }
-
-    if (requestField.getMultiValued()) {
-      throw new IllegalArgumentException("Cannot index multiple values into context suggest field");
-    }
   }
 
   @Override
@@ -64,20 +60,24 @@ public class ContextSuggestFieldDef extends IndexableFieldDef {
   @Override
   public void parseDocumentField(
       Document document, List<String> fieldValues, List<List<String>> facetHierarchyPaths) {
-    if (fieldValues.size() == 1) {
-      ContextSuggestFieldData csfData =
-          GSON.fromJson(fieldValues.get(0), ContextSuggestFieldData.class);
-      CharSequence[] contexts =
-          csfData.getContexts().toArray(new CharSequence[csfData.getContexts().size()]);
-      ContextSuggestField csf =
-          new ContextSuggestField(getName(), csfData.getValue(), csfData.getWeight(), contexts);
-      document.add(csf);
-      if (isStored()) {
-        document.add(new FieldWithData(getName(), fieldType, fieldValues.get(0)));
-      }
-    } else {
-      throw new IllegalArgumentException("Context Suggest Field can only index exactly one value");
+    for (String fieldValue : fieldValues) {
+      parseFieldValueToDocumentField(document, fieldValue);
     }
+  }
+
+  /**
+   * Processes a single fieldValue and adds appropriate fields to the document.
+   *
+   * @param document document to add parsed values to
+   * @param fieldValue string representation of the field value
+   */
+  private void parseFieldValueToDocumentField(Document document, String fieldValue) {
+    ContextSuggestFieldData csfData = GSON.fromJson(fieldValue, ContextSuggestFieldData.class);
+    CharSequence[] contexts =
+        csfData.getContexts().toArray(new CharSequence[csfData.getContexts().size()]);
+    ContextSuggestField csf =
+        new ContextSuggestField(getName(), csfData.getValue(), csfData.getWeight(), contexts);
+    document.add(csf);
   }
 
   protected Analyzer parseIndexAnalyzer(Field requestField) {
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/MyTopSuggestDocsCollector.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/MyTopSuggestDocsCollector.java
@@ -19,11 +19,14 @@ import com.yelp.nrtsearch.server.grpc.CollectorResult;
 import com.yelp.nrtsearch.server.grpc.SearchResponse;
 import java.io.IOException;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import org.apache.lucene.search.Collector;
 import org.apache.lucene.search.CollectorManager;
 import org.apache.lucene.search.ScoreDoc;
 import org.apache.lucene.search.TopDocs;
+import org.apache.lucene.search.TotalHits;
 import org.apache.lucene.search.suggest.document.TopSuggestDocs;
 import org.apache.lucene.search.suggest.document.TopSuggestDocsCollector;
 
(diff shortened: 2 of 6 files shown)


aprudhomme:
I'm looking at the `TopSuggestDocsCollector` and it seems that the `collect()` method assumes that the index is sorted by the weight score. This will likely cause issues when using different scores within the same document, as collection could terminate early. There are a couple ways this could be dealt with:

- Override the `collect` method to not do early termination (more docs scored)
- Enforce that all scores in a document are equal, and sort by that (less flexible)

The index will need to provide the sort setting in the IndexSettings for the second approach. The suggest field is not sortable. Either a separate field could be added to sort by, or perhaps the suggest field could be made sortable. This would probably require something like adding a numeric field to the document, implementing the https://github.com/Yelp/nrtsearch/blob/master/src/main/java/com/yelp/nrtsearch/server/luceneserver/field/properties/Sortable.java interface, and creating the needed `SortField` https://github.com/Yelp/nrtsearch/blob/master/src/main/java/com/yelp/nrtsearch/server/luceneserver/field/NumberFieldDef.java#L268

dirchev:
@aprudhomme 

Good catch!

I will go with the first option and override the `collect` method. We have a use-case in the future where the same document may contain different `SuggestField`s with different scores, so I do not want to block that.

Will start with a quick and probably not the most performant implementation, and we can think of ways of how to improve it in the future.

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/MyTopSuggestDocsCollector.java
```diff
@@ -79,7 +93,119 @@ public TopSuggestDocs reduce(Collection<TopSuggestDocsCollector> collectors)
...
+    public ExhaustiveTopSuggestDocsCollector(int num) {
+      super(num, false);
+      priorityQueue = new SuggestScoreDocPriorityQueue(num);
+    }
+
+    @Override
+    protected void doSetNextReader(LeafReaderContext context) {
+      docBase = context.docBase;
+      // clears priority queue
+      priorityQueue.getResults();
```
aprudhomme, line 140: I don't think you want to clear the queue here. This collector will process multiple leaves (segments) and you will want the queue to rank all the results together.

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/MyTopSuggestDocsCollector.java
```diff
@@ -79,7 +93,119 @@ public TopSuggestDocs reduce(Collection<TopSuggestDocsCollector> collectors)
...
+          }
+        }
+      }
+
+      // retrieve the top this.numHitsToCollect suggestScoreDocs, sorting them by score
+      TopSuggestDocs.SuggestScoreDoc[] suggestScoreDocs =
+          topSuggestScoreDocsUniqueByDoc.values().stream()
+              .sorted((o1, o2) -> -1 * Float.compare(o1.score, o2.score)) // highest score first
+              .limit(this.numHitsToCollect)
+              .toArray(TopSuggestDocs.SuggestScoreDoc[]::new);
```
aprudhomme, line 121: I don't think all of this is needed if deduplication is done by id elsewhere. A given doc id will only exist in a single segment, so there will be no duplication between collectors.

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/MyTopSuggestDocsCollector.java
```diff
@@ -79,7 +93,119 @@ public TopSuggestDocs reduce(Collection<TopSuggestDocsCollector> collectors)
...
+    }
+  }
+
+  public static class SuggestScoreDocPriorityQueue {
+    private final Map<Integer, TopSuggestDocs.SuggestScoreDoc> docIdsMap;
+    private final PriorityQueue<TopSuggestDocs.SuggestScoreDoc> priorityQueue;
+
+    public SuggestScoreDocPriorityQueue(int num) {
+      docIdsMap = new HashMap<>(num);
+      priorityQueue = new PriorityQueue<>(num, (o1, o2) -> -1 * Float.compare(o1.score, o2.score));
```
aprudhomme, line 169: It would be good to use the same comparator
```
          (a, b) -> {
            // sort by higher score
            int cmp = Float.compare(b.score, a.score);
            if (cmp == 0) {
              // tie break by completion key
              cmp = Lookup.CHARSEQUENCE_COMPARATOR.compare(a.key, b.key);
              if (cmp == 0) {
                // prefer smaller doc id, in case of a tie
                cmp = Integer.compare(a.doc, b.doc);
              }
            }
            return cmp;
          }
```

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/MyTopSuggestDocsCollector.java
```diff
@@ -79,7 +93,119 @@ public TopSuggestDocs reduce(Collection<TopSuggestDocsCollector> collectors)
...
+          || docIdsMap.get(suggestScoreDoc.doc).score < suggestScoreDoc.score) {
+        TopSuggestDocs.SuggestScoreDoc previousValue =
+            docIdsMap.put(suggestScoreDoc.doc, suggestScoreDoc);
+        if (previousValue != null) {
+          // remove old value from priority queue
+          priorityQueue.remove(previousValue);
+        }
+        // add new value to priority queue
+        priorityQueue.add(suggestScoreDoc);
+      }
```
aprudhomme, line 191: You will need see how performant this is. Some of the operations are O(n) with queue size.

I suspect that ranking will process the hits sequentially by doc id. If that is confirmed, you only need to keep track of the last doc id and score, and defer adding to the queue until you see a new id. This may also be possible to do at the collector level, so you can use the standard `SuggestScoreDocPriorityQueue`.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/ContextSuggestFieldDef.java
```diff
@@ -64,22 +60,26 @@ public String getType() {
...
-          new ContextSuggestField(getName(), csfData.getValue(), csfData.getWeight(), contexts);
-      document.add(csf);
-      if (isStored()) {
-        document.add(new FieldWithData(getName(), fieldType, fieldValues.get(0)));
-      }
-    } else {
-      throw new IllegalArgumentException("Context Suggest Field can only index exactly one value");
+    for (String fieldValue : fieldValues) {
+      parseFieldValueToDocumentField(document, fieldValue);
     }
```
aprudhomme, line 65: Probably not a big deal right now, but it would be good to fail if multiple values are provided when the field does not have multi value set true.

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/MyTopSuggestDocsCollector.java
dirchev: > I suspect that ranking will process the hits sequentially by doc id. 

I tested this and it is not sequential by doc id.

---

In terms of performance - the slowest operation here `O(n)` is on `priorityQueue.remove`. I have not given it much thought on performance as I said [here](https://github.com/Yelp/nrtsearch/pull/499#issuecomment-1241737156). We can test this once we get more data and see how it performs.

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/MyTopSuggestDocsCollector.java
dirchev: resolved

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/MyTopSuggestDocsCollector.java
dirchev: resolved

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/ContextSuggestFieldDef.java
dirchev: resolved, added exception similar to other field def classes

aprudhomme (approved):

PR #501, opened 2022-09-12
change the max message size in grpc gateway to 100M

waziqi89:
Increase the maxInboundMessageSize in grpc-gateway to 100MB.

Set up local manual test with the same request before and after the change, as work as expected.

### DIFF
--- a/http_wrapper.go
+++ b/http_wrapper.go
@@ -32,7 +32,7 @@ func run() error {
   mux := runtime.NewServeMux()
   runtime.SetHTTPBodyMarshaler(mux)
 
-  opts := []grpc.DialOption{grpc.WithInsecure()}
+  opts := []grpc.DialOption{grpc.WithInsecure(), grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(104_857_600))}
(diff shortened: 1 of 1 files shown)


http_wrapper.go
```diff
@@ -32,7 +32,7 @@ func run() error {
   mux := runtime.NewServeMux()
   runtime.SetHTTPBodyMarshaler(mux)
 
-  opts := []grpc.DialOption{grpc.WithInsecure()}
+  opts := []grpc.DialOption{grpc.WithInsecure(), grpc.WithMaxMsgSize(10,485,760)}
```
vim345, line 35: It looks like [this](https://pkg.go.dev/google.golang.org/grpc#WithMaxMsgSize) function has been deprecated.

```
Deprecated: use WithDefaultCallOptions(MaxCallRecvMsgSize(s)) instead. Will be supported throughout 1.x.
```

Also, do we need the commas for the int?

sarthakn7:
Let's increase it to 100 MB so that this keeps working for larger messages.

waziqi89:
> Let's increase it to 100 MB so that this keeps working for larger messages.

done

sarthakn7 (approved):

PR #503, opened 2022-09-28
Increased max retry police to s3

fragosoluana:
Trying to increase the max retry to make s3 requests more stable.

Related PR: https://github.com/Yelp/nrtsearch/pull/502

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/tools/nrt_utils/incremental/DeleteIncrementalSnapshotsCommand.java
+++ b/src/main/java/com/yelp/nrtsearch/tools/nrt_utils/incremental/DeleteIncrementalSnapshotsCommand.java
@@ -120,8 +120,7 @@ public class DeleteIncrementalSnapshotsCommand implements Callable<Integer> {
   public Integer call() throws Exception {
     if (s3Client == null) {
       s3Client =
-          StateCommandUtils.createS3Client(
-              bucketName, region, credsFile, credsProfile, maxRetry);
+          StateCommandUtils.createS3Client(bucketName, region, credsFile, credsProfile, maxRetry);
     }
     VersionManager versionManager = new VersionManager(s3Client, bucketName);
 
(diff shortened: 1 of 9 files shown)


src/main/java/com/yelp/nrtsearch/tools/nrt_utils/state/StateCommandUtils.java
```diff
@@ -95,7 +95,7 @@ public static AmazonS3 createS3Client(
         new RetryPolicy(
             PredefinedRetryPolicies.DEFAULT_RETRY_CONDITION,
             PredefinedRetryPolicies.DEFAULT_BACKOFF_STRATEGY,
-            5,
```
sarthakn7, line 98: Can we get this parameter from a config?

fragosoluana: Probably. Should we test if this change will actually fix the tron jobs first? If yes, we can create another PR to get this param from a config file.

sarthakn7: Why not just have one PR which gets the parameter value externally? It's never a good practice to hard-code these values in code.

src/main/java/com/yelp/nrtsearch/tools/nrt_utils/incremental/DeleteIncrementalSnapshotsCommand.java
```diff
@@ -103,6 +103,12 @@ public class DeleteIncrementalSnapshotsCommand implements Callable<Integer> {
       description = "Print file deletions, instead of applying to S3")
   private boolean dryRun;
 
+  @CommandLine.Option(
+      names = {"--maxErrorRetry"},
```
sarthakn7, line 107: Nit: just call these parameters `maxRetry`

sarthakn7 (approved):

sarthakn7 (changes requested):
Can you please run `./gradlew clean spotlessApply`?

sarthakn7 (approved):

PR #507, opened 2022-10-21
Merge precopy improvements

sarthakn7:
1. Fix possible race condition where a new replica might not be added to merge pre-copy
2. Avoid iterating over all clients when adding a new replica to merge precopy
3. Remove unused synchronization on preCopy.connections
4. Track transferStatus of merge precopy for new replicas. This will avoid errors like
```
top: ignore exception trying to read byte during warm for segment=<segment details> to replica=ReplicationServerClient(host=..., port=..., discoveryFile=): java.lang.NullPointerException: Cannot invoke "java.util.Iterator.hasNext()" because "transferStatusIterator" is null files=[_ihsw2.cfs, _ihsw2.cfe, _ihsw2.si] java.lang.NullPointerException: Cannot invoke "java.util.Iterator.hasNext()" because "transferStatusIterator" is null
	at com.yelp.nrtsearch.server.luceneserver.NRTPrimaryNode.preCopyMergedSegmentFiles(NRTPrimaryNode.java:368)
	at org.apache.lucene.replicator.nrt.PreCopyMergedSegmentWarmer.warm(PreCopyMergedSegmentWarmer.java:58)
	at org.apache.lucene.index.IndexWriter.mergeMiddle(IndexWriter.java:4610)
	at org.apache.lucene.index.IndexWriter.merge(IndexWriter.java:4057)
	at org.apache.lucene.index.ConcurrentMergeScheduler.doMerge(ConcurrentMergeScheduler.java:625)
	at org.apache.lucene.index.ConcurrentMergeScheduler$MergeThread.run(ConcurrentMergeScheduler.java:662)}
```
5. Make logs more understandable
6. Add a deadline to copyFiles. This is a temporary solution, going to follow up with a configurable deadline

I wasn't sure of a good way to write tests for these changes, but I confirmed that setting the deadline does cancel the transferStatusIterator.hasNext() call if deadline is exceeded.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/NRTPrimaryNode.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/NRTPrimaryNode.java
@@ -27,12 +27,14 @@ import java.io.PrintStream;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Queue;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import org.apache.lucene.index.IndexWriter;
@@ -117,8 +119,7 @@ public class NRTPrimaryNode extends PrimaryNode {
 
   /** Holds all replicas currently warming (pre-copying the new files) a single merged segment */
   static class MergePreCopy {
-    final List<ReplicationServerClient> connections =
-        Collections.synchronizedList(new ArrayList<>());
+    final Set<ReplicationServerClient> connections = Collections.synchronizedSet(new HashSet<>());
     final Map<String, FileMetaData> files;
     private boolean finished;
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/yelp/nrtsearch/server/luceneserver/NRTPrimaryNode.java
```diff
@@ -372,37 +389,39 @@ public void addReplica(int replicaID, ReplicationServerClient replicationServerC
...
+        Iterator<TransferStatus> transferStatusIterator =
+            replicationServerClient.copyFiles(indexName, primaryGen, filesMetadata);
+        logMessage(
+            String.format(
+                "Start precopying merged segments for new replica %s:%d",
+                replicaDetails.replicationServerClient.getHost(),
+                replicaDetails.replicationServerClient.getPort()));
+
+        // If the preCopy is still in progress add this transfer to it as well
+        if (!preCopy.tryAddConnection(replicationServerClient, transferStatusIterator)) {
```
aprudhomme, line 415: A small issue with this is that if the preCopy is already finished, the iterator never gets consumed. I think it might be better to do something like
```
synchronized(preCopy) {
  if (!preCopy.finished()) {
    // start copyFiles request
    // add request iterator to preCopy
  }
}
```

aprudhomme (approved):

PR #510, opened 2022-10-26
Dirchev: Return SuggestScoreDoc's key for CompletionQuery

dirchev:
# Return `SuggestStoreDoc`'s key for `CompletionQuery`

## Problem

When making searches using the new `CompletionQuery`, we have a use-case in which we want to retrieve the suggest key that was matched with the underlying `PrefixCompletionQuery`. This key is used later for raking.

## Solution

Updated the custom collector `MySuggestScoreDoc` collector to insert a new text field for each hit - `__suggest_key`. The field value is set to the `key` value of the `SuggestScoreDoc` for that hit.

## Example

For example, if we have a document with a `name_suggest` field of type `CONTEXT_SUGGEST`:

```json5
{
    "name_suggest": [
        "{\"value\":\"italian pizza\",\"contexts\":[\"US\"],\"weight\":5}",
        "{\"value\":\"pizza\",\"contexts\":[\"US\"],\"weight\":3}"
    ],
   "name": 
        "value": [
            "Italian Pizza"
        ],
    // ... other fields
}
```

And we make a search request using a `CompletionQuery`

```json5
{
    "indexName": "my_index",
    "startHit": 0,
    "topHits": 5,
    "retrieveFields": ["name"],
    "query": {
        "completionQuery": {
            "field": "name_suggest",
            "queryType": "PREFIX_QUERY",
            "text": "pi",
            "contexts": ["US"]
        }
    }
}
```

We get our matched document in the hit list, along with the new `__suggest_key` which corresponds to the matched key from the `name_suggest` field in addition to all other fields that were requested from `retrieveFields`

```
hits {
  luceneDocId: 26901
  score: 3.0
  fields {
    key: "__suggest_key"
    value {
      fieldValue {
        textValue: "pizza"
      }
    }
  }
  fields {
    key: "name"
    value {
      fieldValue {
        textValue: "{value=[Italian Pizza]}"
      }
    }
  }
}
```

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/MyTopSuggestDocsCollector.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/MyTopSuggestDocsCollector.java
@@ -36,6 +36,7 @@ import org.apache.lucene.search.suggest.document.TopSuggestDocs;
 import org.apache.lucene.search.suggest.document.TopSuggestDocsCollector;
 
 public class MyTopSuggestDocsCollector extends DocCollector {
+  public static final String SUGGEST_KEY_FIELD_NAME = "__suggest_key";
   private static final Comparator<TopSuggestDocs.SuggestScoreDoc> SUGGEST_SCORE_DOC_COMPARATOR =
       (a, b) -> {
         // sort by higher score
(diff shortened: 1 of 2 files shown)


src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/MyTopSuggestDocsCollector.java
```diff
@@ -70,6 +70,17 @@ public void fillHitRanking(SearchResponse.Hit.Builder hitResponse, ScoreDoc scor
     if (!Float.isNaN(scoreDoc.score)) {
       hitResponse.setScore(scoreDoc.score);
     }
+    if (scoreDoc instanceof TopSuggestDocs.SuggestScoreDoc) {
+      // set the retrieval key from the SuggestScoreDoc to the hit response as "__suggest_key" field
+      hitResponse.putFields(
+          "__suggest_key",
```
aprudhomme, line 76: nit: can you make this a static final class constant

aprudhomme (approved):
fix and ship

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/MyTopSuggestDocsCollector.java
dirchev: done

PR #511, opened 2022-10-26
Make deadline for merge precopy configurable and limit merge precopy duration

sarthakn7:
When segments are merged the primary asks replicas to copy the merged segment files before publishing an NRT point. This ensures that the replicas already have the new files (which can be large in size) before the NRT point is published, and thus would be having the same index after publishing the NRT point. However if the files are too large or if there are issues with replicas/network etc. the primary might keep waiting for a long time, blocking commits. In addition if new replicas are being added the primary will wait for the new replicas to finish copying as well which can indefinitely prolong the merge precopy. #507 added a hard-coded deadline for individual copyFiles requests but it is still possible for new replicas to prolong the merge precopy.

We add a live setting called `maxMergePreCopyDurationSec` here which can be used to limit the amount of time spent on precopying merged segments. A deadline is computed at the start of the precopy and all copyFiles requests to existing and new replicas use this deadline. If the deadline is exceeded exceptions would be thrown and the merge will continue.


src/main/java/com/yelp/nrtsearch/server/luceneserver/index/ImmutableIndexState.java
```diff
@@ -162,6 +162,7 @@ public class ImmutableIndexState extends IndexState {
           .setDefaultSearchTimeoutSec(DoubleValue.newBuilder().setValue(0).build())
           .setDefaultSearchTimeoutCheckEvery(Int32Value.newBuilder().setValue(0).build())
           .setDefaultTerminateAfter(Int32Value.newBuilder().setValue(0).build())
+          .setMaxMergePreCopyDurationSec(DoubleValue.newBuilder().setValue(0))
```
aprudhomme, line 165: This is the default value used when the field is not specified in the committed state. There are a few other things to do here:

- Create a class field below to hold the instance value
- Add validation to `validateLiveSettings` method
- Assign instance value in the constructor
- Add abstract getter method to `IndexState`
- Implement method in this class to get instance value (maybe just return a sane default for `LegacyIndexState`)

src/main/java/com/yelp/nrtsearch/server/luceneserver/NRTPrimaryNode.java
```diff
@@ -159,6 +162,10 @@ Iterator<TransferStatus> getTransferStatusIteratorForClient(ReplicationServerCli
     }
   }
 
+  private double getCurrentMaxMergePreCopyDurationSec() {
+    return indexStateManager.getLiveSettings().getMaxMergePreCopyDurationSec().getValue();
```
aprudhomme, line 166: This would be something like `indexStateManager.getCurrent().getMaxMergePreCopyDurationSec()`

aprudhomme (commented):
You will probably also want to add this property to the liveSettingsV2 cli command

src/main/java/com/yelp/nrtsearch/server/luceneserver/NRTPrimaryNode.java
```diff
@@ -122,21 +127,24 @@ static class MergePreCopy {
...
+      this.deadline = deadline;
     }
 
     public synchronized boolean tryAddConnection(
         ReplicationServerClient c, String indexName, long primaryGen, FilesMetadata filesMetadata) {
-      if (!finished) {
+      if (!finished && (deadline == null || !deadline.isExpired())) {
         Iterator<TransferStatus> transferStatusIterator =
-            c.copyFiles(indexName, primaryGen, filesMetadata);
+            c.copyFiles(indexName, primaryGen, filesMetadata, deadline);
```
aprudhomme, line 147: I'm not sure if this is the right logic here. There is still a race condition where the deadline can expire before the request is initiated. What is the behavior if this happens, does it immediately throw an exception or would it happen when the iterator is consumed? If it is the former, we would probably want to catch and ignore the exception here, so it doesn't bubble up.

aprudhomme (commented):
Might be good to add _set/_default/_invalid tests for the new property to `ImmutableIndexStateTest` class.

Also needs a gateway regen.

src/main/java/com/yelp/nrtsearch/server/luceneserver/NRTPrimaryNode.java
sarthakn7: @aprudhomme I tested this particular condition by modifying `ReplicationServerTest.copyFiles`. The error is thrown when the iterator is consumed:
```
DEADLINE_EXCEEDED: ClientCall started after deadline exceeded. Deadline exceeded after -1.004448333s. 
io.grpc.StatusRuntimeException: DEADLINE_EXCEEDED: ClientCall started after deadline exceeded. Deadline exceeded after -1.004448333s. 
	at app//io.grpc.Status.asRuntimeException(Status.java:535)
	at app//io.grpc.stub.ClientCalls$BlockingResponseStream.hasNext(ClientCalls.java:660)
	at app//com.yelp.nrtsearch.server.grpc.ReplicationServerTest.copyFiles(ReplicationServerTest.java:151)
```
You can see that even though the error says `ClientCall started after deadline exceeded` the error was thrown in `hasNext()`

aprudhomme: With that behavior, a try-catch shouldn't be needed

aprudhomme (approved):

PR #514, opened 2022-11-07
Clean up legacy state code

aprudhomme:
Remove the legacy state system code. This includes the legacy index/global state classes and all special handling logic. Tests have been update/removed as needed.

As part of fixing tests, some minor functional changes were made:
- The taxonomy index is always opened as `CREATE_OR_APPEND`. The taxonomy index generally has inconsistent support, and doesn't work with nrt replication or incremental backups. We should evaluate more robust support/testing as a follow up action.
- The data index will not be opened as `CREATE` if there is restored data. Restored data must have a valid index commit, so this should be safe.

A small set of tests have been marked as ignored:
- Adding index virtual fields with custom (not javascript) script types. Supporting this will require some refactoring of the doc value accessor and can be done as a follow up task. It is also not a widely used feature.
- Suggest tests. The original suggest support did its own index management, but still stored some state in index state. This was not ported to the new state system, since it is unclear if this is the right way to go about suggest support in general. We should explicitly decide to either add the functionality back, or remove it entirely.


taoyyu:
looks good to me. Just one question, would this pr conflicts other prs? There are 5 prs pending/approved but not yet merged. Should we check all pending prs, merge or close them first and then merge this one?

taoyyu (approved):

PR #517, opened 2022-11-12
Delete Replica unused last commit files

taoyyu:
Tried the idea to decRef of files not used. Need to add some test but might not be straightforward.

Unfortunately replicaNode.lastCommitFiles is not accessible since it is private. I would prefer using that instead of dir.listAll

Let me know if I should change something

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/NRTReplicaNode.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/NRTReplicaNode.java
@@ -15,6 +15,7 @@
  */
 package com.yelp.nrtsearch.server.luceneserver;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.yelp.nrtsearch.server.grpc.FileMetadata;
 import com.yelp.nrtsearch.server.grpc.FilesMetadata;
 import com.yelp.nrtsearch.server.grpc.GetNodesResponse;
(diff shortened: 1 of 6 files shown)


src/main/java/com/yelp/nrtsearch/server/luceneserver/NRTReplicaNode.java
```diff
@@ -213,6 +217,9 @@ public CopyJob launchPreCopyFiles(
   @Override
   protected void finishNRTCopy(CopyJob job, long startNS) throws IOException {
     super.finishNRTCopy(job, startNS);
+    if (replicaDeleterManager != null) {
+      replicaDeleterManager.cleanUpReplicaFiles();
+    }
```
aprudhomme, line 222: Since we do not call ReplicaNode::commit, the initial commit files only really need to be decremented once. This could be done in the NRTReplicaNode::start method, which should be made synchronized. After calling super.start, read the current commit files out of the directory using the same logic as ReplicaNode::start. Decrementing the reference for these files should be all that needs to be done, as the current nrt point files will still be referenced.

aprudhomme (commented):
It would be good to have some kind of test to make sure the clean up works as expected.

src/main/java/com/yelp/nrtsearch/server/luceneserver/NRTReplicaNode.java
taoyyu: I'm writing test for it now. Quick question: I am trying to use [TestServer class](https://github.com/Yelp/nrtsearch/blob/master/src/test/java/com/yelp/nrtsearch/server/grpc/TestServer.java). But it seems it doesn't have the NRT functionality? I am trying to launch primary and replicas, send some docs to primary and commit refresh. Seems replica cannot see the updates?

aprudhomme (approved):

PR #519, opened 2022-11-18
invalidate empty index name and add unit test

xinyiyelp:
The server still tries to launch an index when its name is empty, which was never created before. 

The work here is to invalidate empty index name before everything starts.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
+++ b/src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
@@ -721,9 +721,7 @@ public class LuceneServer {
         responseObserver.onError(
             Status.INVALID_ARGUMENT
                 .withDescription(
-                    String.format(
-                        "error while trying to start index since indexName was empty: "
-                            + startIndexRequest.getIndexName()))
+                    String.format("error while trying to start index since indexName was empty."))
                 .asRuntimeException());
         return;
       }
--- a/src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
+++ b/src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
@@ -717,7 +717,7 @@ public class LuceneServer {
         StartIndexRequest startIndexRequest, StreamObserver<StartIndexResponse> responseObserver) {
       logger.info("Received start index request: {}", startIndexRequest);
       if (startIndexRequest.getIndexName().isEmpty()) {
-        logger.warn("error while trying to start index " + startIndexRequest.getIndexName());
+        logger.warn("error while trying to start index with empty index name.");
         responseObserver.onError(
             Status.INVALID_ARGUMENT
                 .withDescription(
--- a/src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
+++ b/src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
@@ -717,9 +717,13 @@ public class LuceneServer {
         StartIndexRequest startIndexRequest, StreamObserver<StartIndexResponse> responseObserver) {
       logger.info("Received start index request: {}", startIndexRequest);
       if (startIndexRequest.getIndexName().isEmpty()) {
+        logger.warn("error while trying to start index " + startIndexRequest.getIndexName());
         responseObserver.onError(
             Status.INVALID_ARGUMENT
-                .withDescription(String.format("Index name is empty - must not be empty"))
+                .withDescription(
+                    String.format(
+                        "error while trying to start index since indexName was empty: "
+                            + startIndexRequest.getIndexName()))
                 .asRuntimeException());
         return;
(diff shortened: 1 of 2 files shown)


src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
```diff
@@ -716,6 +716,13 @@ public void settingsV2(
     public void startIndex(
         StartIndexRequest startIndexRequest, StreamObserver<StartIndexResponse> responseObserver) {
       logger.info("Received start index request: {}", startIndexRequest);
+      if (startIndexRequest.getIndexName().isEmpty()) {
+        responseObserver.onError(
```
swekannan, line 720: You can add a logger.warn() before this error handling to add info on the indexName for which this failed. And make the error more descriptive such as "error while trying to start index since indexName was empty: " + indexName

src/test/java/com/yelp/nrtsearch/server/grpc/LuceneServerTest.java
```diff
@@ -277,6 +277,24 @@ public void testStartShard() throws IOException {
...
 
+  @Test
+  public void testStartIndex() {
+    String testIndex = grpcServer.getTestIndex();
+    LuceneServerGrpc.LuceneServerBlockingStub blockingStub = grpcServer.getBlockingStub();
+    // create the index
+    blockingStub.createIndex(CreateIndexRequest.newBuilder().setIndexName(testIndex).build());
+    try {
+      // start the index
+      String emptyTestIndex = "";
```
swekannan, line 288: Also test cases where the indexName is not set in addition to testing for empty strings.

src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
xinyiyelp: Yup, added warning with descriptive message

src/test/java/com/yelp/nrtsearch/server/grpc/LuceneServerTest.java
xinyiyelp: Added another test case: launch the index with empty string before creating a index.

sarthakn7: Can we add these two different test cases in two different functions? The create index test can be moved into a test called `testCreateIndexWithEmptyString`

src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
sarthakn7: Empty index name in an error might be confusing (and it's always going to be ""). Maybe skip the index name or log the start index request?

swekannan: Logging the startIndex request sounds like a better option. Will it show when the indexName string is not set vs an empty string since it seems like protobuf sets string defaults to ""?

sarthakn7: It doesn't show unset only empty unfortunately

xinyiyelp: Thank you for reviewing this! @sarthakn7 @swekannan 
Since line 718 has already logged out the `startIndexRequest`, here I skipped the index name and warned it's an empty index name.

src/test/java/com/yelp/nrtsearch/server/grpc/LuceneServerTest.java
xinyiyelp: It sounds a good idea, I splitted the unit tests into two.
For the naming, I name it more specifically: `testCreateAndStartIndexWithEmptyString` since we are actually testing the start index functionality.

sarthakn7: Ohh my bad @swekannan I thought we were testing the createIndex as well. Yeah just test for startIndex is fine if we have not changed anything in the create.

src/test/java/com/yelp/nrtsearch/server/grpc/LuceneServerTest.java
```diff
@@ -277,6 +277,41 @@ public void testStartShard() throws IOException {
...
+    } catch (StatusRuntimeException e) {
+      assertEquals(
+          String.format(
+              "INVALID_ARGUMENT: error while trying to start index since indexName was empty: "),
+          e.getMessage());
+    }
+  }
+
+  @Test
+  public void testCreateAndStartIndexWithEmptyString() {
```
swekannan, line 297: Hi @xinyiyelp you can remove this test then and only keep the startIndex test. You can then re-request review for approval.

swekannan (approved):

PR #521, opened 2022-11-24
support wildcard query

xinyiyelp:
Currently, you need to specify all fields names if you want to retrieve in the nrtsearch search request. This PR enables you to retrieve all index fields with `wild card` option.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/SearchRequestProcessor.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/SearchRequestProcessor.java
@@ -50,7 +50,6 @@ import com.yelp.nrtsearch.server.luceneserver.search.collectors.SortFieldCollect
 import com.yelp.nrtsearch.server.utils.ScriptParamsUtils;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -206,7 +205,8 @@ public class SearchRequestProcessor {
   private static Map<String, FieldDef> getRetrieveFields(
       SearchRequest request, Map<String, FieldDef> queryFields) {
     Map<String, FieldDef> retrieveFields = new HashMap<>();
-    if (request.getRetrieveFieldsList().equals(Arrays.asList("*"))) {
+    final String WILDCARD = "*";
+    if (request.getRetrieveFieldsCount() == 1 && request.getRetrieveFields(0).equals(WILDCARD)) {
       return queryFields;
     }
     for (String field : request.getRetrieveFieldsList()) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/yelp/nrtsearch/server/luceneserver/search/SearchRequestProcessor.java
```diff
@@ -205,6 +206,9 @@ private static Map<String, FieldDef> getVirtualFields(
   private static Map<String, FieldDef> getRetrieveFields(
       SearchRequest request, Map<String, FieldDef> queryFields) {
     Map<String, FieldDef> retrieveFields = new HashMap<>();
+    if (request.getRetrieveFieldsList().equals(Arrays.asList("*"))) {
```
swekannan, line 209: This would create a new single element list each time this code gets executed. You can instead check the first element of the list without any new object creation.
Also you can declare the WILDCARD string as a static variable and use it here.
`request.getRetrieveFieldsCount() == 1 && request.getRetrieveFields(0).equals(WILDCARD)`

src/test/java/com/yelp/nrtsearch/server/grpc/LuceneServerTest.java
```diff
@@ -854,6 +856,56 @@ public void testSearchBasic() throws IOException, InterruptedException {
...
+
+    assertEquals(2, searchResponse.getTotalHits().getValue());
+    assertEquals(2, searchResponse.getHitsList().size());
+    SearchResponse.Hit firstHit = searchResponse.getHits(0);
+    checkHits(firstHit);
+    SearchResponse.Hit secondHit = searchResponse.getHits(1);
+    checkHits(secondHit);
+
+    // support format of "*"
+    SearchResponse searchResponsePlainStar =
```
swekannan, line 890: Do you think this variable would be more descriptive with something like `searchResponseWithWildcard`?

xinyiyelp: This sounds a better naming, changed.

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/SearchRequestProcessor.java
xinyiyelp: This looks more efficient, thanks!

src/test/java/com/yelp/nrtsearch/server/grpc/LuceneServerTest.java
```diff
@@ -854,6 +856,56 @@ public void testSearchBasic() throws IOException, InterruptedException {
...
+                SearchRequest.newBuilder()
+                    .setIndexName(grpcServer.getTestIndex())
+                    .setStartHit(0)
+                    .setTopHits(10)
+                    .addRetrieveFields("*")
+                    .build());
+
+    assertEquals(2, searchResponseWithWildcard.getTotalHits().getValue());
+    assertEquals(2, searchResponseWithWildcard.getHitsList().size());
+    SearchResponse.Hit firstHitPlainStar = searchResponseWithWildcard.getHits(0);
```
swekannan, line 903: nit: plainStar in the variable name does not seem very descriptive.

xinyiyelp: Thanks, changed the var names to `firstHitWithArrOfWildcard` and `firstHitWithStrOfWildcard` respectively, how do they sound now?

PR #524, opened 2022-12-02
Call weight.getQuery().toString() once per query instead of once per leaf

sarthakn7:
`toString` on a query object can be significantly expensive for large queries. Trying to avoid doing it multiple times for a single query. I also changed the order of the calls so the toString would not be called at all for CompletionQuery.

It would be better to avoid calling toString at all....but I couldn't find any test for which `weight.getQuery().toString().contains("DrillSidewaysQuery")` is true.


aprudhomme (approved):
This can be helpful short term, but we should probably evaluate if this workaround is really needed anymore. I'm guessing this was from the original luceneserver code and may not apply to a newer lucene version or how we do faceting

sarthakn7:
@aprudhomme agreed. Would be good to confirm if we ever enter that branch other than for completion queries.

PR #525, opened 2022-12-06
Add MatchPhrasePrefixQuery and multi match support

aprudhomme:
Add phrase prefix support to the multi match query. This was done by adding the `MatchPhrasePrefixQuery` query type, then adding the option to use it in multi match.

The phrase prefix query is built by analyzing the query text, then expanding the prefix term into a number of concrete terms from the index. All this is used to create a lucene `MultiPhraseQuery`. Much of the query building logic comes from the [Elasticsearch version](https://github.com/elastic/elasticsearch/blob/v7.2.0/server/src/main/java/org/elasticsearch/common/lucene/search/MultiPhrasePrefixQuery.java).

Tests have been added for `TEXT` and `ATOM` fields.


sarthakn7 (commented):
Lgtm. Can we also add/update the query documentation?

sarthakn7 (approved):

PR #528, opened 2022-12-22
Added local rst documentation generation using gradle plugin + some documentation

sarthakn7:
Using a gradle plugin to generate html for the rst files under docs. I've also added documentation for the same under a new "Development" section in the docs and also added instructions to add a new live setting to nrtsearch.

The generated documentation looks like this:
Screenshot 2022-12-21 at 7 06 46 PM
Screenshot 2022-12-22 at 2 31 07 PM

### DIFF
--- a/docs/development/add_live_setting.rst
+++ b/docs/development/add_live_setting.rst
@@ -1,7 +1,7 @@
 Adding A New Live Setting
 ==========================
 
-* Add new live setting to ``IndexSettings`` in ``luceneserver.proto`` and regenerate the protobuf files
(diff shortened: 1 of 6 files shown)


docs/development/add_live_setting.rst
```diff
@@ -0,0 +1,15 @@
+Adding A New Live Setting
+==========================
+
+* Add new live setting to ``IndexSettings`` in ``luceneserver.proto`` and regenerate the protobuf files
```
aprudhomme, line 4: It would be good to mention that all new fields must be messages, with primitives being wrapped

aprudhomme (approved):

PR #531, opened 2023-01-17
Add filter collector

aprudhomme:
Add a filtering collector aggregation that only passes through select documents to nested collectors.

The filter can be specified in one of two ways:
- `Query`: Given an arbitrary nrtsearch query, only pass documents this query matches
- `TermInSetQuery`: Specialized filter for a set query. Passes documents if any of the doc values are present in the query set. This can be more efficient if the query set is large and recall is relatively small, since building the scorer may be expensive in the query approach.


taoyyu (approved):
looks sane to me for the optimization. Let's push this out since this seems to be necessary for the social rollout.

I'm wondering if we can optimize filter case for the main query as well. But that may be trickier

swekannan (approved):
Looks good to me. Since it is a new collector, do we want to add a test for a large term set for some performance numbers?

PR #533, opened 2023-01-19
RP-6869: Add option to filter incompatible segment readers during refresh

aprudhomme:
This change is intended to address an issue when using incremental backups to remote storage. The remote storage is the source of truth for index data, but replicas may have uncommitted data distributed directly through gRPC nrt points. When the primary restarts, it pulls the last commit from remote storage. As the new primary does not know about uncommitted segments from the old primary, segment names may be reused. The segment id conflict keeps the replicas from opening new searcher versions until all conflicting files one the new primary are merged away.

## Solution
Add option to filter incompatible segment readers during searcher refresh. Typically, all previous segment readers are passed to the reopen operation, so that they can be reused as needed. Instead, create a new implementation of `SegmentInfosSearcherManager` that preemptively checks for segment reader conflicts and filters them. The reopen operation will create a new reader from the new data on disk.

## Limitations
Even with this option, replicas may not immediately sync from a new primary. The replica index version may be higher, which will make them ignore nrt points from the new primary until the index version eclipses that of the old primary.

## Testing
Tests have been added:
- Replica can sync from restarted primary, even when there are segment conflicts
- Older searcher versions are still usable after the conflict is resolved
- Old version of conflicting segments have their reader closed during searcher cleanup

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/ShardState.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/ShardState.java
@@ -903,6 +903,7 @@ public class ShardState implements Closeable {
       throw new IllegalStateException("index \"" + name + "\" was already started");
     }
     IndexState indexState = indexStateManager.getCurrent();
+    LuceneServerConfiguration configuration = indexState.getGlobalState().getConfiguration();
 
     // nocommit share code better w/ start and startPrimary!
     try {
@@ -913,10 +914,7 @@ public class ShardState implements Closeable {
         indexDirFile = rootDir.resolve(INDEX_DATA_DIR_NAME);
       }
       origIndexDir =
-          indexState
-              .getDirectoryFactory()
-              .open(
-                  indexDirFile, indexState.getGlobalState().getConfiguration().getPreloadConfig());
+          indexState.getDirectoryFactory().open(indexDirFile, configuration.getPreloadConfig());
       // nocommit don't allow RAMDir
       // nocommit remove NRTCachingDir too?
       if ((origIndexDir instanceof MMapDirectory) == false) {
@@ -934,13 +932,12 @@ public class ShardState implements Closeable {
       manager = null;
       nrtPrimaryNode = null;
 
-      boolean verbose = indexState.getGlobalState().getConfiguration().getIndexVerbose();
+      boolean verbose = configuration.getIndexVerbose();
 
       HostPort hostPort =
           new HostPort(
               indexState.getGlobalState().getHostName(),
               indexState.getGlobalState().getReplicationPort());
-      LuceneServerConfiguration configuration = indexState.getGlobalState().getConfiguration();
       nrtReplicaNode =
           new NRTReplicaNode(
               indexState.getName(),
@@ -959,7 +956,7 @@ public class ShardState implements Closeable {
         nrtReplicaNode.startWithLastPrimaryGen();
       }
 
-      if (indexState.getGlobalState().getConfiguration().getSyncInitialNrtPoint()) {
+      if (configuration.getSyncInitialNrtPoint()) {
         nrtReplicaNode.syncFromCurrentPrimary(
             INITIAL_SYNC_PRIMARY_WAIT_MS, INITIAL_SYNC_MAX_TIME_MS);
       }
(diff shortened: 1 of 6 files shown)


src/main/java/org/apache/lucene/replicator/nrt/FilteringSegmentInfosSearcherManager.java
```diff
@@ -0,0 +1,132 @@
...
+      for (int i = 0; i < old.getIndexReader().leaves().size(); ++i) {
+        final SegmentReader sr = (SegmentReader) old.getIndexReader().leaves().get(i).reader();
+        oldReadersMap.put(sr.getSegmentName(), i);
+      }
+      subs = new ArrayList<>();
+      for (SegmentCommitInfo commitInfo : newInfos) {
+        Integer oldReaderIndex = oldReadersMap.get(commitInfo.info.name);
+        if (oldReaderIndex != null) {
+          SegmentReader oldReader =
+              (SegmentReader) old.getIndexReader().leaves().get(oldReaderIndex).reader();
```
sarthakn7, line 88: `indexReader.leaves()` doesn't look like a simple getter. Can we call it just once right after `oldReadersMap` is defined and store in a variable? There are checks that the IndexReader is still open within those function calls, but I doubt that the index will be closed during this.

src/test/java/com/yelp/nrtsearch/server/grpc/PrimaryRestartTests.java
```diff
@@ -110,7 +120,377 @@ public void testPrimaryRestartReplication() throws IOException {
...
+        .getShard(0)
+        .slm
+        .release(previousSearcher1);
+
+    primaryServer.addSimpleDocs("test_index", 6, 7, 8);
+    primaryServer.refresh("test_index");
+
+    replicaServer.waitForReplication("test_index");
+    replicaServer.verifySimpleDocIds("test_index", 1, 2, 3, 4, 5, 6, 7, 8);
+    long previousSearcherVersion2 =
```
sarthakn7, line 338: The searcher versions and searchers are retrieved and searcher released several times in this test. Can extract them into methods to make this more readable.

src/test/java/com/yelp/nrtsearch/server/grpc/PrimaryRestartTests.java
```diff
@@ -110,7 +120,377 @@ public void testPrimaryRestartReplication() throws IOException {
...
+
+    verifySimpleDocIdsForVersion(
+        replicaServer, previousSearcherVersion1, "test_index", 1, 2, 3, 4, 5);
+    verifySimpleDocIdsForVersion(
+        replicaServer, previousSearcherVersion2, "test_index", 1, 2, 3, 4, 5, 6, 7, 8);
+
+    verifySegmentsClosed(previousLeaves1);
+    verifySegmentsClosed(previousLeaves2);
+    verifySegmentsClosed(currentLeaves1);
+    verifySegmentsClosed(currentLeaves2);
```
sarthakn7, line 444: When first reading this test without reading the `verifySegmentsClosed` method it appears as if we are checking that all segments in these leaves are closed. Maybe add a comment before this block saying that we're checking that the segments are still open.

src/test/java/com/yelp/nrtsearch/server/grpc/PrimaryRestartTests.java
```diff
@@ -110,7 +120,377 @@ public void testPrimaryRestartReplication() throws IOException {
...
+
+    Thread.sleep(3000);
+    replicaServer.getGlobalState().getIndex("test_index").getShard(0).slm.prune(new PruneByAge(1));
+
+    verifySegmentsClosed(previousLeaves1, "_1");
+    verifySegmentsClosed(previousLeaves2, "_1", "_2");
+    verifySegmentsClosed(currentLeaves1);
+    verifySegmentsClosed(currentLeaves2);
```
sarthakn7, line 452: It might be useful to verify exactly what segments are open

src/main/java/com/yelp/nrtsearch/server/luceneserver/ShardState.java
```diff
@@ -939,6 +940,7 @@ public synchronized void startReplica(ReplicationServerClient primaryAddress, lo
           new HostPort(
               indexState.getGlobalState().getHostName(),
               indexState.getGlobalState().getReplicationPort());
+      LuceneServerConfiguration configuration = indexState.getGlobalState().getConfiguration();
```
sarthakn7, line 943: There are two more usages of this variable before. This can be defined before line 915 and and used in those places too.

src/main/java/org/apache/lucene/replicator/nrt/FilteringSegmentInfosSearcherManager.java
```diff
@@ -80,12 +81,12 @@ protected IndexSearcher refreshIfNeeded(IndexSearcher old) throws IOException {
         final SegmentReader sr = (SegmentReader) old.getIndexReader().leaves().get(i).reader();
         oldReadersMap.put(sr.getSegmentName(), i);
       }
+      List<LeafReaderContext> leaves = old.getIndexReader().leaves();
```
sarthakn7, line 84: Can this be done before line 80? There are more usages before

sarthakn7 (approved):

PR #541, opened 2023-01-27
Extract code for s3 download from ArchiverImpl

sarthakn7:
Extracting the code to download from S3 from ArchiverImpl for easy re-use. I'm going to use this to download plugins from S3 but there might be other future use-cases as well.
I have also created an S3 module which initializes AmazonS3 object.


src/main/java/com/yelp/nrtsearch/server/utils/S3Downloader.java
```diff
@@ -0,0 +1,152 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class S3Downloader {
+  private static final Logger logger = LoggerFactory.getLogger(S3Downloader.class);
+
+  public static final int NUM_S3_THREADS = 20;
+
+  private final ThreadPoolExecutor executor =
+      (ThreadPoolExecutor) Executors.newFixedThreadPool(NUM_S3_THREADS);
```
aprudhomme, line 42: Instead of creating a new thread pool for every instance, could this (optionally?) be an argument. At the very least, it might be good to be able to set the number of threads.

src/main/java/com/yelp/nrtsearch/server/utils/S3Downloader.java
```diff
@@ -0,0 +1,152 @@
...
+import java.io.SequenceInputStream;
+import java.util.Enumeration;
+import java.util.LinkedList;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.ThreadPoolExecutor;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class S3Downloader {
```
aprudhomme, line 36: If this class creates a thread pool, it might be good to make it `Closeable` to clean up the resources.

aprudhomme (approved):

PR #543, opened 2023-02-03
Add a simple example repo of an nrtsearch plugin

sarthakn7:
Add a module that acts as an example for nrtsearch plugins. This module doesn't have the main nrtsearch module as it's parent and instead includes it as a dependency so that it is easy to create plugins from this module.
This example can be enhanced further in the future to have more plugin implementations.


example-plugin/build.gradle
```diff
@@ -0,0 +1,70 @@
...
+            from('src/main/plugin-metadata/plugin-metadata.yaml') {
+                expand(
+                        classname: 'com.yelp.nrtsearch.plugins.example.ExamplePlugin',
+                        name: "nrtsearch-plugin-example",
+                        version: project.version,
+                        description: 'Example plugin for nrtsearch',
+                        server_version: "${nrtsearchVersion}"
+                )
+            }
+            from('src/main/plugin-metadata/plugin-security.policy')
```
aprudhomme, line 53: I think we can remove this policy file. I had thought we might want to add enforcement at some point, but given that the security manager was deprecated in java 17, I don't think that will happen.

example-plugin/build.gradle
```diff
@@ -0,0 +1,70 @@
...
+    testImplementation "junit:junit:4.13"
+    testImplementation "org.assertj:assertj-core:3.19.0"
+}
+
+distributions {
+    main {
+        contents {
+            from(jar)
+            from(this.project.configurations.runtimeClasspath)
+            from('src/main/plugin-metadata/plugin-metadata.yaml') {
```
aprudhomme, line 44: If there is a way to have gradle fully generate this file it might be cleaner, but that doesn't necessarily need to be solved now.

example-plugin/src/main/resources/register_fields.json
```diff
@@ -0,0 +1,14 @@
+{
```
aprudhomme, line 1: Is there a reason this file is a main resource and not a test resource?

sarthakn7: ++ good catch, moved it to test

sarthakn7:
I've removed java 14 from github workflow since the compatibility for this example-plugin is set to 17.

aprudhomme (approved):

fragosoluana (approved):

PR #544, opened 2023-02-03
Add highlight plugin service

waziqi89:
RP-7396
Add Highlighter Plugin Service to support custom highlighter. (Next step: add support of plain highlighter)

General request flow:

1. add highlightFetchTask and feed in the corresponding highlighter from the highlight service
2. highlightFetchTask will check the highlightability of the request at the contrunction time
3. highlightFetchTask will be executed once at the fetch phase during the search execution flow, in which the previous assigned highlighter will be called.

More details:
- We added a new service - highlighter service to initialize the plugins at the startup time and register them, so that later on,  the search request processor can select the corresponding highlighter out of the registry. (The fvh is the only builtin highlighter atm.)
- We also define the highlighter interface and the highlightPlugin interface to provide the list of the custom highlighters. Each highlighter must implement `getHighlights` to do highlighting; `verifyFieldIsSupported` to check the highlightablility for the specific implementation - eg. termvector_order_position for fvh


clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -792,35 +792,34 @@ message FilterResult {
...
-        // Specify a query here if highlighting is desired against a different query than the search query
-        Query highlight_query = 6;
-        // Set to true to highlight fields only if specified in the search query
-        bool field_match = 7;
-    }
-
-    // Highlight settings
-    Settings settings = 1;
-    // Fields to highlight
-    repeated string fields = 2;
```
sarthakn7, line 823: Can you please revert these changes? These fields are already in use and modifying them will break the clients. Additionally please follow the [standard protobuf style](https://developers.google.com/protocol-buffers/docs/style) for new fields in this message which specifies using `underscore_separated_names` for field names. We have not followed the standard style in the past but want to do it for new proto objects.

src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
```diff
@@ -350,6 +351,7 @@ private void initExtendableComponents(
       RescorerCreator.initialize(configuration, plugins);
       ScriptService.initialize(configuration, plugins);
       SimilarityCreator.initialize(configuration, plugins);
+      HighlighterService.initialize(configuration, plugins);
```
sarthakn7, line 354: Please add this in alphabetical order

sarthakn7 (changes requested):
I have taken a quick look since this is a draft, but I need a lot more context on these changes. Can you please add it to the PR description? If you can add more details on the changes in the design, what existing classes are being replaced and why that would also be very helpful. Also I think you're referring to ticket RP-7396 not 7395.

src/main/java/com/yelp/nrtsearch/server/luceneserver/highlights/HighlightFetchTask.java
```diff
@@ -95,4 +112,30 @@ public void processHit(SearchContext searchContext, LeafReaderContext hitLeaf, B
...
+                "Field %s is not a text field and does not support highlights", fieldName));
+      }
+      if (!((TextBaseFieldDef) field).isSearchable()) {
+        throw new IllegalArgumentException(
+            String.format("Field %s is not searchable and cannot support highlights", fieldName));
+      }
+      if (!((TextBaseFieldDef) field).isStored()) {
+        throw new IllegalArgumentException(
+            String.format("Field %s is not stored and cannot support highlights", fieldName));
+      }
```
sarthakn7, line 137: I think this check is specific to fast vector highlighter (Unified does not necessarily need it) and can be moved to its class

clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -823,4 +824,36 @@ message Highlight {
...
+        // Number of characters in highlighted fragment, 100 by default.
+        google.protobuf.UInt32Value fragment_size = 4;
+        // Maximum number of highlight fragments to return, 5 by default. If set to 0 returns entire text as a single fragment ignoring fragmentSize.
+        google.protobuf.UInt32Value max_number_of_fragments = 5;
+        // Specify a query here if highlighting is desired against a different query than the search query.
+        Query highlight_query = 6;
+        // Set to true to highlight fields only if specified in the search query.(Default is false)
+        google.protobuf.BoolValue field_match = 7;
+        // Sorts highlighted fragments by score when set to true. By default, fragments will be output in the order they appear in the field. (Default is false)
+        google.protobuf.BoolValue score_ordered = 8;
```
sarthakn7, line 846: Can we have the default behavior as score_ordered?

src/main/java/com/yelp/nrtsearch/server/luceneserver/highlights/FastVectorHighlighter.java
```diff
@@ -0,0 +1,141 @@
...
+  private static Logger logger = LoggerFactory.getLogger(FastVectorHighlighter.class);
+
+  public static FastVectorHighlighter getInstance() {
+    return INSTANCE;
+  }
+
+  static {
+    // always treat multivalue field discretely
+    SIMPLE_FRAGMENTS_BUILDER.setDiscreteMultiValueHighlighting(true);
+    SCORE_ORDER_FRAGMENTS_BUILDER.setDiscreteMultiValueHighlighting(true);
```
sarthakn7, line 65: Can we skip any changes to the behavior of fast vector highlighter as part of this PR? We would need to check how this would affect existing use-cases.
You can add more context in description on any changes in Fast vector highlighter if they are needed as part of this PR.

sarthakn7 (commented):
Please also update the documentation for highlighting for any changes to the query or new fields https://github.com/Yelp/nrtsearch/blob/master/docs/highlighting.rst

clientlib/src/main/proto/yelp/nrtsearch/search.proto
waziqi89: order = none is default in ES, I was trying to match it. However, v1 default is true, v2 default is false introduces confusion to clients. I will change it to be true for both

clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -823,4 +824,36 @@ message Highlight {
...
+        // Select Fragmenter between span (DEFAULT) and simple. This is only applicable for plain highlighters.
+        google.protobuf.StringValue fragmenter = 9;
+
+    }
+    // Specify the highlighter name to use. Each highlighter name must be unique. The default is "fast-vector-highlighter"
+    string highlighter_type = 1;
```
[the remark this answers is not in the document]
waziqi89: I feel it's more than that. 
highlighter_type is now at the top level, not in settings. (doesn't make sense to be overwrite-able for each filed)
match_field now is a wrapped bool for field to overwrite
Also as we are still under dev, more breaking changes might come in later.

waziqi89:
> Please also update the documentation for highlighting for any changes to the query or new fields https://github.com/Yelp/nrtsearch/blob/master/docs/highlighting.rst

I will update the docs immediately after this PR merged. The PR is not finalized yet, adding docs now is hard.
Additionally, as v1 is still 100% valid after this PR, we can temporarily keep the docs as it is between the PR merge and the read the docs update. This is another + for having all new settings separate in v2.

src/main/java/com/yelp/nrtsearch/server/luceneserver/highlights/Highlighter.java
```diff
@@ -0,0 +1,39 @@
...
+
+import com.yelp.nrtsearch.server.luceneserver.field.TextBaseFieldDef;
+import java.io.IOException;
+import java.util.Map;
+import org.apache.lucene.index.IndexReader;
+
+public interface Highlighter {
+  String getName();
+
+  String[] getHighlights(
```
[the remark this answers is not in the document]
sarthakn7: Unless there is some harm in exposing the cache from SearchContext or benefit in isolating it to highlights, I don't see why we should create another context object just to store a map. I also think that by adding to SearchContext it will be available to other plugins and other use-cases as well.

src/main/java/com/yelp/nrtsearch/server/luceneserver/highlights/FastVectorHighlighter.java
```diff
@@ -0,0 +1,138 @@
...
+import org.apache.lucene.search.vectorhighlight.SingleFragListBuilder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * The FVH highlighter class for a search query. The highlighted field must be indexed with term
+ * vector with positions and offsets. And the multivalue fields are always fragmented and scored
+ * discretely.
+ */
+public class FastVectorHighlighter implements Highlighter {
```
taoyyu, line 40: I think it might be good to rename this class like FastVectorHighlighterProvider or some name with NRT as prefix to avoid having the same name as org.apache.lucene.search.vectorhighlight.FastVectorHighlighter. But it is totally your call since Elasticsearch is actually using the same name in their repo 😀

clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -813,8 +816,19 @@ message Highlight {
...
+        // Set to true to highlight fields only if specified in the search query. (Default is false)
+        google.protobuf.BoolValue field_match_v2 = 8;
+        // Sorts highlighted fragments by score when set to true. By default, fragments will be output in the order they appear in the field. (Default is true)
+        google.protobuf.BoolValue score_ordered = 9;
+        // Select Fragmenter between span (default) and simple. This is only applicable for plain highlighters.
+        google.protobuf.StringValue fragmenter = 10;
+        // Let the fragment builder respect the multivalue fields. Each fragment won't cross multiple value fields if set true. (Default is false)
+        google.protobuf.BoolValue discrete_multivalue = 11;
+        // When highlighter_type is CUSTOM, use this string identifier to specify the highlighter. It is ignored for any other highlighter_types.
+        string custom_highlighter_name = 12;
```
taoyyu, line 831: One question. As a plugin here, do we expect the user to pass custom defined params to the highlighter?

waziqi89: In current design, the highlighter is a singleton instance handling all requests. We do not create highlighter per request, so we shouldn't alter it per request. Otherwise it may cause non idempotent issues.

But maybe I am not aware of some use cases where the highlighter must be altered during the execution. Please let me know.

src/main/java/com/yelp/nrtsearch/server/luceneserver/highlights/FastVectorHighlighter.java
waziqi89: Agree. We can add NRT prefix to all builtin highlighters. but the `customHighlighterName`

waziqi89: plainHighlighter actually has the same issue. Lucene calls it Highlighter, and ES add prefix"plain" to distinguish them

clientlib/src/main/proto/yelp/nrtsearch/search.proto
taoyyu: Not so far. Just a random question here since this is an extensible plugin. But settings should be good enough

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/SearchContext.java
```diff
@@ -48,6 +48,7 @@ public class SearchContext implements FieldFetchContext {
   private final List<RescoreTask> rescorers;
   private final SharedDocContext sharedDocContext;
   private final HighlightFetchTask highlightFetchTask;
+  private final Map<String, Object> highlighterContext;
```
sarthakn7, line 51: I suggest renaming this to something like `extraContext` or `cache` or `customContext` or something more general since even though this is called `highlighterContext` it is totally possible for any user of SearchContext to add any custom context in here. Also would be good to add a comment here saying that this is a common context that can be used by plugins otherwise someone reading this code would think that the field is unused and can be removed.

clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -813,8 +816,19 @@ message Highlight {
         google.protobuf.UInt32Value max_number_of_fragments = 5;
         // Specify a query here if highlighting is desired against a different query than the search query
         Query highlight_query = 6;
-        // Set to true to highlight fields only if specified in the search query
-        bool field_match = 7;
+        // Set to true to highlight fields only if specified in the search query.
+        // This field is identical to field_match_v2 but doesn't support field override. This setting will be ignored if field_match_v2 is present.
+        bool field_match = 7 [deprecated = true];
+        // Set to true to highlight fields only if specified in the search query. (Default is false)
+        google.protobuf.BoolValue field_match_v2 = 8;
```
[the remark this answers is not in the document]
waziqi89: It is compatible.  so removing field_match_v2 now

src/main/java/com/yelp/nrtsearch/server/luceneserver/highlights/HighlightFetchTask.java
```diff
@@ -95,4 +107,27 @@ public void processHit(SearchContext searchContext, LeafReaderContext hitLeaf, B
...
+      if (!(field instanceof TextBaseFieldDef)) {
+        throw new IllegalArgumentException(
+            String.format(
+                "Field %s is not a text field and does not support highlights", fieldName));
+      }
+      if (!((TextBaseFieldDef) field).isSearchable()) {
+        throw new IllegalArgumentException(
+            String.format("Field %s is not searchable and cannot support highlights", fieldName));
+      }
+      highlighter.verifyFieldIsSupported(((TextBaseFieldDef) field));
```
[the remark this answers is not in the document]
waziqi89: This won't be changed as exceptions can give more information to the users for the failure reason through the error messages.

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/SearchContext.java
waziqi89: OK. In this case, it shall be non-nullable even if the highlighter task doesn't exists.

docs/highlighting.rst
```diff
@@ -3,71 +3,100 @@ Highlighting
...
 Highlights are used to retrieve the information about matched terms in a hit. Nrtsearch currently supports the Fast Vector Highlighter in Lucene.
 
+Custom highlighters can be provided with HighlighterPlugin interface.
+
 Requirements
 ------------
 
 To be able to highlight a field the following settings must be enabled when registering the field:
 
+general:
```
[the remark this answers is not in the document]
waziqi89: `Requirements` using `-` is a subsection.
`general` using `^` is a subsubsection. general should be under requirements, so I think it's good to keep it as this.

docs/highlighting.rst
```diff
@@ -3,71 +3,100 @@ Highlighting
...
+* The field ``field_match`` (v1) is deprecated as it doesn't support the field override. For any new client, please use ``field_match_v2`` instead. ``field_match`` will only take effect when ``field_match_v2`` is omitted. And ``field_match`` (v1) will be removed eventually.
 
 Example Queries
 ---------------
 
 Below examples use a simple index called "test_index" with two fields doc_id and comment. The index has two documents:
 
-* doc_id: 1, comment: the food here is amazing, service was good
-* doc_id: 2, comment: This is my first time eating at this restaurant. The food here is pretty good, the service could be better. My favorite food was chilly chicken.
+.. table:: Documents in test_index
```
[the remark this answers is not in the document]
waziqi89: I was using sublime during the rst development. Will use the tool mentioned here before pushes

docs/highlighting.rst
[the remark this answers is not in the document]
sarthakn7: I meant it should start with upper-case i.e. 
```
General:
^^^^^^^
```

src/main/java/com/yelp/nrtsearch/server/luceneserver/highlights/HighlightUtils.java
```diff
@@ -128,19 +171,27 @@ private static HighlightSettings createGlobalFieldSettings(
...
+        return DEFAULT_HIGHLIGHTER_NAME;
+      case PLAIN:
+        throw new UnsupportedOperationException("plain-highlighter is not supported yet.");
+      case FAST_VECTOR:
+        return NRTFastVectorHighlighter.HIGHLIGHTER_NAME;
+      case CUSTOM:
+        return settings.getCustomHighlighterName();
+      default:
+        throw new IllegalArgumentException(
+            String.format("Unknown highlighter_type [%s]", settings.getHighlighterType()));
```
sarthakn7, line 194: Nit: change to `"Unknown highlighter_type: %s"` so it's formatted similar to most other errors. You can also use just `+` to append the strings if you want

sarthakn7 (approved):

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/SearchRequestProcessor.java
```diff
@@ -158,12 +157,12 @@ public static SearchContext buildContextForRequest(
...
     if (!highlight.getFieldsList().isEmpty()) {
-      verifyHighlights(indexState, highlight);
       HighlightFetchTask highlightFetchTask =
-          new HighlightFetchTask(indexState, searcherAndTaxonomy, query, highlight);
+          new HighlightFetchTask(
+              indexState, searcherAndTaxonomy, query, HighlighterService.getInstance(), highlight);
       contextBuilder.setHighlightFetchTask(highlightFetchTask);
     }
-
+    contextBuilder.setExtraContext(new HashMap<>());
```
taoyyu, line 165: Is this used already? Do we need to worry about threadsafe for hashmap here now?

waziqi89: It is not used yet. We want to use it in plan highlighter to cache highlighter per field.
It shall be thread safe though

taoyyu (approved):

PR #545, opened 2023-02-06
Added a JUnit rule for testing with S3 and replaced most direct usages of S3Mock

sarthakn7:
Created a JUnit test rule `AmazonS3Provider` to help with tests requiring a mock AmazonS3 client. This will make it easier to write these tests and also reduce the boilerplate code. 
Usage of the rule from the javadoc for it:
```
/**
 * A JUnit {@link org.junit.Rule} for mock S3 tests. It provides a mock S3 client which stores any
 * files in and retrieves from a {@link TemporaryFolder} so that all data added to S3 is deleted
 * after the test.
 *
 * <p>Example of usage:
 *
 * <pre>
 * public static class S3Test {
 *  &#064;Rule
 *  public AmazonS3Provider s3Provider = new AmazonS3Provider("test-bucket");
 *
 *  &#064;Test
 *  public void testUsingS3() throws IOException {
 *      AmazonS3 s3Client = s3Provider.getAmazonS3();
 *      s3Client.putObject("test_bucket", "key", file);
 *      // ...
 *     }
 * }
 * </pre>
 */
```

I've replaced most direct usages of S3Mock with this rule. As a bonus this also helped reduce the usage of a deprecated constructor for AmazonS3.

### DIFF
--- a/src/test/java/com/yelp/nrtsearch/server/backup/ArchiverTest.java
+++ b/src/test/java/com/yelp/nrtsearch/server/backup/ArchiverTest.java
@@ -20,13 +20,11 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
-import com.amazonaws.auth.AnonymousAWSCredentials;
 import com.amazonaws.services.s3.AmazonS3;
-import com.amazonaws.services.s3.AmazonS3Client;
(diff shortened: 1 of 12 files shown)


fragosoluana (changes requested):

fragosoluana (approved):

fragosoluana:
++ for making the test code more succinct

PR #548, opened 2023-02-16
New feature - download plugin from S3 when S3 path is provided instead of plugin name

sarthakn7:
Nrtsearch will now download a plugin from S3 if the plugin name in config is an S3 path. 
- The plugin in S3 must be zip, and the example-plugin already packages it in the correct way. 
- There is a new parameter in the config: savePluginBeforeUnzip. If true it will first download the plugin zip to a local file and then unzip it. If false (default) it will try to unzip as the file is being downloaded. The latter might not work with some kind of zip files which require a seekable stream (https://commons.apache.org/proper/commons-compress/zip.html for more details). The latter is more performant and should ideally work with the gradle packaged zip files. If it does not work with then it should be easy to add the config parameter to save the file first.

For testing:
- Created a helper class called NrtsearchTest which can be extended by test classes. Unlike `ServerTestCase` which creates an instance of LuceneServerImpl directly, NrtsearchTest creates a LuceneServer instance 
- This was required to properly test the plugin download, since the `PluginsService` is created in `LuceneServer` constructor and plugins are loaded in `LuceneServer.start()`. It was possible to directly test the plugin loading in PluginsService but that would not have tested the complete initialization path
- NrtsearchTest uses TestLuceneServer Junit rule to create the test server. The test rule uses a custom Guice module `TestLuceneServerModule` to create a `LuceneServer` instance
- I moved the shutdown hook from `LuceneServer.start()` to the call path for main() method since it's not required in tests and should be called for running nrtsearch servers only


src/test/java/com/yelp/nrtsearch/server/plugins/PluginS3DownloadTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+  @Override
+  protected List<String> getPlugins() {
+    String path = String.format("s3://%s/%s", getS3BucketName(), PLUGIN_S3_KEY);
+    return List.of(path);
+  }
+
+  @Test
+  public void testPluginS3Path() {
+    assertThat(Files.exists(getPluginSearchPath().resolve("example-plugin-0.0.1"))).isTrue();
```
fragosoluana, line 62: nit: could use `assertTrue` instead

src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
```diff
@@ -186,21 +189,9 @@ private void start() throws IOException {
...
-                // Use stderr here since the logger may have been reset by its JVM shutdown hook.
-                logger.error("*** shutting down gRPC server since JVM is shutting down");
-                LuceneServer.this.stop();
-                logger.error("*** server shut down");
-              }
-            });
   }
 
-  private void stop() {
+  public void stop() {
```
fragosoluana, line 194: nit: you could use the @VisibleForTesting to indicate this is widely visible just for testing purpose

fragosoluana (approved):

src/test/java/com/yelp/nrtsearch/server/plugins/PluginS3DownloadTest.java
sarthakn7: Made this change, although it's just different syntax (assertj vs junit)

src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
sarthakn7: I don't see any harm if someone wants to import the server and start LuceneServer within their code, but still added it

fragosoluana (approved):

PR #549, opened 2023-02-21
plain highlighter

waziqi89:
RP-7434 (more motivations are discussed on the ticket.)

- implement plain highlighter.
    - get all fields if multivalue field, highlight each, then sort if scoredOrder
   - cache highlighter per field in the extraContext.
- remove the logics of modification when numberOfMaxFragments == 0 in highlightsettingUtils. It is already implemented in fhv 
- fix a test in fvh.


waziqi89:
should merge changes from

https://github.com/Yelp/nrtsearch/pull/550

waziqi89:
re-open as it is adapted to the change in PR-550

waziqi89:
Won't do for now

PR #551, opened 2023-02-24
Highlighter bug fixing

waziqi89:
Fixes two issues:

1. settings toBuilder missed field `discreteMultivalue`
2. highlight utils causes some failures in highlight single fragments when fragmentNumber is 0 (it is handled in NRTFastVectorHighlighter already)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,7 +24,7 @@ sourceCompatibility = 1.14
 targetCompatibility = 1.14
 
 allprojects {
-    version = '0.20.0'
+    version = '0.20.1'
     group = 'com.yelp.nrtsearch'
 }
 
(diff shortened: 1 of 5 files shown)


waziqi89:
https://github.com/Yelp/nrtsearch/blob/master/src/main/java/com/yelp/nrtsearch/server/luceneserver/highlights/NRTFastVectorHighlighter.java#L88
@sarthakn7 
This is a duplicate setting as it presents in the fvh already; and the test looks more correct

sarthakn7 (commented):
@waziqi89 can we change [this](https://github.com/Yelp/nrtsearch/blob/master/src/main/java/com/yelp/nrtsearch/server/luceneserver/highlights/NRTFastVectorHighlighter.java#L89) comment to say that this is required to support the feature to return the entire text as a single fragment? The description is [here](https://github.com/Yelp/nrtsearch/blob/master/clientlib/src/main/proto/yelp/nrtsearch/search.proto#L816) and I think ES does the same

waziqi89:
> @waziqi89 can we change [this](https://github.com/Yelp/nrtsearch/blob/master/src/main/java/com/yelp/nrtsearch/server/luceneserver/highlights/NRTFastVectorHighlighter.java#L89) comment to say that this is required to support the feature to return the entire text as a single fragment? The description is [here](https://github.com/Yelp/nrtsearch/blob/master/clientlib/src/main/proto/yelp/nrtsearch/search.proto#L816) and I think ES does the same. It's not a hack.

Done; also bump the patch version

sarthakn7 (approved):

PR #553, opened 2023-03-03
RP-7509: Add constant score query support

aprudhomme:
Add support for a constant score query. This query wraps a filter query and sets matching document scores to a value of 1.0. This constant value can be modified by setting the query `boost` parameter.


src/test/java/com/yelp/nrtsearch/server/luceneserver/search/query/ConstantScoreQueryTest.java
```diff
@@ -0,0 +1,172 @@
...
+                    .setIndexName(DEFAULT_TEST_INDEX)
+                    .setTopHits(10)
+                    .addRetrieveFields("doc_id")
+                    .setQuery(
+                        Query.newBuilder()
+                            .setConstantScoreQuery(
+                                ConstantScoreQuery.newBuilder()
+                                    .setFilter(
+                                        Query.newBuilder()
+                                            .setMatchQuery(
```
vim345, line 111: Is it possible to add another match query to see if the final calculated score is the sum of the scores or the multiplication of them?

vim345 (approved):
Just a small suggestion.

src/test/java/com/yelp/nrtsearch/server/luceneserver/search/query/ConstantScoreQueryTest.java
aprudhomme: The score of the match query is not used, as it is replaced by the constant score.

PR #555, opened 2023-03-06
fix a bug from fetching top N fields from fhv

waziqi89:
RP-7515

During smart snippets PoC. we find that we shall use `SINGLE_FRAG_LIST_BUILDER` in case of fetching the whole field.
Otherwise, it returns the top 1 field only.

Now, when set `fragment_size == Integer.MAX_VALUE` (or 0 which is easier for users), we use SINGLE_FRAG_LIST_BUILDER.

### DIFF
--- a/clientlib/src/main/proto/yelp/nrtsearch/search.proto
+++ b/clientlib/src/main/proto/yelp/nrtsearch/search.proto
@@ -819,7 +819,7 @@ message Highlight {
         repeated string pre_tags = 2;
         // Used along with pre_tags to specify how to wrap the highlighted text.
         repeated string post_tags = 3;
-        // Number of characters in highlighted fragment, 100 by default.
+        // Number of characters in highlighted fragment, 100 by default. Set it to be 0/Integer.MAX_VALUE to fetch the entire field.
         google.protobuf.UInt32Value fragment_size = 4;
         // Maximum number of highlight fragments to return, 5 by default. If set to 0 returns entire text as a single fragment ignoring fragment_size.
         google.protobuf.UInt32Value max_number_of_fragments = 5;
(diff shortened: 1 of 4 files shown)


sarthakn7 (commented):
@waziqi89 you had already made the fix to highlight entire text in the field when max num fragments was set to 0 (#551 ) but I found that we were not setting the max num fragments for each field correctly. Due to this it worked as expected if max num fragments was set to 0 in the global setting but did not work if it was set to 0 for field setting. I have sent out the fix for this (#556 ).
Per my understanding [Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/highlighting.html) also has the same behavior, i.e. set max num fragments to 0 to ignore fragment size. I think we should try to keep the same behavior so that migrations to nrtsearch are easier.
If you think it's needed we can have the same behavior when fragment_size is set to 0 as well but we shouldn't do it for Integer.MAX_VALUE since that number can vary for clients.

taoyyu (approved):

waziqi89:
Hi, @sarthakn7.
For this PR, I think it’s about a different setting: the fragment size, not the number of fragment. During the test, when fragment size == Max && frament count > 1, the fbv can only fetch the first field.

sarthakn7:
@waziqi89 As far as I understand setting max num fragments to zero will return the entire text with all highlights in a single fragment. With your changes setting fragment_size to zero/Integer.MAX_VALUE will add the exact same behavior. So why not just set max num fragments to zero? Not sure if I'm missing something here.

waziqi89:
> @waziqi89 As far as I understand setting max num fragments to zero will return the entire text with all highlights in a single fragment. With your changes setting fragment_size to zero/Integer.MAX_VALUE will add the exact same behavior. So why not just set max num fragments to zero? Not sure if I'm missing something here.

It's mainly for the multivalue field. when discreteValues is set true, and number of fragments is 0, it will return all value fields from this multivalue field <-- which is as expected
However, in case discreteValues is set true, and number of fragments is a number > 1, saying 5, and fragment size is 0/MaxInt, we expect it returns 5 value fields. And without this fix, it will only return 1.

PR #557, opened 2023-03-07
resolve the confusion in setting fragment_size = MaxInt

waziqi89:
RP-7515
Integer max value is dependent on platforms.
We shall not encourage clients to use it, because what we mean is 2147483647 in java only.


sarthakn7 (commented):
Should we keep the behavior only for fragment_size = 0? i.e. change the code to not use SingleFragListBuilder for Integer.MAX_VALUE? Related - is it worth or is it needed to have the behavior for Integer.MAX_VALUE?

sarthakn7 (approved):

PR #558, opened 2023-03-08
add explain support to search query

taoyyu:
add explain option to search request for score details.

I only support the search with explain here. The ES explain api is not very necessary here.

### DIFF
--- a/clientlib/src/main/proto/yelp/nrtsearch/search.proto
+++ b/clientlib/src/main/proto/yelp/nrtsearch/search.proto
@@ -414,6 +414,7 @@ message SearchRequest {
     string responseCompression = 23;
     // Specify how to highlight matched text
     Highlight highlight = 24;
+    bool explain = 25;
 }
 
 /* Virtual field used during search */
@@ -562,6 +563,7 @@ message SearchResponse {
         map<string, CompositeFieldValue> sortedFields = 4; // Sorted field name to value
         // Field name to highlighted text fragments
         map<string, Highlights> highlights = 5;
+        string explain = 6;
     }
 
     message SearchState {
--- a/clientlib/src/main/proto/yelp/nrtsearch/search.proto
+++ b/clientlib/src/main/proto/yelp/nrtsearch/search.proto
@@ -414,6 +414,7 @@ message SearchRequest {
     string responseCompression = 23;
     // Specify how to highlight matched text
     Highlight highlight = 24;
+    // If Lucene explanation should be included in the response
     bool explain = 25;
 }
 
@@ -563,6 +564,7 @@ message SearchResponse {
         map<string, CompositeFieldValue> sortedFields = 4; // Sorted field name to value
         // Field name to highlighted text fragments
         map<string, Highlights> highlights = 5;
+        // Lucene explanation of the hit
         string explain = 6;
     }
 
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/SearchHandler.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/SearchHandler.java
@@ -180,8 +180,8 @@ public class SearchHandler implements Handler<SearchRequest, SearchResponse> {
       List<Explanation> explanations = null;
       if (searchRequest.getExplain()) {
         explanations = new ArrayList<>();
-        for (int i = 0; i < hits.scoreDocs.length; i++) {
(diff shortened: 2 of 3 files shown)


clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -562,6 +563,7 @@ message SearchResponse {
         map<string, CompositeFieldValue> sortedFields = 4; // Sorted field name to value
         // Field name to highlighted text fragments
         map<string, Highlights> highlights = 5;
+        string explain = 6;
```
waziqi89, line 566: We might need to add comments, and update *rst

taoyyu: I'll add comments first. Explain should be part of nrtsearch request/response rst files. But these rst files do not exist yet. I'll create a ticket to add these rst files

waziqi89 (approved):

src/main/java/com/yelp/nrtsearch/server/luceneserver/SearchHandler.java
```diff
@@ -176,6 +177,14 @@ public SearchResponse handle(IndexState indexState, SearchRequest searchRequest)
         hits = searcherResult.getTopDocs();
       }
 
+      List<Explanation> explanations = null;
+      if (searchRequest.getExplain()) {
+        explanations = new ArrayList<>();
+        for (int i = 0; i < hits.scoreDocs.length; i++) {
+          explanations.add(s.searcher.explain(searchContext.getQuery(), i));
+        }
```
sarthakn7, line 185: This can be faster if the explanations are created for each document in parallel. The fetchFields method already handles some parallelism at document and field level and this can be added there as well. The downside is that this becomes more complicated as you would need both serial and parallel implementations but the performance improvement especially if large number of documents are retrieved might be worth it.

src/test/java/com/yelp/nrtsearch/server/grpc/QueryTest.java
```diff
@@ -911,6 +922,22 @@ private void testWithBoost(Query originalQuery, SearchResponse searchResponse) {
     }
   }
 
+  private void testWithExplain(Query query, SearchResponse searchResponse) {
```
sarthakn7, line 925: It seems like this method is not called anywhere. It might be better to make this a `@Test` method for explain.

src/test/java/com/yelp/nrtsearch/server/grpc/QueryTest.java
```diff
@@ -911,6 +922,22 @@ private void testWithBoost(Query originalQuery, SearchResponse searchResponse) {
...
+    boolean explain = true;
+    SearchResponse searchResponseExplained =
+        grpcServer.getBlockingStub().search(buildSearchRequestWithExplain(query, explain));
+    for (int i = 0; i < searchResponse.getHitsCount(); i++) {
+      SearchResponse.Hit hit = searchResponse.getHits(i);
+      SearchResponse.Hit explainedHit = searchResponseExplained.getHits(i);
+      SearchResponse.Hit hitWithoutExplain =
+          SearchResponse.Hit.newBuilder(hit).setExplain("").build();
```
sarthakn7, line 933: Nit: `hit.toBuilder()` would be a bit more concise as compared to `SearchResponse.Hit.newBuilder(hit)`. Also if you compare the explanations first you won't need to create a builder to compare fields other than explain.

src/test/java/com/yelp/nrtsearch/server/grpc/QueryTest.java
```diff
@@ -911,6 +922,22 @@ private void testWithBoost(Query originalQuery, SearchResponse searchResponse) {
...
+      SearchResponse.Hit explainedHitWithoutExplain =
+          SearchResponse.Hit.newBuilder(explainedHit).setExplain("").build();
+      assertEquals(hitWithoutExplain, explainedHitWithoutExplain);
+      assertTrue(explainedHit.getExplain().contains(String.valueOf(explainedHit.getScore())));
```
sarthakn7, line 937: Can we assert for the complete string in the explain?

waziqi89:
test failed

taoyyu:
> test failed

tests passed locally. Let me rerun it

waziqi89 (approved):

PR #559, opened 2023-03-08
Add config option for max s3 client retries

aprudhomme:
Makes the maximum number of retries by the server s3 client configurable with the `maxS3ClientRetries` option. When this option is set to a value > 0, a retry policy is configured using the default retry condition and backoff strategy. If the value is <= 0 (default -1), the default client retry policy is used.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/config/LuceneServerConfiguration.java
+++ b/src/main/java/com/yelp/nrtsearch/server/config/LuceneServerConfiguration.java
@@ -45,6 +45,7 @@ public class LuceneServerConfiguration {
   public static final Path DEFAULT_INDEX_DIR =
       Paths.get(DEFAULT_USER_DIR.toString(), "default_index");
   private static final String DEFAULT_BUCKET_NAME = "DEFAULT_ARCHIVE_BUCKET";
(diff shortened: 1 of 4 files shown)


vim345 (approved):

src/main/java/com/yelp/nrtsearch/server/config/LuceneServerConfiguration.java
```diff
@@ -45,6 +45,7 @@ public class LuceneServerConfiguration {
   public static final Path DEFAULT_INDEX_DIR =
       Paths.get(DEFAULT_USER_DIR.toString(), "default_index");
   private static final String DEFAULT_BUCKET_NAME = "DEFAULT_ARCHIVE_BUCKET";
+  static final int DEFAULT_MAX_S3_CLIENT_RETRIES = -1;
```
sarthakn7, line 48: Instead of -1 should we have a default which does retry some sane number of times? I feel this would be good to have by default and also seems safe enough

aprudhomme: The thought was to make this a noop unless set, so it was a very safe change. The sdk default is 3, we could maybe update this after we've done some testing of alternate values.

sarthakn7 (approved):

PR #562, opened 2023-03-21
Enable global bucket access

sarthakn7:
If `enableGlobalBucketAccess` is set to true in the server config enable forceGlobalBucketAccess on AmazonS3 client.
This can be helpful if plugins are in a bucket in a different region from the index, or if plugins are in several different buckets.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/module/S3Module.java
+++ b/src/main/java/com/yelp/nrtsearch/server/module/S3Module.java
@@ -83,15 +83,14 @@ public class S3Module extends AbstractModule {
         clientBuilder.setClientConfiguration(clientConfiguration);
       }
 
-      AmazonS3ClientBuilder amazonS3ClientBuilder = clientBuilder
-          .withCredentials(awsCredentialsProvider)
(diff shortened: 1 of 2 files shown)


src/main/java/com/yelp/nrtsearch/server/module/S3Module.java
```diff
@@ -82,11 +83,14 @@ protected AmazonS3 providesAmazonS3(LuceneServerConfiguration luceneServerConfig
         clientBuilder.setClientConfiguration(clientConfiguration);
       }
 
-      return clientBuilder
-          .withCredentials(awsCredentialsProvider)
-          .withEndpointConfiguration(
-              new AwsClientBuilder.EndpointConfiguration(serviceEndpoint, region))
-          .build();
+      AmazonS3ClientBuilder amazonS3ClientBuilder =
```
aprudhomme, line 86: The `clientBuilder` is already a local reference to the `AmazonS3ClientBuilder`. Do we need to create another one?

aprudhomme (approved):

PR #563, opened 2023-03-21
custom highlighter params

waziqi89:
RP-7642

allow custom highlighter params to extend the flexibility of custom highlighter.

No deepMerge for field overriding.


sarthakn7 (commented):
Can you please regenerate the grpc-gateway as well?

PR #564, opened 2023-03-23
override lucene BaseFragmentsBuilder to fix a bug

waziqi89:
https://github.com/Yelp/nrtsearch/issues/567
RP-7664

Fix a bug in lucene core, where the fragment end is calculated wrongly when using discreteMultiValue.

- overriding the buggy class with a fix in line https://github.com/Yelp/nrtsearch/pull/564/files#diff-df0e127b960bddcdd8d618e76351dc785218a9c8cbebcb6ba88b3d649401095aR157
 - fragEnd ==> fragEnd - 1
- copied classes inherits the buggy class to the extend the new class


src/main/java/org/apache/lucene/search/vectorhighlight/MultiValueBaseFragmentsBuilder.java
```diff
@@ -0,0 +1,179 @@
...
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import org.apache.lucene.document.Field;
+import org.apache.lucene.search.vectorhighlight.FieldFragList.WeightedFragInfo;
+import org.apache.lucene.search.vectorhighlight.FieldFragList.WeightedFragInfo.SubInfo;
+import org.apache.lucene.search.vectorhighlight.FieldPhraseList.WeightedPhraseInfo.Toffs;
+
+public abstract class MultiValueBaseFragmentsBuilder extends BaseFragmentsBuilder {
```
sarthakn7, line 30: Instead of creating a new class `MultiValueBaseFragmentsBuilder` should we override `MultiValueBaseFragmentsBuilder` itself? i.e. Create class with the name `BaseFragmentsBuilder` with the fix. Making similar changes is the reason why we have some classes in the org.apache.lucene package in nrtsearch.

sarthakn7 (commented):
++ for finding this bug. Can you please create an issue in nrtsearch and then reference that issue number here instead of the ticket number?

waziqi89:
reported to Lucene-core https://github.com/Yelp/nrtsearch/issues/567

PR #574, opened 2023-05-05
Add bucket ordering specification for terms collector

aprudhomme:
Add the ability to specify ordering for terms collection sorting. Previously, this defaulted to ordering by the bucket count in descending order.

A `BucketOrder` gRPC message specifies:
- key: What to sort by, can be `_count` for the bucket count or the name of a nested collector
- order: One of ascending (ASC) or descending (DESC)

The only collector that currently supports use for nested sorting is the `Max` collector.

The `BucketOrder` class was added to the clientlib so that is can be used to help in merging the results of a scatter gather.


src/test/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/additional/NestedCollectorOrderTest.java
```diff
@@ -0,0 +1,353 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import org.apache.lucene.index.IndexWriter;
+import org.apache.lucene.index.NoMergePolicy;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+public class NestedCollectorOrderTest extends ServerTestCase {
+  @ClassRule public static final GrpcCleanupRule grpcCleanup = new GrpcCleanupRule();
```
sarthakn7, line 54: This doesn't need to be defined again since the rule is already there in the ServerTestCase class. We've unnecessarily added it to a lot of tests, need to remove them sometime.

src/test/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/additional/NestedCollectorOrderTest.java
```diff
@@ -0,0 +1,353 @@
...
+
+    // write docs in random order with 10 per segment
+    Collections.shuffle(docs);
+    for (int i = 0; i < 10; ++i) {
+      List<AddDocumentRequest> chunk = new ArrayList<>();
+      for (int j = 0; j < 10; ++j) {
+        chunk.add(docs.get((10 * i) + j));
+      }
+      addDocuments(chunk.stream());
+      writer.commit();
```
sarthakn7, line 106: Nit: I think it'll be better to call commit via grpc api and minimize direct usage of IndexWriter

src/test/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/additional/NestedCollectorOrderTest.java
```diff
@@ -0,0 +1,353 @@
...
+    return Collections.singletonList(DEFAULT_TEST_INDEX);
+  }
+
+  protected FieldDefRequest getIndexDef(String name) throws IOException {
+    return getFieldsFromResourceFile("/search/collection/nested_order.json");
+  }
+
+  protected void initIndex(String name) throws Exception {
+    IndexWriter writer = getGlobalState().getIndex(name).getShard(0).writer;
+    // don't want any merges for these tests
```
sarthakn7, line 66: Can we also tell why we don't want merges in this comment?

sarthakn7 (commented):
Can we add some documentation for this as well?

sarthakn7 (commented):
Can we add some documentation for this as well?

src/test/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/additional/NestedCollectorOrderTest.java
aprudhomme: I'm not sure the rule works right as a static field of the base class, but I'll remove this line since it isn't doing anything useful. We might want to verify that the rule works as expected at some point, and modify the interface if needed.

sarthakn7 (approved):

PR #577, opened 2023-05-26
InnerHit

waziqi89:
RP-7863
Create innerHit fetchTask.
This is on top of https://github.com/Yelp/nrtsearch/pull/576


src/main/java/com/yelp/nrtsearch/server/luceneserver/innerhit/InnerHitContext.java
```diff
@@ -0,0 +1,80 @@
...
+import com.yelp.nrtsearch.server.luceneserver.search.SearchContext;
+import com.yelp.nrtsearch.server.luceneserver.search.collectors.DocCollector;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import org.apache.lucene.search.Query;
+import org.apache.lucene.search.join.BitSetProducer;
+import org.apache.lucene.search.join.QueryBitSetProducer;
+
+public class InnerHitContext extends SearchContext {
```
aprudhomme, line 32: Does this really need to extend `SearchContext`? If it needs to retrieve document fields, it can just implement the `FieldFetchContext` interface.

src/main/java/com/yelp/nrtsearch/server/luceneserver/innerhit/InnerHitFetchTask.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.List;
+import java.util.concurrent.atomic.DoubleAdder;
+import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.search.IndexSearcher;
+import org.apache.lucene.search.ScoreDoc;
+import org.apache.lucene.search.TopDocs;
+import org.apache.lucene.search.join.ParentChildrenBlockJoinQuery;
+
+public class InnerHitFetchTask implements FetchTask {
+  private static final double TEN_TO_THE_POWER_SIX = Math.pow(10, 6);
```
aprudhomme, line 39: nit: maybe a more descriptive name, like `NS_PER_MS`

src/main/java/com/yelp/nrtsearch/server/luceneserver/innerhit/InnerHitFetchTask.java
```diff
@@ -0,0 +1,111 @@
...
+      throws IOException {
+    long startTime = System.nanoTime();
+
+    ParentChildrenBlockJoinQuery parentChildrenBlockJoinQuery =
+        new ParentChildrenBlockJoinQuery(
+            innerHitContext.getParentFilter(), innerHitContext.getQuery(), hit.getLuceneDocId());
+
+    IndexSearcher searcher = innerHitContext.getSearcherAndTaxonomy().searcher;
+    TopDocs topDocs =
+        searcher.search(parentChildrenBlockJoinQuery, innerHitContext.getCollector().getManager());
```
aprudhomme, line 65: You will want to use the `search` method that takes a `Collector` rather than a `CollectorManager`, otherwise this will be executed as a parallel search. Having the fetch threadpool use the search threadpool introduces a deadlock scenario. Since this query will only use a single segment, parallelism is not needed.

src/main/java/com/yelp/nrtsearch/server/luceneserver/innerhit/InnerHitFetchTask.java
```diff
@@ -0,0 +1,111 @@
...
+        TotalHits.newBuilder()
+            .setRelation(TotalHits.Relation.valueOf(topDocs.totalHits.relation.name()))
+            .setValue(topDocs.totalHits.value)
+            .build();
+    innerHitResultBuilder.setTotalHits(totalInnerHits);
+    for (int hitIndex = 0; hitIndex < topDocs.scoreDocs.length; hitIndex++) {
+      var hitResponse = innerHitResultBuilder.addHitsBuilder();
+      ScoreDoc innerHit = topDocs.scoreDocs[hitIndex];
+      hitResponse.setLuceneDocId(innerHit.doc);
+      hitResponse.setScore(innerHit.score);
```
aprudhomme, line 85: This scoring information will be different based on if relevance or sorting is used, see https://github.com/Yelp/nrtsearch/blob/master/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/additional/TopHitsCollectorManager.java#L182

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/SearchRequestProcessor.java
```diff
@@ -163,6 +173,60 @@ public static SearchContext buildContextForRequest(
       contextBuilder.setHighlightFetchTask(highlightFetchTask);
     }
     contextBuilder.setExtraContext(new ConcurrentHashMap<>());
+
+    if (searchRequest.getInnerHitsCount() > 0) {
+      List<InnerHitFetchTask> innerHitFetchTasks =
```
aprudhomme, line 178: nit: this block is probably big enough to be a separate method

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/SearchRequestProcessor.java
```diff
@@ -163,6 +173,60 @@ public static SearchContext buildContextForRequest(
...
+          new ArrayList<>(searchRequest.getInnerHitsCount());
+      for (String innerHitName : searchRequest.getInnerHitsMap().keySet()) {
+        InnerHit innerHit = searchRequest.getInnerHitsOrThrow(innerHitName);
+        Query childQuery =
+            extractQuery(indexState, "", innerHit.getInnerQuery(), innerHit.getQueryNestedPath());
```
aprudhomme, line 183: Since this is an arbitrary query, you will want to use `searcher.rewrite()` on it

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/SearchRequestProcessor.java
```diff
@@ -163,6 +173,60 @@ public static SearchContext buildContextForRequest(
...
+        DocCollector innerHitCollector =
+            buildDocCollector(
```
aprudhomme, line 185: Creating a full `DocCollector` may be overkill. Creating a relevance/sort collector could be done similar to https://github.com/Yelp/nrtsearch/blob/master/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/additional/TopHitsCollectorManager.java#L120

src/main/java/com/yelp/nrtsearch/server/luceneserver/innerhit/InnerHitFetchTask.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    // sort hits by lucene doc id
+    List<Hit.Builder> hitBuilders = new ArrayList<>(innerHitResultBuilder.getHitsBuilderList());
+    hitBuilders.sort(Comparator.comparing(Hit.Builder::getLuceneDocId));
```
aprudhomme, line 90: nit: I think `comparingInt` would save some boxing

src/main/java/com/yelp/nrtsearch/server/luceneserver/innerhit/InnerHitContext.java
waziqi89: My only doubt is that do we want to put the rootSearchContext for fetchTask.processHit, or the "parentContext" in case of multi-layer innerHit.

It looks like be out of the current concern now. Change to FieldFetchContext makes sense to me.

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/SearchRequestProcessor.java
waziqi89: [indexSearch.search()](https://github.com/apache/lucene/blob/45110a6a46ab8ab6aad271e1c5b55cb6296fdab9/lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java#L514)
will enforce a rewrite and ParentChildrenBlockJoinQuery will rewrite the childQuery

src/main/java/com/yelp/nrtsearch/server/luceneserver/innerhit/InnerHitContext.java
waziqi89: Forget there were dead loop issue without this design (extends searchContext).

When doFillTask, it will get searchContext and execute all fetchtasks (include innerHit). If we set the searchContext to the rootSearchContext in the innerHit, innerHit's doFillTask will find all rootFetchTasks, and execute all fetchTasks again, which includes itself. So infitnie loop is formed.

I will keep the design, and let getSearchContext to return itself for innerHit. This sounds like a more correct way.

aprudhomme: A way to avoid that situation is move `getHighlightFetchTask()` and `getInnerHitFetchTasks()` into the `FieldFetchContext` interface. That is already the case for the standard `FetchTasks`.

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/SearchRequestProcessor.java
aprudhomme: That might be ok if the rewrite is not too expensive, since it will happen for each hit

clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -426,6 +426,21 @@ message SearchRequest {
...
+    string queryNestedPath = 1; //nested path to search against assuming same index as the parent Query.
+    int32 startHit = 2; // Which hit to start from (for pagination); default: 0
+    int32 topHits = 3; // How many top hits to retrieve; default: 10. It limits the hits returned, starting from index 0. For pagination: set it to startHit + window_size.
+    Query innerQuery = 4;  // InnerHit query to query against the children documents specified by queryNestedPat.h
+    repeated VirtualField virtualFields = 5; // Not supported yet
+    repeated string retrieveFields = 6; // Fields to retrieve; Parent's fields except its id field are unavailable in the innerHit.
+    QuerySortField querySort = 7; //Sort hits by field (default is by relevance).
+    map<string, Collector> collectors = 8;  // Not supported yet
+    Highlight highlight = 9; // Highlight the children documents.
+    bool profile = 10;  // Not supported yet
```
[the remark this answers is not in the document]
waziqi89: Changed most of them. Some new fields are added to the old message where camelCase is used, so to keep consistency in those old messages, I keep the camelCase there.

src/main/java/com/yelp/nrtsearch/server/luceneserver/innerhit/InnerHitContext.java
waziqi89: Thanks for the suggestion. I add the highlight and innerHit into the FetchTasks

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/SearchRequestProcessor.java
waziqi89: get it now. Each hit will rewrite itself if we don't rewrite at the top level.

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/additional/TopHitsCollectorManager.java
```diff
@@ -149,6 +149,9 @@ public String getName() {
   @Override
   public void setSearchContext(SearchContext searchContext) {
     this.searchContext = searchContext;
+    // Backward compatibility: We reuse the highlightFetchTask from the SearchContext
+    this.retrievalContext.fetchTasks.setHighlightFetchTask(
+        searchContext.getFetchTasks().getHighlightFetchTask());
```
aprudhomme, line 154: It may be better to define highlighting as part of the collector definition, but that can be sorted out later

aprudhomme (commented):
It may be good to add some docstrings for public classes and methods

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/SearchRequestProcessor.java
```diff
@@ -142,7 +154,33 @@ public static SearchContext buildContextForRequest(
...
+      innerHitFetchTasks =
+          searchRequest.getInnerHitsMap().keySet().stream()
+              .map(
+                  innerHitName ->
+                      buildInnerHitContext(
+                          indexState,
+                          shardState,
+                          queryFields,
+                          searcherAndTaxonomy,
+                          IndexState.ROOT,
```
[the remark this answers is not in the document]
waziqi89: This is the parent's path to create parentFilter. child's is read from innerHit message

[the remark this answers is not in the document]
aprudhomme: Wouldn't the parent filter be based on `searchRequest.getQueryNestedPath()`, rather than assumed to be `ROOT`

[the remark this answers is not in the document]
waziqi89: Yes, this way sounds more sustainable. But we don't necessary have this field set in root. I can enforce to set it.

waziqi89:
Will start writing the docs.
will bump the version and build the gateway afterwards

aprudhomme (approved):

PR #582, opened 2023-07-12
support adding new child fields

waziqi89:
RP-8578

Support adding new child fields.
Current test covers nested/non-nested object field and text field.

This requires change some mandatory changes in some unmodifiable `Field` and `FieldAndFacetState.Builder`.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/field/IndexableFieldDef.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/field/IndexableFieldDef.java
@@ -23,7 +23,6 @@ import com.yelp.nrtsearch.server.luceneserver.doc.LoadedDocValues;
 import com.yelp.nrtsearch.server.luceneserver.similarity.SimilarityCreator;
 import com.yelp.nrtsearch.server.utils.StructValueTransformer;
 import java.io.IOException;
-import java.util.Collections;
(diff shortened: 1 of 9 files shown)


src/main/java/com/yelp/nrtsearch/server/luceneserver/field/IndexableFieldDef.java
```diff
@@ -110,9 +109,9 @@ protected IndexableFieldDef(String name, Field requestField) {
         }
         childFields.put(childName, (IndexableFieldDef) fieldDef);
       }
-      this.childFields = Collections.unmodifiableMap(childFields);
+      this.childFields = childFields;
     } else {
-      this.childFields = Collections.emptyMap();
+      this.childFields = new HashMap<>();
```
aprudhomme, line 114: This field is intentionally immutable. It is accessed by other indexing threads, and changing it in place can cause issues:
- A ConcurrentModificationException could be thrown in other threads
- The field will be visible to index parsing before it is visible to the ServerCodec, causing the flush to fail
- The field will be visible before the index state is committed, and if there is an error the local state will remain inconsistent

It would be best if the new child field is not visible until the next index state is built. I can't say the best way to do this, without doing some digging. But it may be something like moving the child field maps out of the field classes, and making them accessible through the IndexState.

PR #590, opened 2023-08-02
Support unit in SortType

waziqi89:
RP-8875

We support units in geoRadiusQuery, but not in the SortType, i.e. the returning units are always in m(eter).
To pair with the query, we add this to the SortType.

This is realized by creating a new SortContext object to carry all sort related information. And call the FieldDef specific parser to parse the value (in this case, doing the conversion).


src/main/java/com/yelp/nrtsearch/server/luceneserver/search/SortContext.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.Map;
+import org.apache.lucene.search.FieldDoc;
+import org.apache.lucene.search.Sort;
+import org.apache.lucene.search.SortField;
+
+/**
+ * * Context class to convey all the sorted information, and provides helpers to retrieve the sorted
+ * values.
+ */
+public class SortContext {
```
aprudhomme, line 36: I might suggest restructuring this a bit. The logic could be moved out of the `SortContext` and into `SortParser`, the context would just be a container for processed values.

The `SortParser` could expose something like:
public static Sort parseSort(List<SortType> fields, Map<String, FieldDef> queryFields);
public static SortContext parseSortContext(List<SortType> fields, Map<String, FieldDef> queryFields);
public static Map<String, CompositeFieldValue> getAllSortedValues(FieldDoc fd, SortContext context);

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/properties/HasSortableValueParser.java
```diff
@@ -0,0 +1,36 @@
...
+
+import com.yelp.nrtsearch.server.grpc.SearchResponse.Hit.CompositeFieldValue;
+import com.yelp.nrtsearch.server.grpc.SortType;
+
+/**
+ * * Trait interface for {@link com.yelp.nrtsearch.server.luceneserver.field.FieldDef} types,
+ * denoting that this field's sorted value shall be processed before return. We implicitly assume
+ * any class implement this interface is also {@link Sortable}.
+ */
+public interface HasSortableValueParser {
```
aprudhomme, line 26: I'm a little torn here, on if this should be its own property or part of the existing `Sortable`. If part of the existing property, it could add methods with default implementations that maintain the status quo. I might be persuaded on this.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/properties/HasSortableValueParser.java
```diff
@@ -0,0 +1,36 @@
...
+
+  /**
+   * *
+   *
+   * @param querySortField the sort settings in query
+   * @param value the original sorted value
+   * @return the parsed sorted value in wrapped in {@link CompositeFieldValue}
+   */
+  CompositeFieldValue parseSortedValue(SortType querySortField, Object value);
```
aprudhomme, line 35: I might suggest a usage change here. Rather than this method doing the parsing, have it return an extractor method. Something like:

SortedValueExtractor getSortedValueExtractor(SortType querySortField);

Where `SortedValueExtractor` is a functional interface that takes a `SortField` and `Object` and produces a `CompositeFieldValue`. The default implementation could invoke (or just be) the existing `SortParser` method.

You can hold an array of extractors in the `SortContext` and don't need to worry about holding the `SortType` or `HasSortableValueParser` after parsing. This also lets you catch parsing errors, like invalid units, up front rather than the end of query execution.

src/test/java/com/yelp/nrtsearch/server/grpc/SortFieldTest.java
```diff
@@ -33,6 +36,7 @@ public class SortFieldTest extends ServerTestCase {
   private static final String TEST_INDEX = "test_index";
   private static final int NUM_DOCS = 100;
   private static final int SEGMENT_CHUNK = 10;
+  private static ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
aprudhomme, line 39: tiny nit: this can be final

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/properties/HasSortableValueParser.java
waziqi89: I thought about this extractor idea before. And I don't know how worth it is to create and cache extractors for each sortedField. Feels like in most cases, we just need the default extractor, and it will be repeated over and over again. Unless we want to create extractor per fieldDef to avoid switch-case in current default extractor, I don't think it worth the work of refactoring.

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/properties/HasSortableValueParser.java
waziqi89: I think I shall create 2 methods in sortable:

hasSortedValueExtractor()
extractSortedValue()

and call them in the sortParser

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/properties/HasSortableValueParser.java
aprudhomme: I think my biggest issue with doing things this way is that queries don't fast fail on malformed requests. If an invalid units string is provided, an exception is not thrown until most the work is done.

waziqi89: Changed into sortValueExtractor

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/SortContext.java
waziqi89: fixed

aprudhomme (approved):

PR #592, opened 2023-08-10
make NRTSearch empty boolean query result constant score 1

taoyyu:
This is to match the ES behavior. https://github.com/elastic/elasticsearch/blob/main/server/src/main/java/org/elasticsearch/index/query/BoolQueryBuilder.java#L310C17-L310C17

For ES, empty boolean query will be converted to MatchAll with constant score 1.
In the current NRT, empty boolean query result will only have score 0 since the occur is Filter if there are no positive clauses.
Change it to match ES to unblock migration

### DIFF
--- a/src/test/java/com/yelp/nrtsearch/server/grpc/QueryTest.java
+++ b/src/test/java/com/yelp/nrtsearch/server/grpc/QueryTest.java
@@ -891,7 +891,7 @@ public class QueryTest {
   public void testEmptyBooleanQuery() {
     Query query = Query.newBuilder().setBooleanQuery(BooleanQuery.newBuilder().build()).build();
     SearchResponse response = grpcServer.getBlockingStub().search(buildSearchRequest(query));
-    assertTrue(response.getHitsCount() > 0);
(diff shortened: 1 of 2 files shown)


src/test/java/com/yelp/nrtsearch/server/grpc/QueryTest.java
```diff
@@ -887,6 +887,16 @@ public void testSearchExplain() {
     }
   }
 
+  @Test
+  public void testEmptyBooleanQuery() {
+    Query query = Query.newBuilder().setBooleanQuery(BooleanQuery.newBuilder().build()).build();
+    SearchResponse response = grpcServer.getBlockingStub().search(buildSearchRequest(query));
+    assertTrue(response.getHitsCount() > 0);
```
sarthakn7, line 894: Can we assert the exact count?

taoyyu: Done

sarthakn7 (approved):

PR #599, opened 2023-10-17
Add page fault metrics

aprudhomme:
Add some page fault metrics from the process procfs `stat` handle. There does not appear to be a clean way to get these values from within the java process. The method used is not fully portable and is only best effort. Collection stops if the proc file does not exist, or if there is an error.


vim345 (approved):
It looks like it's identical to what you did with the snapshot branch here: https://github.com/Yelp/nrtsearch/pull/597/files

sarthakn7 (approved):

PR #600, opened 2023-10-18
Update gradle and plugin versions, java 14->17

aprudhomme:
Update build for `1.0.0-SNAPSHOT` to use latest gradle/plugin versions. Also, updated java14 references to java17.


src/main/java/com/yelp/nrtsearch/server/grpc/ReplicationServerClient.java
```diff
@@ -41,7 +41,8 @@ public class ReplicationServerClient implements Closeable {
   public static final int FILE_UPDATE_INTERVAL_MS = 10 * 1000; // 10 seconds
 
   private static final ObjectMapper OBJECT_MAPPER =
-      new ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);;
+      new ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+  ;
```
sarthakn7, line 45: This can be removed

src/main/java/com/yelp/nrtsearch/server/luceneserver/SimpleCopyJob.java
```diff
@@ -101,7 +101,8 @@ public void start() throws IOException {
 
   @Override
   public void runBlocking() throws Exception {
-    while (visit() == false) ;
+    while (visit() == false)
+      ;
```
sarthakn7, line 105: This formatting seems weird. I would prefer having the semi-colon in line with while only. If not maybe we can use empty braces.

src/test/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/CollectorTest.java
```diff
@@ -395,8 +395,12 @@ public void testCustomCollector() throws InvalidProtocolBufferException {
             "7", 93.0,
             "8", 92.0);
     Map<String, Double> collectedHits =
-        collectorsResult.get(COLLECTOR_NAME).getAnyResult().unpack(SearchResponse.class)
-            .getHitsList().stream()
+        collectorsResult
+            .get(COLLECTOR_NAME)
```
sarthakn7, line 399: It looks like the formatter wants `toBuilder()` on the same line but any other function call in a chain of calls on a different line? Kinda confusing, but I guess we can live with it if it can't be changed.

aprudhomme: Hmm, I guess this is just what spotless does now

sarthakn7 (approved):

PR #602, opened 2023-10-19
Deprecate tokenize parameter in field definition

aprudhomme:
These are some proposed changes for the `1.0.0-SNAPSHOT` branch around the handling of tokenization of text based fields.

- The `tokenize` parameter in the field definition is deprecated and no longer used
- `TEXT` type fields will always be tokenized
- `ATOM` type fields will never be tokenized
- `ATOM` norms are alway omitted, this is the default in ES. Ideally, we would want to make this configurable, but the grpc type does not allow for detecting when this is unset (and `TEXT` fields need the opposite default). Since this is not a commonly used option, I think hard coding the value is ok for now.
- `ATOM` fields are no longer indexed for search when the `search` property is false

Additionally
- references to the `tokenize` parameter have been removed
- the parameter has been removed from schema files
- the `search` parameter has been added to `ATOM` fields storing `doc_id` values in schemas, as they are no longer searchable without it. It would be good to switch these to `_ID` fields at some point, but that is beyond the scope of this branch.


aprudhomme:
The example plugin test is an expected failure, since it builds against the latest published branch

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/IdFieldDef.java
```diff
@@ -47,10 +47,10 @@ protected IdFieldDef(String name, Field requestField) {
    * @param requestField field properties to validate
    */
   protected void validateRequest(Field requestField) {
-    if (requestField.getMultiValued() || requestField.getTokenize()) {
+    if (requestField.getMultiValued()) {
```
sarthakn7, line 50: It looks like we're allowing tokenization on the ID field. Let's add that to the PR description as well.

src/test/java/com/yelp/nrtsearch/server/grpc/LuceneServerIdFieldTest.java
```diff
@@ -274,24 +273,11 @@ public void testDuplicateIdFieldInIndexState() throws IOException, InterruptedEx
...
       String message =
           "INVALID_ARGUMENT: error while trying to RegisterFields for index: test_index\n"
-              + "field: doc_id cannot have multivalued fields or tokenization as it's an _ID field";
-      assertEquals(message, e.getMessage());
-      throw e;
-    }
-  }
-
-  @Test(expected = StatusRuntimeException.class)
-  public void testTokenized() throws Exception {
```
sarthakn7, line 288: It would be good to have a test of tokenized ID field working for search

src/main/java/com/yelp/nrtsearch/server/luceneserver/field/IdFieldDef.java
aprudhomme: For `_ID` fields, the tokenization is hard coded to `false` https://github.com/aprudhomme/nrtsearch/blob/deprecate_tokenize_param/src/main/java/com/yelp/nrtsearch/server/luceneserver/field/IdFieldDef.java#L74 so the request values does not matter.

sarthakn7 (approved):

PR #605, opened 2023-10-30
Remove legacy Archiver and backup api

aprudhomme:
Remove the deprecated legacy `Archiver` and the legacy backup system in the `1.0.0-SNAPSHOT` branch.

The `Archiver` interface in general is outdated, and doesn't really fit its usage anymore. I have proposed a `RemoteBackend` interface to replace it, which tries to better abstract away the backend implementation from the other nrtsearch components. This interface is used to replace the legacy archiver for warming query interactions.

The next steps would be to move the incremental archiver interactions (global state, index state, index data) to use this interface as well. This should make it much easier to make future changes related to backend resource structure.


src/test/java/com/yelp/nrtsearch/server/remote/s3/S3BackendTest.java
```diff
@@ -0,0 +1,390 @@
...
+
+    downloadStream =
+        s3Backend.downloadStream(
+            "download_service_4", "download_index_4", IndexResourceType.WARMING_QUERIES);
+    assertEquals("resource_data_2", convertToString(downloadStream));
+  }
+
+  @Test
+  public void testUploadFile() throws IOException {
+    File uploadFile = S3_PROVIDER.getTemporaryFolder().newFile("upload_file");
```
sarthakn7, line 241: It seems weird to get TemporaryFolder from S3Provider and use it. Unless someone reads the code, they might just put things in the s3 directory via the TemporaryFolder without meaning to. Maybe just create TemporaryFolder here for this class and remove the getter?

sarthakn7 (approved):

PR #608, opened 2023-11-04
Add live setting for verbose index metrics

aprudhomme:
Add and index live setting toggle to control collection of more detailed index metrics, defaulting to `false`.

When enabled, the following metrics are collected and published:
- `nrt_search_response_size_bytes` - summary of response protobuf serialized size
- `nrt_search_response_total_hits` - summary of response total hits
- `nrt_search_stage_latency_ms` - summary of latency for each search stage (recall/highlight/fetch/rescorers/etc)
- `nrt_search_timeout_count` - counter of responses with a search recall timeout
- `nrt_search_terminated_early_count` - counter of responses that terminated early for recall count


clientlib/src/main/proto/yelp/nrtsearch/luceneserver.proto
```diff
@@ -1089,6 +1089,8 @@ message IndexLiveSettings {
     google.protobuf.Int32Value defaultTerminateAfter = 13;
     // Merge precopy would be stopped after this time, or 0 for no checks, default: 0
     google.protobuf.UInt64Value maxMergePreCopyDurationSec = 14;
+    // Collect and publish additional index metrics, default: false
```
vim345, line 1092: Should the commend indicate this may introduce performance penalty if set to `true`?

vim345 (approved):
LGTM. I just had a question if enabling this will use more resources on the cluster and if we should let developers know about that.

PR #613, opened 2023-12-20
Add geo polygon query type

aprudhomme:
Add support for geo polygon query.

This query applies to the `LAT_LON` field type with property `search: true`. The query searches for documents with points contained within a specified set of polygons. Polygons may optionally specify holes to exclude areas.

Polygons are specified in clockwise order, except for holes which are counter clockwise. They must not cross themselves. They must not cross the 180th meridian. See https://lucene.apache.org/core/8_4_0/core/org/apache/lucene/geo/Polygon.html

Polygons must be closed shapes. The first point is automatically duplicated as the last point to accomplish this.


sarthakn7 (approved):

docs/queries/geo_polygon.rst
```diff
@@ -0,0 +1,15 @@
+Geo-Polygon Query
+==========================
+
+A query that matches documents with geo point within a defined set of polygons.
```
sarthakn7, line 4: @aprudhomme should we add the restrictions mentioned in the notes in https://lucene.apache.org/core/8_4_0/core/org/apache/lucene/geo/Polygon.html ?

aprudhomme: The important ones are in the `Polygon` comments, but I can add them here too.

I tried to do some digging, and it looks like some of the requirements have to do with conforming to the GeoJson spec. https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.6 Oddly, the winding order in the spec looks like the opposite of the class notes. The spec also says it is ok to ignore the wrong winding order.

I also see that the lucene class detects the winding order in the constructor, so it really might not be vital to functionality. I think I will try writing some tests that use the opposite order, and relax the requirements if they work.

aprudhomme:
I added a reference to the GeoJson standard. Winding order does not seem to matter, so I removed the explicit references from the documentation.

I also added detection for open vs closed polygons, instead of assuming they are all open.

sarthakn7 (approved):

PR #614, opened 2023-12-27
nrt_utils support defaultCredentialsProvider

waziqi89:
RP-9895

Support default credential providers for nrt_utils.
This allows us to use instance profile for AmazonS3.

manual test from dev: https://fluffy.yelpcorp.com/i/mfBgzt5W6T7s6KTNV1n38SpSDRg0gkcm.html

edge case:
When a user want to use default cred-file-path with a non-default profile name:
1. manually put `-c ~/.aws/credentials -p profile-name`
2. provide profile name as an env: AWS_PROFILE=profile-name

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/tools/nrt_utils/incremental/DeleteIncrementalSnapshotsCommand.java
+++ b/src/main/java/com/yelp/nrtsearch/tools/nrt_utils/incremental/DeleteIncrementalSnapshotsCommand.java
@@ -69,12 +69,13 @@ public class DeleteIncrementalSnapshotsCommand implements Callable<Integer> {
 
   @CommandLine.Option(
       names = {"-c", "--credsFile"},
-      description = "File holding AWS credentials, uses default locations if not set")
+      description =
+          "File holding AWS credentials, uses default locations if not set; Will use DefaultCredentialProvider if this is unset.")
   private String credsFile;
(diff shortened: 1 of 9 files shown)


vim345 (approved):

waziqi89:
Hi @aprudhomme! Just to confirm: how do we apply the change the the tron jobs?
merge this PR, and create a new base image? And then the tron job will automatically pick up the latest utils?

src/main/java/com/yelp/nrtsearch/tools/nrt_utils/incremental/DeleteIncrementalSnapshotsCommand.java
```diff
@@ -69,12 +69,13 @@ public class DeleteIncrementalSnapshotsCommand implements Callable<Integer> {
 
   @CommandLine.Option(
       names = {"-c", "--credsFile"},
-      description = "File holding AWS credentials, uses default locations if not set")
+      description =
+          "File holding AWS credentials, uses default locations if not set; Will use DefaultCredentialProvider if this is unset.")
```
fragosoluana, line 73: Should the description contain only `uses default credential provider chain if this is unset` instead of also including `uses default locations if not set`?

fragosoluana (approved):

PR #619, opened 2024-01-25
Avoid calling query.toString()

sarthakn7:
If a query is large and complex, `query.toString()` can be expensive. We had earlier changed it so that it's only executed once per query (#524) but it's better to not call it at all.
Replacing the toString call with instanceof here. Since `DrillSidewaysQuery` is package-private, we need to do the instaceof check in a class in `org.apache.lucene.facet` package.


umeshdangat (approved):
lgtm

PR #621, opened 2024-02-03
Add search diagnostics to deadline exceeded exceptions

swekannan:
Add search diagnostics for deadline exceeded exceptions for easier debugging of latencies.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/grpc/DeadlineUtils.java
+++ b/src/main/java/com/yelp/nrtsearch/server/grpc/DeadlineUtils.java
@@ -15,6 +15,7 @@
  */
 package com.yelp.nrtsearch.server.grpc;
 
+import com.yelp.nrtsearch.server.grpc.SearchResponse.Diagnostics;
 import com.yelp.nrtsearch.server.monitoring.DeadlineMetrics;
 import io.grpc.Context;
 import io.grpc.Deadline;
@@ -50,6 +51,20 @@ public class DeadlineUtils {
         DeadlineMetrics.nrtDeadlineCancelCount.labels(operation).inc();
         throw Status.CANCELLED
             .withDescription("Request deadline exceeded: " + message)
+            .asRuntimeException();
+      }
+    }
+  }
+
+  public static void checkDeadline(String message, Diagnostics diagnostics, String operation) {
+    if (cancellationEnabled) {
+      Deadline deadline = Context.current().getDeadline();
+      if (deadline != null && deadline.isExpired()) {
+        DeadlineMetrics.nrtDeadlineCancelCount.labels(operation).inc();
+        System.out.println(diagnostics);
+        throw Status.CANCELLED
+            .withDescription(
+                "Request deadline exceeded: " + message + ", Search Diagnostics: " + diagnostics)
             .asRuntimeException();
       }
     }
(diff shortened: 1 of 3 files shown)


sarthakn7 (commented):
Can you please check the test failure?

src/main/java/com/yelp/nrtsearch/server/luceneserver/SearchHandler.java
```diff
@@ -191,7 +191,8 @@ public SearchResponse handle(IndexState indexState, SearchRequest searchRequest)
 
       diagnostics.setFirstPassSearchTimeMs(((System.nanoTime() - searchStartTime) / 1000000.0));
 
-      DeadlineUtils.checkDeadline("SearchHandler: post recall", "SEARCH");
+      DeadlineUtils.checkDeadline(
```
sarthakn7, line 194: Maybe add a diagnostics parameter to `checkDeadline` function, and append it to the message only when the error is being thrown? You can also consider keeping the existing function and adding a new checkDeadline with additional parameter in the signature.

swekannan: The method is getting used in many places in the code, so I will create a new method with the additional param.

src/main/java/com/yelp/nrtsearch/server/luceneserver/SearchHandler.java
```diff
@@ -191,7 +191,7 @@ public SearchResponse handle(IndexState indexState, SearchRequest searchRequest)
 
       diagnostics.setFirstPassSearchTimeMs(((System.nanoTime() - searchStartTime) / 1000000.0));
 
-      DeadlineUtils.checkDeadline("SearchHandler: post recall", "SEARCH");
+      DeadlineUtils.checkDeadline("SearchHandler: post recall", diagnostics.build(), "SEARCH");
```
swekannan, line 194: @sarthakn7 I am a bit unsure about this part. Is it okay to log the build diagnostics or should I do some string conversion before logging? What would be the preferred way?

src/main/java/com/yelp/nrtsearch/server/grpc/DeadlineUtils.java
```diff
@@ -54,4 +55,18 @@ public static void checkDeadline(String message, String operation) {
       }
     }
   }
+
+  public static void checkDeadline(String message, Diagnostics diagnostics, String operation) {
+    if (cancellationEnabled) {
+      Deadline deadline = Context.current().getDeadline();
+      if (deadline != null && deadline.isExpired()) {
+        DeadlineMetrics.nrtDeadlineCancelCount.labels(operation).inc();
+        System.out.println(diagnostics);
```
sarthakn7, line 64: Left over from debug

src/main/java/com/yelp/nrtsearch/server/luceneserver/SearchHandler.java
sarthakn7: You can convert it to json before printing if you want to, but it's toString is also fine. You can also just log the Builder object rather than calling build.

swekannan: Ok. I used toString() in this case

sarthakn7 (approved):

PR #629, opened 2024-03-08
Close old reference manager when using filtering searcher manager

aprudhomme:
Using the `FilteringSegmentInfosSearcherManager` requires replacing the reference to the default `IndexSearcher` `ReferenceManager`. However, the previous reference manager was not being closed, which leaves a reference to the initial index segment readers.

This branch fixes the issue by closing the old reference manager after it is replaced.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/NRTReplicaNode.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/NRTReplicaNode.java
@@ -39,6 +39,8 @@ import org.apache.lucene.replicator.nrt.FilteringSegmentInfosSearcherManager;
 import org.apache.lucene.replicator.nrt.NodeCommunicationException;
 import org.apache.lucene.replicator.nrt.ReplicaDeleterManager;
 import org.apache.lucene.replicator.nrt.ReplicaNode;
+import org.apache.lucene.search.IndexSearcher;
+import org.apache.lucene.search.ReferenceManager;
 import org.apache.lucene.search.SearcherFactory;
 import org.apache.lucene.store.Directory;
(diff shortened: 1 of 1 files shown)


src/main/java/com/yelp/nrtsearch/server/luceneserver/NRTReplicaNode.java
```diff
@@ -121,7 +123,9 @@ public synchronized void start(long primaryGen) throws IOException {
       // Swap in a SearcherManager that filters incompatible segment readers during refresh.
       // Updating the reference is not thread safe, but since this happens under the object lock
       // and before the shard has stared, nothing should access the manager before the swap.
+      ReferenceManager<IndexSearcher> oldMgr = mgr;
       mgr = new FilteringSegmentInfosSearcherManager(getDirectory(), this, mgr, searcherFactory);
```
sarthakn7, line 127: It appears as if `FilteringSegmentInfosSearcherManager` is using the old reference manager because it's passed in the constructor, but it only gets the current SegmentInfos: `((SegmentInfosSearcherManager) mgr).getCurrentInfos()`
Can we change the constructor to accept SegmentInfos and then pass it from here so that it's clear that the old reference manager is not accessed any more?

aprudhomme: Unfortunately, `SegmentInfosSearcherManager` is package private and cannot be used from `NRTReplicaNode` unless we relocate the class.

sarthakn7: Makes sense 👍

sarthakn7 (approved):

PR #633, opened 2024-03-15
Add option to specify max threads based on num cpus

aprudhomme:
Add an alternative method for specifying the maximum number of threads for a thread pool. Allows you to compute number of threads based on the number of cpus, with the formula: `(numCPUs * multiplier) + offset` bound by a `min` and `max`.

The max threads are still specified by the `threadPoolConfiguration.max*Threads` config keys. If the config value is a number, the number is used directly as max threads (status quo). If the config is a dict, field values are used to compute the max threads.

Config fields:
- min - minimum value for max threads, default: 1
- max - maximum value for max threads, default: INT_MAX
- multiplier - max threads formula multiplier, default: 1.0
- offset - max threads formula offset, default: 0

Example:
```
threadPoolConfiguration:
  maxSearchingThreads:
    min: 5
    multiplier: 1.5
```

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/config/ThreadPoolConfiguration.java
+++ b/src/main/java/com/yelp/nrtsearch/server/config/ThreadPoolConfiguration.java
@@ -16,7 +16,7 @@
 package com.yelp.nrtsearch.server.config;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
-import com.fasterxml.jackson.databind.ObjectMapper;
+import com.yelp.nrtsearch.server.utils.JsonUtils;
 import java.util.Map;
 
 /** Configuration for various ThreadPool Settings used in nrtsearch */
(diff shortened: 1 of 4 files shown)


src/main/java/com/yelp/nrtsearch/server/config/ThreadPoolConfiguration.java
```diff
@@ -39,6 +43,8 @@ public class ThreadPoolConfiguration {
   private static final int DEFAULT_MAX_GRPC_REPLICATIONSERVER_BUFFERED_ITEMS =
       DEFAULT_MAX_INDEXING_BUFFERED_ITEMS;
 
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
sarthakn7, line 46: Can we use this ObjectMapper? https://github.com/Yelp/nrtsearch/blob/master/src/main/java/com/yelp/nrtsearch/server/utils/JsonUtils.java

sarthakn7 (approved):

PR #634, opened 2024-03-19
fix bug: missinglat not working for virtual field

xinyiyelp:


aprudhomme (commented):
It would be good to add tests to verify this fixes the issue

xinyiyelp:
The script provided by @aprudhomme fixed the issue

```
"script": {
  "lang": "painless",
  "source": "List l = doc[\"score\"]; return l.isEmpty() ? Double.NEGATIVE_INFINITY : l.get(0);"
}
```

PR #646, opened 2024-04-09
Add min collector to Nrtsearch

swethakann:
Collect a minimum double value across all aggregated documents. This value is produced by execution of a score script per document.


clientlib/src/main/java/com/yelp/nrtsearch/server/collectors/BucketOrder.java
```diff
@@ -122,7 +122,8 @@ public static BucketOrder createBucketOrder(
       throw new IllegalArgumentException("Nested collector not found: " + grpcBucketOrder.getKey());
     }
     // currently, max is the only aggregation that can be used for ordering
```
aprudhomme, line 124: Update comment

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/additional/ScriptValueProvider.java
```diff
@@ -0,0 +1,87 @@
...
+import com.yelp.nrtsearch.server.utils.ScriptParamsUtils;
+import java.io.IOException;
+import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.search.DoubleValues;
+import org.apache.lucene.search.DoubleValuesSource;
+import org.apache.lucene.search.Scorable;
+
+/** Value provider that uses a {@link ScoreScript}. */
+class ScriptValueProvider implements ValueProvider {
+  static final double UNSET_VALUE = -Double.MAX_VALUE; // default value
```
aprudhomme, line 32: This value with need to be configurable, now that this is shared between min and max

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/additional/TermsCollectorManager.java
```diff
@@ -198,7 +198,7 @@ public static TermsCollectorManager buildManager(
   protected TermsCollectorManager(
       String name,
       int size,
-      Map<String, Supplier<AdditionalCollectorManager<? extends Collector, CollectorResult>>>
+      Map<String, Supplier<AdditionalCollectorManager<?, CollectorResult>>>
```
swethakann, line 201: Hi @aprudhomme I am getting this error if I don't do this change after making UNSET_VALUE configurable. Seems a bit unrelated coz I put a debugger point and it does not seem to be in the flow. 

`/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/additional/TermsCollectorManager.java:209: error: incompatible types: Map<String,Supplier<AdditionalCollectorManager<? extends Collector,CollectorResult>>> cannot be converted to Map<String,Supplier<AdditionalCollectorManager<?,CollectorResult>>>
: new NestedCollectorManagers(nestedCollectorSuppliers);`

aprudhomme (approved):

PR #650, opened 2024-04-17
Add AUTO fuzziness option for match queries

swethakann:
**Description**
Nrtsearch currently supports inexact fuzzy matching, using the fuzziness parameter in queries. Fuzziness is interpreted as a Levenshtein Edit Distance — the number of one character changes that need to be made to one string to make it the same as another string.

**Requirements**
Support AUTO option for fuzzy params to determine edit distance based on the term length.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/MatchQueryBuilder.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/MatchQueryBuilder.java
@@ -24,14 +24,16 @@ import org.apache.lucene.util.QueryBuilder;
 
 public class MatchQueryBuilder extends QueryBuilder {
 
-  int maxEdits;
+  private final int DEFAULT_LOW = 3;
+  private final int DEFAULT_HIGH = 6;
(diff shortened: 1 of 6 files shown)


src/main/java/com/yelp/nrtsearch/server/luceneserver/MatchQueryBuilder.java
```diff
@@ -24,25 +24,62 @@
...
   protected Query newTermQuery(Term term) {
+    FuzzyParams.FuzzinessCase fuzziness = fuzzyParams.getFuzzinessCase();
+    int maxEdits;
+    switch (fuzziness) {
+      case AUTO -> maxEdits = computeMaxEditsFromTermLength(term);
+      case MAXEDITS -> maxEdits = fuzzyParams.getMaxEdits();
+      default -> {
+        return super.newTermQuery(term); // If fuzziness is not set
+      }
+    }
```
aprudhomme, line 52: Can the max edits computation be moved into a standalone static utility method

PR #651, opened 2024-04-22
Use direct executor for DrillSideways tasks

aprudhomme:
`DrillSideways` takes an `ExecutorService`, which it used to enqueue drill down sub tasks during search. We are currently passing the search thread pool.

Since both `DrillSideways` and parallel search use the same thread pool, there is an execution dependency that can lead to deadlock. This can happen when a burst of requests fill the search thread pool will just `DrillSideways` task, preventing spawned search tasks from running.

This branch uses a direct executor to run the `DrillSideways` tasks in the calling thread (grpc thread). Since we never add any additional drill downs to the query, there will be only one `DrillSideways` task.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/SearchHandler.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/SearchHandler.java
@@ -16,6 +16,7 @@
 package com.yelp.nrtsearch.server.luceneserver;
 
 import com.google.common.collect.Lists;
+import com.google.common.util.concurrent.MoreExecutors;
 import com.google.protobuf.Struct;
 import com.google.protobuf.util.JsonFormat;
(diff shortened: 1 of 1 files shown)


src/main/java/com/yelp/nrtsearch/server/luceneserver/SearchHandler.java
```diff
@@ -148,7 +147,7 @@ public SearchResponse handle(IndexState indexState, SearchRequest searchRequest)
                 shardState,
                 searchContext.getQueryFields(),
                 grpcFacetResults,
-                threadPoolExecutor,
+                DIRECT_EXECUTOR,
```
sarthakn7, line 150: Can we add some comments here so that it's clear why we are using the direct executor?

sarthakn7 (approved):

PR #655, opened 2024-05-17
Add ProportionalCopyThread

aprudhomme:
The status quo copy thread processes chunks of data from queued copy jobs, starting with those of high priority (nrt point data). It seems that under high indexing load, this slows down the processing of merge pre copy data.

I refactored the current copy thread, and added an implementation that gives priority to low priority tasks a given percentage of time. This can let merges make progress more consistently, and may help smooth out the merge rate under indexing load.

The default value is 0%, which uses the status quo single job queue implementation.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/nrt/ProportionalCopyThread.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/nrt/ProportionalCopyThread.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2024 Yelp Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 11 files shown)


src/main/java/com/yelp/nrtsearch/server/luceneserver/nrt/ProportionalCopyThread.java
```diff
@@ -0,0 +1,74 @@
...
+
+  @Override
+  boolean hasJob() {
+    return !highPriorityQueue.isEmpty() || !lowPriorityQueue.isEmpty();
+  }
+
+  @Override
+  CopyJob getJob() {
+    CopyJob job;
+    if (counter++ % 100 < lowPriorityCopyPercentage) {
```
ravirajYelp, line 53: Just curious and wanted to check couple of things:

1.  when counter starts from 0 (say 0 to 10) it would pick from low priority queue even if there are tasks in higher priority queue.  But I assume its okay given large number of tasks.
2.  Am I right that when counter hits Long.MAX_VALUE `counter++ % 100` will be a negative value for a long time and would satisfy the first condition.

aprudhomme: I'm not sure 1 is a big deal. Initially it will process whatever job arrives first. It is unknown where the counter will be when one of each priority job is available. Another option would be to use random numbers, but I'm not sure it makes a big difference, and the batching might actually be better for acked transfers.

For 2, I don't think we need to worry about an overflow. The maximum long value is 9,223,372,036,854,775,807. Even if we incremented 1M a second, it would take 100K+ years to overflow.

sarthakn7 (approved):

PR #660, opened 2024-06-07
Collect more accurate totalHits in the collector

xinyiyelp:
In this PR, we implement the feature to continue to collect doc but skip scoring to not introduce any latency. The approach we are taking here is to overwrite the scorer to constant scorer. More details in the ticket:

Ticket: https://jira.yelpcorp.com/browse/RP-10853

**Definition of Done:**
- [x] Make the simple query work after we reset the scorer
Screenshot 2024-07-02 at 4 17 57 PM

- [x] Find production query and confirm resetting scorer works
I used this [query](https://fluffy.yelpcorp.com/i/H4CJ8kDcjh2LTfmdm3P2SRS2V88qG3gB.html) to run against the NRTSearch, if we remove all restrictions, it's supposed to hit 2M+ records, and the resetting of the scorer actually works
Screenshot 2024-07-03 at 12 22 27 PM

- [x] Write unite tests to cover the changes in the PR
- [x] Manually tested the following cases:
- [DrillSideways](https://sourcegraph.yelpcorp.com/Yelp/nrtsearch@9ffeb9d59d74dac348f5a7f3d20709f0b7f44f5a/-/blob/src/main/java/com/yelp/nrtsearch/server/luceneserver/SearchHandler.java?L163)
- [normal searcher search](https://sourcegraph.yelpcorp.com/Yelp/nrtsearch@9ffeb9d59d74dac348f5a7f3d20709f0b7f44f5a/-/blob/src/main/java/com/yelp/nrtsearch/server/luceneserver/SearchHandler.java?L183)

- [x] fixed this issues:
We have two collectors:
[relevance collector](https://sourcegraph.yelpcorp.com/Yelp/nrtsearch@9ffeb9d59d74dac348f5a7f3d20709f0b7f44f5a/-/blob/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/RelevanceCollector.java?L50) and [sort field collector](https://sourcegraph.yelpcorp.com/Yelp/nrtsearch@9ffeb9d59d74dac348f5a7f3d20709f0b7f44f5a/-/blob/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/SortFieldCollector.java?L54) that set the totalHitsThreshold to Integer.MAX_VALUE . In this case, we respect either totalHitsThreshold or terminateAfter, whichever is bigger. I used this [query](https://fluffy.yelpcorp.com/i/k97PP7NrSwtKqJVv3j29w4H6mGF9zkgn.html) to test this.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/MyIndexSearcher.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/MyIndexSearcher.java
@@ -25,6 +25,8 @@ import java.util.List;
 import java.util.Objects;
 import java.util.PriorityQueue;
 import java.util.concurrent.Executor;
+
+import com.yelp.nrtsearch.server.luceneserver.search.TerminateAfterWrapper;
 import org.apache.lucene.index.IndexReader;
 import org.apache.lucene.index.LeafReaderContext;
 import org.apache.lucene.search.BulkScorer;
@@ -33,6 +35,8 @@ import org.apache.lucene.search.Collector;
 import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.search.IndexSearcher;
 import org.apache.lucene.search.LeafCollector;
+import org.apache.lucene.search.MatchAllDocsQuery;
+import org.apache.lucene.search.ScoreMode;
 import org.apache.lucene.search.Scorer;
 import org.apache.lucene.search.Weight;
 import org.apache.lucene.search.suggest.document.CompletionQuery;
@@ -250,6 +254,11 @@ public class MyIndexSearcher extends IndexSearcher {
       final LeafCollector leafCollector;
       try {
         leafCollector = collector.getLeafCollector(ctx);
+        if (collector instanceof TerminateAfterWrapper.TerminateAfterCollectorWrapper && ((TerminateAfterWrapper.TerminateAfterCollectorWrapper) collector).getTerminatedEarly()) {
+          // overwrite the scorer as constant scorer if the search was terminated early
+          Weight dummyWeight = (new MatchAllDocsQuery()).createWeight(this, ScoreMode.COMPLETE_NO_SCORES, 1);
+          leafCollector.setScorer(dummyWeight.scorer(ctx));
+        }
       } catch (CollectionTerminatedException e) {
         // there is no doc of interest in this reader context
         // continue with the following leaf
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/MyIndexSearcher.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/MyIndexSearcher.java
@@ -25,8 +25,6 @@ import java.util.List;
 import java.util.Objects;
 import java.util.PriorityQueue;
 import java.util.concurrent.Executor;
-
-import com.yelp.nrtsearch.server.luceneserver.search.TerminateAfterWrapper;
 import org.apache.lucene.index.IndexReader;
 import org.apache.lucene.index.LeafReaderContext;
 import org.apache.lucene.search.BulkScorer;
@@ -35,8 +33,6 @@ import org.apache.lucene.search.Collector;
 import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.search.IndexSearcher;
 import org.apache.lucene.search.LeafCollector;
-import org.apache.lucene.search.MatchAllDocsQuery;
-import org.apache.lucene.search.ScoreMode;
 import org.apache.lucene.search.Scorer;
 import org.apache.lucene.search.Weight;
 import org.apache.lucene.search.suggest.document.CompletionQuery;
@@ -254,11 +250,6 @@ public class MyIndexSearcher extends IndexSearcher {
       final LeafCollector leafCollector;
       try {
         leafCollector = collector.getLeafCollector(ctx);
-        if (collector instanceof TerminateAfterWrapper.TerminateAfterCollectorWrapper && ((TerminateAfterWrapper.TerminateAfterCollectorWrapper) collector).getTerminatedEarly()) {
-          // overwrite the scorer as constant scorer if the search was terminated early
-          Weight dummyWeight = (new MatchAllDocsQuery()).createWeight(this, ScoreMode.COMPLETE_NO_SCORES, 1);
-          leafCollector.setScorer(dummyWeight.scorer(ctx));
-        }
       } catch (CollectionTerminatedException e) {
         // there is no doc of interest in this reader context
         // continue with the following leaf
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java
@@ -23,9 +23,15 @@ import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.lucene.index.LeafReaderContext;
 import org.apache.lucene.search.Collector;
 import org.apache.lucene.search.CollectorManager;
+import org.apache.lucene.search.ConstantScoreScorer;
+import org.apache.lucene.search.ConstantScoreWeight;
+import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.search.LeafCollector;
+import org.apache.lucene.search.MatchAllDocsQuery;
 import org.apache.lucene.search.Scorable;
 import org.apache.lucene.search.ScoreMode;
+import org.apache.lucene.search.Scorer;
+import org.apache.lucene.search.Weight;
 
 /**
  * Collector manager wrapper that terminates collection after a given number of documents. This doc
@@ -97,6 +103,7 @@ public class TerminateAfterWrapper<C extends Collector>
 
     private final C collector;
     private boolean terminatedEarly = false;
+    private LeafReaderContext currentContext;
 
     public TerminateAfterCollectorWrapper(C collector) {
       this.collector = collector;
@@ -104,16 +111,13 @@ public class TerminateAfterWrapper<C extends Collector>
 
     @Override
     public LeafCollector getLeafCollector(LeafReaderContext context) throws IOException {
+      currentContext = context;
       return new TerminateAfterLeafCollectorWrapper(collector.getLeafCollector(context));
     }
 
     @Override
     public ScoreMode scoreMode() {
       return collector.scoreMode();
-    }
-
-    public boolean getTerminatedEarly() {
-      return terminatedEarly;
     }
 
     /**
@@ -130,7 +134,24 @@ public class TerminateAfterWrapper<C extends Collector>
 
       @Override
       public void setScorer(Scorable scorer) throws IOException {
-        leafCollector.setScorer(scorer);
+        if (terminatedEarly) {
+          // setup constant scorer to avoid spending time on scoring
+          float score = 1;
+          Weight dummyWeight = new ConstantScoreWeight(new MatchAllDocsQuery(), score) {
+            @Override
+            public boolean isCacheable(LeafReaderContext leafReaderContext) {
+              return true;
+            }
+
+            @Override
+            public Scorer scorer(LeafReaderContext context) throws IOException {
+              return new ConstantScoreScorer(this, this.score(), ScoreMode.COMPLETE_NO_SCORES, DocIdSetIterator.all(context.reader().maxDoc()));
+            }
+          };
+          leafCollector.setScorer(dummyWeight.scorer(currentContext));
+        } else {
+          leafCollector.setScorer(scorer);
+        }
       }
 
       @Override
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java
@@ -21,7 +21,6 @@ import java.util.Collection;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.lucene.index.LeafReaderContext;
-import org.apache.lucene.search.CollectionTerminatedException;
 import org.apache.lucene.search.Collector;
 import org.apache.lucene.search.CollectorManager;
 import org.apache.lucene.search.LeafCollector;
@@ -105,15 +104,16 @@ public class TerminateAfterWrapper<C extends Collector>
 
     @Override
     public LeafCollector getLeafCollector(LeafReaderContext context) throws IOException {
-      if (terminatedEarly) {
-        throw new CollectionTerminatedException();
-      }
       return new TerminateAfterLeafCollectorWrapper(collector.getLeafCollector(context));
     }
 
     @Override
     public ScoreMode scoreMode() {
       return collector.scoreMode();
+    }
+
+    public boolean getTerminatedEarly() {
+      return terminatedEarly;
     }
 
     /**
@@ -137,7 +137,6 @@ public class TerminateAfterWrapper<C extends Collector>
       public void collect(int doc) throws IOException {
         if (collectedDocCount.incrementAndGet() > terminateAfter) {
           terminatedEarly = true;
-          throw new CollectionTerminatedException();
         }
         leafCollector.collect(doc);
       }
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java
@@ -132,7 +132,7 @@ public class TerminateAfterWrapper<C extends Collector>
         }
 
         @Override
-        public float score() throws IOException{
+        public float score() throws IOException {
           if (!terminatedEarly) {
             return in.score();
           } else {
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java
@@ -23,15 +23,9 @@ import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.lucene.index.LeafReaderContext;
 import org.apache.lucene.search.Collector;
 import org.apache.lucene.search.CollectorManager;
-import org.apache.lucene.search.ConstantScoreScorer;
-import org.apache.lucene.search.ConstantScoreWeight;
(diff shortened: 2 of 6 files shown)


src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java
```diff
@@ -135,11 +134,18 @@ public void setScorer(Scorable scorer) throws IOException {
...
-          throw new CollectionTerminatedException();
         }
-        leafCollector.collect(doc);
+
+        if (!terminatedEarly) {
+          // If we haven't terminated early, or if the collector is a TopHitsCollector, collect the doc
+          leafCollector.collect(doc);
+        } else if (leafCollector instanceof TotalHitCountCollector) {
+          // If we haven't terminated early, or if the collector is a TopHitsCollector, collect the doc
+          leafCollector.collect(doc);
```
xinyiyelp, line 147: This is how lucene collects the totalHits:
https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/TotalHitCountCollector.java#L60

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java
```diff
@@ -137,9 +130,15 @@ public void setScorer(Scorable scorer) throws IOException {
...
-          throw new CollectionTerminatedException();
         }
-        leafCollector.collect(doc);
+
+        if (!terminatedEarly) {
+          // If we haven't terminated early, or if the collector is a TopHitsCollector, collect the doc
+          leafCollector.collect(doc);
+        } else if (leafCollector instanceof SearchCollectorManager.SearchCollector.SearchLeafCollector) {
+          // If we haven't terminated early, or if the collector is a TopHitsCollector, collect the doc
+          leafCollector.collect(doc);
```
taoyyu, line 140: SearchLeafCollector is just a simple wrapper of the lucene collectors. In our case, this else if condition is pretty much else if (True).

So with this pr, we are scoring all candidates. The terminatedEarly is ignored.

taoyyu: Our requirements for this feature would be when totalhitsthresshold > terminateAfter
The collector keeps counting the totalhit. But it does not do the real scoring any more. Otherwise, we may hit latency performance issues.

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java
```diff
@@ -137,9 +130,16 @@ public void setScorer(Scorable scorer) throws IOException {
...
-          throw new CollectionTerminatedException();
         }
-        leafCollector.collect(doc);
+
+        if (!terminatedEarly) {
+          // If we haven't terminated early, collect the doc
+          leafCollector.collect(doc);
+        } else {
+          // if we have terminated early, we skip the scoring then continue to collect the doc
+          leafCollector.setScorer(null);
```
taoyyu, line 140: Please add tests for this change. Especially make sure this line won't cause NullPointerException

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java
```diff
@@ -20,13 +20,9 @@
...
 import java.util.concurrent.atomic.AtomicInteger;
+
+import org.apache.lucene.search.*;
 import org.apache.lucene.index.LeafReaderContext;
-import org.apache.lucene.search.CollectionTerminatedException;
-import org.apache.lucene.search.Collector;
-import org.apache.lucene.search.CollectorManager;
-import org.apache.lucene.search.LeafCollector;
-import org.apache.lucene.search.Scorable;
-import org.apache.lucene.search.ScoreMode;
```
taoyyu, line 29: Don't use .* Here. You can change your intellij config to avoid this.

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java
xinyiyelp: Thanks for review! Sorry this work was not done, more experimental change, actually this line will throw error, will turn this into draft PR for now until I fully test it.

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java
xinyiyelp: Thanks! has configured the IJ and disabled the * import.

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java
xinyiyelp: Thanks for review, will address these points

taoyyu (commented):
https://github.com/Yelp/nrtsearch/pull/660/files#diff-613a2312c665f7b9d29125dd2e5a42dd32cb7c06429ee412737f571e7fb50e93R280

According to the code above, the scorer is actually reset. The dummy scorer you created is not actually used here.

Also, even if you reset it in the right place, you can see code https://github.com/Yelp/nrtsearch/pull/660/files#diff-613a2312c665f7b9d29125dd2e5a42dd32cb7c06429ee412737f571e7fb50e93R286. The leaf collector is collecting the docs now. So the termianteEarly parameter will be false in the beginning and becomes true while more and more docs are collected.

xinyiyelp:
> https://github.yelpcorp.com/sysgit/datastore_schemas/blob/master/elasticlucy/elasticsearch/schemas/elasticlucy.yaml#L718
> 
> According to the code above, the scorer is actually reset. The dummy scorer you created is not actually used here.


Sorry, did you actually want to reference this [line](https://github.yelpcorp.com/sysgit/datastore_schemas/blob/master/elasticlucy/elasticsearch/schemas/elasticlucy.yaml#L718)? Sorry I don't quite understand this here.

> 
> Also, even if you reset it in the right place, you can see code https://github.com/Yelp/nrtsearch/pull/660/files#diff-613a2312c665f7b9d29125dd2e5a42dd32cb7c06429ee412737f571e7fb50e93R286. The leaf collector is collecting the docs now. So the termianteEarly parameter will be false in the beginning and becomes true while more and more docs are collected.

Yes, you are right, the effects of the new scorer only applies to the new collector, for the undergoing collector, it still uses leafCollector(hence the potentially heavy `score()` function). We could reset the scorer in the leafCollector level, but I have two concerns:

1. It probably means we will call the reset significant number of times repetitively, which might have performance impact OR
2. We introduce a flag to check if the scorer has been reset or not. However, if the piece of code is multithreaded, the flag might be messed up(Please correct me if I were wrong cause I'm not very confident of how JAVA concurrency works)

src/main/java/com/yelp/nrtsearch/server/luceneserver/MyIndexSearcher.java
```diff
@@ -250,6 +254,11 @@ protected void search(List<LeafReaderContext> leaves, Weight weight, Collector c
       final LeafCollector leafCollector;
       try {
         leafCollector = collector.getLeafCollector(ctx);
+        if (collector instanceof TerminateAfterWrapper.TerminateAfterCollectorWrapper && ((TerminateAfterWrapper.TerminateAfterCollectorWrapper) collector).getTerminatedEarly()) {
+          // overwrite the scorer as constant scorer if the search was terminated early
+          Weight dummyWeight = (new MatchAllDocsQuery()).createWeight(this, ScoreMode.COMPLETE_NO_SCORES, 1);
```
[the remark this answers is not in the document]
xinyiyelp: Yea, you are right, we could create the scorer directly without the dummy weight.

And great catch for the score of the constant scorer!

src/main/java/com/yelp/nrtsearch/server/luceneserver/MyIndexSearcher.java
```diff
@@ -250,6 +254,11 @@ protected void search(List<LeafReaderContext> leaves, Weight weight, Collector c
...
+          leafCollector.setScorer(dummyWeight.scorer(ctx));
```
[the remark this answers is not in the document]
xinyiyelp: Yup, I'll double check this part.

taoyyu:
> > https://github.yelpcorp.com/sysgit/datastore_schemas/blob/master/elasticlucy/elasticsearch/schemas/elasticlucy.yaml#L718
> > According to the code above, the scorer is actually reset. The dummy scorer you created is not actually used here.
> 
> Sorry, did you actually want to reference this [line](https://github.yelpcorp.com/sysgit/datastore_schemas/blob/master/elasticlucy/elasticsearch/schemas/elasticlucy.yaml#L718)? Sorry I don't quite understand this here.
> 
> > Also, even if you reset it in the right place, you can see code https://github.com/Yelp/nrtsearch/pull/660/files#diff-613a2312c665f7b9d29125dd2e5a42dd32cb7c06429ee412737f571e7fb50e93R286. The leaf collector is collecting the docs now. So the termianteEarly parameter will be false in the beginning and becomes true while more and more docs are collected.
> 
> Yes, you are right, the effects of the new scorer only applies to the new collector, for the undergoing collector, it still uses leafCollector(hence the potentially heavy `score()` function). We could reset the scorer in the leafCollector level, but I have two concerns:
> 
> 1. It probably means we will call the reset significant number of times repetitively, which might have performance impact OR
> 2. We introduce a flag to check if the scorer has been reset or not. However, if the piece of code is multithreaded, the flag might be messed up(Please correct me if I were wrong cause I'm not very confident of how JAVA concurrency works)

The line is updated. Sorry I pasted the wrong line before.

Yea. In this case, instead of setting the constant wrapper, We should consider creating a wrapper here https://github.com/Yelp/nrtsearch/blob/8be3759988fad972bed6a59e8876c722a4d8ad02/src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java#L132. Instead of directly set the scorer, you can create a new scorer class which extends the scorable abstract class. The scorer should have the score inside as class variables. In the new scorer score() method, you can check for termiante early and decide whether return scorer.score() or constant score.

src/main/java/com/yelp/nrtsearch/server/luceneserver/MyIndexSearcher.java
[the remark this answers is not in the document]
xinyiyelp: Sorry @taoyyu I looked into the code but found we still need a dummy weight to create the scorer. I implemented out and please leave comment if I were wrong.

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java
```diff
@@ -130,14 +134,30 @@ public TerminateAfterLeafCollectorWrapper(LeafCollector leafCollector) {
 
       @Override
       public void setScorer(Scorable scorer) throws IOException {
-        leafCollector.setScorer(scorer);
+        if (terminatedEarly) {
+          // setup constant scorer to avoid spending time on scoring
+          float score = 1;
```
xinyiyelp, line 139: the `1` is temporary, we need to finalize a value here in order the constant scores not to impact the collected hits.

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java
```diff
@@ -130,14 +134,30 @@ public TerminateAfterLeafCollectorWrapper(LeafCollector leafCollector) {
...
+            }
+
+            @Override
+            public Scorer scorer(LeafReaderContext context) throws IOException {
+              return new ConstantScoreScorer(this, this.score(), ScoreMode.COMPLETE_NO_SCORES, DocIdSetIterator.all(context.reader().maxDoc()));
+            }
+          };
+          leafCollector.setScorer(dummyWeight.scorer(currentContext));
+        } else {
+          leafCollector.setScorer(scorer);
```
taoyyu, line 153: I want you to create new scorable class which has  scorer inside,
and when not terminated early, call scorer.score()
when termianted early, return constant score.

I'm expecting you to define a nested class like below in the TerminateAfterWrapper.java.

```
class TerminatedEarlyScorer extends Scorable {
        private final Scorable in;
        ...
        
        @Override
        public float score() throws IOException{
          if (terminatedEarly) {
            return in.score();
          } else {
            // return the min constant score
          }
        }
      }
```

And use this wrapper to wrap the real scorer and set it.

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java
taoyyu: Yes. You should try to find the min score supported by lucene and use the min score.

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java
xinyiyelp: Oh Sorry @taoyyu I thought my implementation is an alternative solution to what we want to do, let me implement your approach and let's see how things go, I'll ping you when this is done

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java
xinyiyelp: Yup, sounds good, will update accordingly

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java
taoyyu: The setScorer is only called once or twice with in TerminatedAfterWrapper. You are not able to set it to a different scorer later by checking terminatedEarly. You'll have to create the scorer which can check termiantedEarly internally. You can try run the code with your approach to confirm it.

This is just a demo. You don't need to define the TermiantedEarlyScorer. In Java, you are able to directly new abstract class with override. You can check some java materials for this.

aprudhomme:
Is there a reason you can't just keep track of the total hits in the wrapper and avoid calling the inner collector? It would probably look something like https://github.com/Yelp/nrtsearch/commit/c8cb1fd271f08841bb4718ba2adfe4ed49be160c 

How is the hit count expected to behave if the search timeout is hit? Should it stop collecting or keep counting?

xinyiyelp:
> Is there a reason you can't just keep track of the total hits in the wrapper and avoid calling the inner collector? It would probably look something like [c8cb1fd](https://github.com/Yelp/nrtsearch/commit/c8cb1fd271f08841bb4718ba2adfe4ed49be160c)
> 
> How is the hit count expected to behave if the search timeout is hit? Should it stop collecting or keep counting?

Thank you @aprudhomme for your commit! I complete some of the details and opened another [PR](https://github.com/Yelp/nrtsearch/pull/666) as an alternative approach, for now it's working but I haven't fully tested that change yet and later we could benchmark these two approaches.

Thank you again for your contribution! Truly appreciate!

taoyyu:
> Is there a reason you can't just keep track of the total hits in the wrapper and avoid calling the inner collector? It would probably look something like [c8cb1fd](https://github.com/Yelp/nrtsearch/commit/c8cb1fd271f08841bb4718ba2adfe4ed49be160c)
> 
> How is the hit count expected to behave if the search timeout is hit? Should it stop collecting or keep counting?

Thanks for the review. This approach definitely works. To avoid confusion here, @xinyiyelp will continue the learning and debuging/investigation in the existing draft. We will discuss about this approach in the next step for xinyi

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java
```diff
@@ -128,16 +124,39 @@ public TerminateAfterLeafCollectorWrapper(LeafCollector leafCollector) {
         this.leafCollector = leafCollector;
       }
 
+      class TerminatedEarlyScorer extends Scorable {
```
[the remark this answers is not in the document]
xinyiyelp: Yes, I've tried to create anonymous class like following:

```
leafCollector.setScorer(new Scorable(scorer) {
          private Scorable in;

          public void Scorable(Scorable in) {
            this.in = in;
          }

          @Override
          public float score() throws IOException{
            if (!terminatedEarly) {
              return in.score();
            } else {
              // return the min constant scorer
              return 0;
            }
          }

          @Override
          public int docID() {
            return in.docID();
          }
        });

```

But the compiler does not allow me to pass in `scorer` to `new Scorable()` and the intelliJ suggested me to use the named class. That's why I used the names wrapper. @taoyyu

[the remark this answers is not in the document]
taoyyu: You do not need to define Scorable in. You can just call the scorer in the method

src/main/java/com/yelp/nrtsearch/server/luceneserver/MyIndexSearcher.java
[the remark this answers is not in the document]
xinyiyelp: Yes I confirm when we execute the simple query, the new scorer is used after terminateAfter:
Screenshot 2024-07-02 at 4 17 57 PM

src/main/java/com/yelp/nrtsearch/server/luceneserver/MyIndexSearcher.java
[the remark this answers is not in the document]
xinyiyelp: Addressed

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java
xinyiyelp: Updated

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java
xinyiyelp: thanks for the comments, updated

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java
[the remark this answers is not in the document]
xinyiyelp: Updated

PR #662, opened 2024-06-17
Add span near query support to Nrtsearch

vim345:
There's a whole bunch of other Span queries that can be nested inside a SnapNearQuery: https://lucene.apache.org/core/8_7_0/core/index.html?org/apache/lucene/search/spans/SpanNearQuery.html

They are not in the requirements to support for now. We can add them whenever they are needed.

### DIFF
--- a/clientlib/src/main/proto/yelp/nrtsearch/search.proto
+++ b/clientlib/src/main/proto/yelp/nrtsearch/search.proto
@@ -404,10 +404,9 @@ message ConstantScoreQuery {
 
 // Wrapper message for different types of SpanQuery
 message SpanQuery {
-  string field = 1;
   oneof query {
-    TermQuery spanTermQuery = 3;
-    SpanNearQuery spanNearQuery = 4;
+    TermQuery spanTermQuery = 1;
+    SpanNearQuery spanNearQuery = 2;
     // Add other SpanQuery types as needed
   }
 }
@@ -446,7 +445,7 @@ enum QueryType {
     PREFIX = 19;
     CONSTANT_SCORE_QUERY = 20;
     GEO_POLYGON = 21;
-    SPAN_NEAR_QUERY = 22;
+    SPAN_QUERY = 22;
 }
 
 // Defines a full query consisting of a QueryNode which may be one of several types.
@@ -477,7 +476,7 @@ message Query {
         PrefixQuery prefixQuery = 22;
         ConstantScoreQuery constantScoreQuery = 23;
         GeoPolygonQuery geoPolygonQuery = 24;
-        SpanNearQuery spanNearQuery = 25;
+        SpanQuery spanQuery = 25;
     }
 }
 
--- a/clientlib/src/main/proto/yelp/nrtsearch/search.proto
+++ b/clientlib/src/main/proto/yelp/nrtsearch/search.proto
@@ -402,6 +402,26 @@ message ConstantScoreQuery {
     Query filter = 1;
 }
 
+// Wrapper message for different types of SpanQuery
+message SpanQuery {
+  string field = 1;
+  oneof query {
+    TermQuery spanTermQuery = 3;
+    SpanNearQuery spanNearQuery = 4;
+    // Add other SpanQuery types as needed
+  }
+}
+
+// A query that matches documents containing terms within a specified range.
+message SpanNearQuery {
+    // Clauses for a span near query.
+    repeated SpanQuery clauses = 1;
+    // Maximum number of positions between matching terms.
+    int32 slop = 2;
+    // True if the matching terms must be in the same order as the query.
+    bool inOrder = 3;
+}
+
 // Defines different types of QueryNodes.
 enum QueryType {
     NONE = 0;
@@ -426,6 +446,7 @@ enum QueryType {
     PREFIX = 19;
     CONSTANT_SCORE_QUERY = 20;
     GEO_POLYGON = 21;
+    SPAN_NEAR_QUERY = 22;
 }
 
 // Defines a full query consisting of a QueryNode which may be one of several types.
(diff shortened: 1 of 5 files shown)


clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -402,6 +402,26 @@ message ConstantScoreQuery {
     Query filter = 1;
 }
 
+// Wrapper message for different types of SpanQuery
+message SpanQuery {
+  string field = 1;
```
aprudhomme, line 407: Is `field` needed here?

clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -456,6 +477,7 @@ message Query {
         PrefixQuery prefixQuery = 22;
         ConstantScoreQuery constantScoreQuery = 23;
         GeoPolygonQuery geoPolygonQuery = 24;
+        SpanNearQuery spanNearQuery = 25;
```
aprudhomme, line 480: Instead of adding an entry for each span query type, would it make sense to add a single entry for `SpanQuery`?

src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
```diff
@@ -643,4 +636,36 @@ private Query getConstantScoreQuery(
...
   }
+
+  private SpanNearQuery getSpanNearQuery(
+      com.yelp.nrtsearch.server.grpc.SpanNearQuery protoSpanNearQuery) {
+    List<SpanQuery> clauses = new ArrayList<>();
+
+    for (com.yelp.nrtsearch.server.grpc.SpanQuery protoClause :
+        protoSpanNearQuery.getClausesList()) {
+      SpanQuery luceneClause = convertToLuceneSpanQuery(protoClause);
+      if (luceneClause != null) {
```
aprudhomme, line 647: Can this clause ever be `null`?

src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
```diff
@@ -643,4 +636,36 @@ private Query getConstantScoreQuery(
...
+      com.yelp.nrtsearch.server.grpc.SpanQuery protoSpanQuery) {
+    com.yelp.nrtsearch.server.grpc.TermQuery termQuery = protoSpanQuery.getSpanTermQuery();
+    if (termQuery != null) {
+      return new SpanTermQuery(new Term(termQuery.getField(), termQuery.getTextValue()));
+    }
+    com.yelp.nrtsearch.server.grpc.SpanNearQuery spanNearQuery = protoSpanQuery.getSpanNearQuery();
+    if (spanNearQuery != null) {
+      return getSpanNearQuery(spanNearQuery);
+    }
+    throw new IllegalArgumentException("Unsupported Span Query: " + protoSpanQuery);
```
aprudhomme, line 669: This can be done with a switch statement using the field case value. Something similar to https://github.com/Yelp/nrtsearch/blob/master/src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java#L137

clientlib/src/main/proto/yelp/nrtsearch/search.proto
vim345: Not really. I removed it.

clientlib/src/main/proto/yelp/nrtsearch/search.proto
vim345: Yeah, that makes sense. Although this way all of span queries should be nested under `spanQuery`, but I think it is easier to maintain.

src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
vim345: No. i just removed it.

src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
vim345: Good suggestions. I applied the necessary changes.

src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
```diff
@@ -643,4 +636,52 @@ private Query getConstantScoreQuery(
     Query filterQuery = getQuery(constantScoreQueryGrpc.getFilter(), indexState, docLookup);
     return new ConstantScoreQuery(filterQuery);
   }
+
+  private SpanQuery getSpanQuery(
+      com.yelp.nrtsearch.server.grpc.SpanQuery protoSpanQuery, IndexState state) {
+    List<SpanQuery> clauses = new ArrayList<>();
+
+    if (protoSpanQuery.hasSpanNearQuery()) {
```
aprudhomme, line 644: This should be able to use a switch statement like below.

src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
```diff
@@ -643,4 +636,52 @@ private Query getConstantScoreQuery(
...
+      com.yelp.nrtsearch.server.grpc.SpanNearQuery protoSpanNearQuery =
+          protoSpanQuery.getSpanNearQuery();
+      for (com.yelp.nrtsearch.server.grpc.SpanQuery protoClause :
+          protoSpanNearQuery.getClausesList()) {
+        SpanQuery luceneClause = convertToLuceneSpanQuery(protoClause, state);
```
aprudhomme, line 649: Is there a reason this can't just recursively call `getSpanQuery`?

vim345: Yeah, the above method looks identical to this now. Removed it.

aprudhomme (approved):

PR #667, opened 2024-06-27
Created SpanMultiTermQueryWrapper queries

vim345:
I'm not including range queries in this PR as it seems to be deprecated: https://lucene.apache.org/core/2_9_4/api/core/org/apache/lucene/search/RangeQuery.html

But we still have proto definitions for it: https://github.com/Yelp/nrtsearch/blob/9ffeb9d59d74dac348f5a7f3d20709f0b7f44f5a/clientlib/src/main/proto/yelp/nrtsearch/search.proto#L254-L260

I need to do more investigations on that to see if it's better to use `TermRangeQuery` instead.

UPDATE: I ended up only implementing `TermRangeQuery` as `PointRangeQuery` isn't a subclass of `MultiTermQuery`. 

Also, the `RegexpQuery` contains an experimental [option](https://lucene.apache.org/core/8_10_1/core/org/apache/lucene/util/automaton/Automaton.html) that I didn't include in the Nrtsearch.


src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
```diff
@@ -660,8 +664,109 @@ private SpanQuery getSpanQuery(
...
+        FuzzyQuery fuzzyQuery = getFuzzyQuery(protoSpanMultiTermQueryWrapper);
+        return new SpanMultiTermQueryWrapper<>((MultiTermQuery) fuzzyQuery);
+      case PREFIXQUERY:
+        Query prefixQuery = getPrefixQuery(protoSpanMultiTermQueryWrapper.getPrefixQuery(), state);
+        if (prefixQuery instanceof MultiTermQuery) {
+          return new SpanMultiTermQueryWrapper<>((MultiTermQuery) prefixQuery);
+        }
+      case REGEXPQUERY:
+        RegexpQuery regexpQuery = getRegexpQuery(protoSpanMultiTermQueryWrapper);
+        if (regexpQuery instanceof MultiTermQuery) {
```
aprudhomme, line 701: Do we need all this type checking/casting. The queries should all inherit from `MultiTermQuery`.

src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
```diff
@@ -660,8 +664,109 @@ private SpanQuery getSpanQuery(
...
+        && maxExpansions != FuzzyQuery.defaultMaxExpansions
+        && !transpositions) {
+      return new FuzzyQuery(term, maxEdits, prefixLength, maxExpansions, transpositions);
+    } else if (maxEdits != FuzzyQuery.defaultMaxEdits
+        && prefixLength != FuzzyQuery.defaultPrefixLength) {
+      return new FuzzyQuery(term, maxEdits, prefixLength);
+    } else if (maxEdits != FuzzyQuery.defaultMaxEdits) {
+      return new FuzzyQuery(term, maxEdits);
+    }
+    return new FuzzyQuery(term);
```
aprudhomme, line 732: To make this a bit easier to understand, it may be better to set all the parameter to either the specified or default values, then invoke a single constructor will everything.

src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
vim345: I only needed to keep the casting for `prefixQuery` as it was `Query` unlike the other ones.

clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -407,10 +425,84 @@ message SpanQuery {
...
+}
+
+// Message for RegexpQuery
+message RegexpQuery {
+    string field = 1;
+    string text = 2;
+    // Optional flags for the regular expression
+    RegexpFlag flag = 3;
+    // maximum number of states that compiling the automaton for the regexp can result in. Set higher to allow more complex queries and lower to prevent memory exhaustion.
+    int32 maxDeterminizedStates = 4;
```
aprudhomme, line 499: What is the default value if not set?

src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
```diff
@@ -660,8 +666,131 @@ private SpanQuery getSpanQuery(
...
+
+    int maxEdits =
+        protoFuzzyQuery.getMaxEdits() == FuzzyQuery.defaultMaxEdits
+            ? FuzzyQuery.defaultMaxEdits
+            : protoFuzzyQuery.getMaxEdits();
+
+    int prefixLength =
+        protoFuzzyQuery.getPrefixLength() == FuzzyQuery.defaultPrefixLength
+            ? FuzzyQuery.defaultPrefixLength
+            : protoFuzzyQuery.getPrefixLength();
```
aprudhomme, line 729: I'm not sure I understand the defaults logic here. If the set value is the default value, use the default value, else use the set value. Wouldn't this just be always using the set value? I don't think this matches the doc strings behavior.

src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
```diff
@@ -660,8 +666,131 @@ private SpanQuery getSpanQuery(
...
+          default -> RegExp.ALL;
+        };
+
+    if (maxDeterminesizedStates != 0) {
+      return new RegexpQuery(term, flags, maxDeterminesizedStates);
+    }
+    RegexpQuery regexpQuery = new RegexpQuery(term, flags);
+    regexpQuery.setRewriteMethod(
+        getRewriteMethod(protoRegexpQuery.getRewrite(), protoRegexpQuery.getRewriteTopTermsSize()));
+    return regexpQuery;
```
aprudhomme, line 778: This will skip setting the rewrite method if `maxDeterminesizedStates` is specified.

src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
vim345: The default value for this is `0`. So `protoFuzzyQuery.getPrefixLength()` always returns the correct value, even if no value is set for it. If the default value is anything other than `0` using the set value will not use the correct default value. I'm not too sure if `defaultPrefixLength` would ever change in future versions of Lucene. But I think it's a good idea to keep the logic this way so that the default value won't be ever overriden.

vim345: Sure, I changed all fields with default values to `optional`.

aprudhomme (approved):

PR #669, opened 2024-07-09
Remove optional from proto definitions

vim345:
grpc-gateway is using verison `3.11.4` of protoc, which doesn't support `optional` fields. The support for optional field was only added in version `3.12`. This causes the build to fail.

I'm removing all the optional fields that I added earlier to get the build working and unblock others as a short term quick solution. There's an issue with setting default values for boolean fields to `true` and an issue if the value of the default is something other than `0`. I think using `optional` is the best way to handle such cases. That's why I will try to get the `grpc-gateway` working with the `3.12` proto as a long term solution so that we can use it to handle default values for different object types.


waziqi89 (approved):

waziqi89:
looks like v2 protoc-gen-grpc-gateway from 4 years ago has already supported the proto3. Let me try first

waziqi89 (changes requested):
[Let me come to a conclusion today before merging this PR](https://github.com/Yelp/nrtsearch/pull/670)

waziqi89 (approved):

PR #672, opened 2024-07-15
upgrade grpc-gateway to support protobuf3

waziqi89:
RP-11296
As we fixed the itests, we can reapply those changes


vim345 (approved):

waziqi89:
> Hi @waziqi89 can you please add details of manual testing to the PR description? There are no itests for grpc-gateway so please test it locally first.

We simplified the itest to not running the complicated inference query in servces/nrtsearch. And instead provide a simple test query for gateway only for sanity check. The gateway test file now is prefixed with "restful_"

https://github.yelpcorp.com/services/nrtsearch/commit/c6ed77e0181065c4371f96155bc22a6e57990f1a

sarthakn7:
> > Hi @waziqi89 can you please add details of manual testing to the PR description? There are no itests for grpc-gateway so please test it locally first.
> 
> We simplified the itest to not running the complicated inference query in servces/nrtsearch. And instead provide a simple test query for gateway only for sanity check. The gateway test file now is prefixed with "restful_"
> 

That test is in the internal repo. Can you please either add a test in this repo or do a manual test before merging?

waziqi89:
@sarthakn7 I tested with the curl version of the example in README. Also updated it in the readme: https://github.com/Yelp/nrtsearch/pull/673

PR #673, opened 2024-07-15
update readme with rest api commands

waziqi89:
Update the README file to test through grpc-gateway.

Also 
- refactor the indentation of the code
- fix a confusing point in the search example:
  - "queryText" uses full text search instead of term search, where "first vendor" will match both "first vendor" and "second vendor" in text because "vendor" are found in both.

### DIFF
--- a/search.json
+++ b/search.json
@@ -1,7 +1,0 @@
-{
-  "indexName": "testIdx",
-  "startHit": 0,
-  "topHits": 100,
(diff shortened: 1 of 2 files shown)


search.json
```diff
@@ -0,0 +1,7 @@
+{
```
sarthakn7, line 1: No need to add this file here

waziqi89: This is a local test file. Let me exclude it from the PR

PR #676, opened 2024-07-26
Merge master to 1.0.0-SNAPSHOT

sarthakn7:


src/main/java/com/yelp/nrtsearch/server/luceneserver/QueryNodeMapper.java
```diff
@@ -623,8 +650,158 @@ private static MultiTermQuery.RewriteMethod getRewriteMethod(
...
+          case REGEXP_INTERSECTION -> RegExp.INTERSECTION;
+          case REGEXP_INTERVAL -> RegExp.INTERVAL;
+          case REGEXP_NONE -> RegExp.NONE;
+          default -> RegExp.ALL;
+        };
+
+    int maxDeterminizedStates =
+        protoRegexpQuery.hasMaxDeterminizedStates()
+            ? protoRegexpQuery.getMaxDeterminizedStates()
+            : Operations.DEFAULT_DETERMINIZE_WORK_LIMIT;
```
sarthakn7, line 784: The nomenclature has been changed from max number of states to work limit. The change is there in the constructor value as well. This constant was `DEFAULT_MAX_DETERMINIZED_STATES` in master but had to be changed to `DEFAULT_DETERMINIZE_WORK_LIMIT`. The value of the constant is still 10000.
I'm not changing anything in the protobuf value yet, but if this query is not being used it might be a good idea to do so. cc: @vim345

clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -491,8 +611,10 @@ message SearchRequest {
     bool explain = 25;
     // Search nested object fields for each hit
     map<string, InnerHit> inner_hits = 26;
+    // Defines runtime fields for this query
+    repeated RuntimeField runtimeFields = 27;
     // Request kNN vector search queries, results will be combined with the standard query (if provided) using the boolean query SHOULD logic
-    repeated KnnQuery knn = 27;
+    repeated KnnQuery knn = 28;
```
sarthakn7, line 617: The field number for knn had to be changed to 28 since 27 is used by runtimeFields.

sarthakn7:
This test will fail because java 17 has been removed now but it's passing locally.

aprudhomme (approved):

PR #677, opened 2024-07-29
Deprecate sort and group options and clean up docvalue typing for TEXT and ATOM fields

sarthakn7:
RP-9680

Deprecate sort and group field options. For fields other than TEXT or ATOM, just specifying `storeDocValues: true` will result in the correct docvalue type being picked. For TEXT and ATOM fields, a new property `textDocValuesType` has been added which can be used to specify if sorted or binary docvalues are desired. Not specifying the field will continue to use the current defaults (sorted for text and binary for atom). Multivalued text and atom fields will always use sorted set.

Usages of sort and group have also been removed from all tests.


clientlib/src/main/proto/yelp/nrtsearch/luceneserver.proto
```diff
@@ -571,6 +577,8 @@ message Field {
     string vectorSimilarity = 31;
     // Indexing options for search enabled VECTOR field type. This is optional, defaulting to HNSW with m=16, ef_construction=100 when not set.
     VectorIndexingOptions vectorIndexingOptions = 32;
+    //  Specify docvalues type for TEXT/ATOM field types. This is optional, defaulting to SORTED for TEXT and BINARY for ATOM when not set. Multivalued fields will always use SORTED_SET.
```
aprudhomme, line 580: What do you think about the default doc value types, does it still make sense to have `TEXT` be sorted and `ATOM` binary? This would be the opportunity to change it if we want to.

sarthakn7: These are their javadocs for reference:
1. Binary
A per-document byte[]. Values may be larger than 32766 bytes, but different codecs may enforce their own limits.

2. Sorted
A pre-sorted byte[]. Fields with this type only store distinct byte values and store an additional offset pointer per document to dereference the shared byte[]. The stored byte[] is presorted and allows access via document id, ordinal and by-value. Values must be <= 32766 bytes.

It's hard to say if either of these should be the default for text or atom. 
Binary can make sense for atom when it's used for unique values like ids, names, etc. Sorted might just end up taking more space. But then it's also used for address fields where state, country, city, etc. can be common and sorted might be better. There are also things like categories where atom is used.
Same for text. There are some types of text fields where each value is unique and maybe long - e.g. reviews or comments. But there are short values too which are not always unique, and sorted would be better for those.

One thing we can consider is to keep binary default for both of the fields for consistency, and override to use sorted when needed.

aprudhomme: I think it might make more sense to go the other way, with everything defaulted to `SORTED`. Ordinal based fields would be more efficient for aggregations. This would also be consistent with the elasticsearch keyword, which is always `SORTED_SET` https://github.com/elastic/elasticsearch/blob/d8df7e434225cc9ac320447c184fb4d9cfc5a6ae/server/src/main/java/org/elasticsearch/index/mapper/KeywordFieldMapper.java#L100

sarthakn7: > Ordinal based fields would be more efficient for aggregations. 
This makes sense, but we don't use aggregations for most fields. With Kibana, Elasticsearch users might end up using aggregations way more than we would.
Also, should we merge this PR which retains the default behavior and then open a new one to change it?

aprudhomme: Yeah, it can be done as a follow up. It might also be good to add indexing size checks at the same time for the sorted case.

aprudhomme (approved):

PR #682, opened 2024-08-07
Expose intra-merge parallelism for vectors

aprudhomme:
Allow use of intra-merge parallelism for vector fields.

Vector fields can now specify `merge_workers` to set merge parallelism. When set to a value > 1, this uses a new `VECTOR_MERGE` thread pool.

To facilitate obtaining the thread pool, I have modified the `ThreadPoolExecutorFactory` to be a static singleton. I also added thread pool caching by type to ensure only one thread pool per type will be created.


src/main/java/com/yelp/nrtsearch/server/utils/ThreadPoolExecutorFactory.java
```diff
@@ -36,19 +34,57 @@ public enum ExecutorType {
...
+    instance = new ThreadPoolExecutorFactory(threadPoolConfiguration);
+  }
+
   /**
+   * Get the instance of the factory.
+   *
+   * @return instance of the factory
+   */
+  public static ThreadPoolExecutorFactory getInstance() {
+    return instance;
```
sarthakn7, line 64: Do you want to add a null check here and throw an exception if `instance` is not initialized? Would also require setting it to null initially.

sarthakn7 (approved):

PR #683, opened 2024-08-08
Add support of quantized vector search

aprudhomme:
Adds support for quantized vectors for search, using the provided `Lucene99HnswScalarQuantizedVectorsFormat`. This feature only works with the `FLOAT` element type.

New indexing options:
- `quantized_confidence_interval` - The confidenceInterval the Lucene codec uses for scalar quantizing the vectors. Can be unset, 0, or from 0.9 to 1.0.
- `quantized_bits` - Number of bit to quantize the vector values to. Can be 4, 7, or 8.
- `quantized_compress` - For 4 bit quantization, if multiple values should be compressed into a single byte.


src/main/java/com/yelp/nrtsearch/server/luceneserver/field/VectorFieldDef.java
```diff
@@ -143,11 +207,11 @@ protected VectorFieldDef(String name, Field requestField) {
...
     this.elementType = requestField.getVectorElementType();
     if (isSearchable()) {
+      VectorSearchType vectorSearchType =
+          getSearchType(requestField.getVectorIndexingOptions(), elementType);
       this.similarityFunction = getSimilarityFunction(requestField.getVectorSimilarity());
       this.vectorsFormat =
-          requestField.hasVectorIndexingOptions()
-              ? createVectorsFormat(requestField.getVectorIndexingOptions())
-              : null;
+          createVectorsFormat(vectorSearchType, requestField.getVectorIndexingOptions());
```
sarthakn7, line 214: Do we want to add the vector format even when indexing options are not specified because we'll pick defaults? Is there a case where we would add a vector field but wouldn't want to add this?

aprudhomme: I think we will always want to create the vector format now, since otherwise it won't get the new size limit.

aprudhomme: This is in an `isSearchable()` block. The vector format is always null if search is false.

sarthakn7: Ah I missed that, I was looking for `getSearch()`. Thanks for clarifying!

sarthakn7 (approved):

PR #687, opened 2024-08-13
Use primitive types for doc value loading

aprudhomme:
Change `LoadedDocValues` implementation for basic types (boolean, int, long, float, double) to load as primitive types. This prevents all the loaded data from being boxed on load. This will reduce garbage when only some of the values are accessed. If also makes future optimizations possible (avoiding boxing for aggregations, etc).

Also, the `getValue` method structure has been made consistent.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/doc/LoadedDocValues.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/doc/LoadedDocValues.java
@@ -15,10 +15,12 @@
  */
 package com.yelp.nrtsearch.server.luceneserver.doc;
 
-import com.google.gson.Gson;
+import static org.apache.lucene.util.ArrayUtil.oversize;
+
 import com.google.protobuf.ByteString;
 import com.google.protobuf.ListValue;
 import com.google.protobuf.Struct;
+import com.google.protobuf.Value;
 import com.google.protobuf.util.JsonFormat;
 import com.google.type.LatLng;
 import com.yelp.nrtsearch.server.grpc.SearchResponse;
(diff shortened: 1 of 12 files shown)


src/main/java/com/yelp/nrtsearch/server/luceneserver/doc/LoadedDocValues.java
```diff
@@ -190,6 +366,10 @@ public SingleDateTime(NumericDocValues docValues) {
       super(docValues, DATE_DECODER);
     }
 
+    public Instant getValue() {
```
sarthakn7, line 369: SingleDataTime is the only implementation of SingleNumericValue now. Do you want to remove the SingleNumericValue class and put its functionality here?

src/main/java/com/yelp/nrtsearch/server/luceneserver/doc/LoadedDocValues.java
```diff
@@ -217,7 +397,6 @@ public void setDocId(int docID) throws IOException {
           values.add(decoder.apply(docValues.nextValue()));
         }
       }
-      values.trimToSize();
```
sarthakn7, line 220: Can this result in higher heap usage? Or is the array already right-sized?

src/main/java/com/yelp/nrtsearch/server/luceneserver/doc/LoadedDocValues.java
aprudhomme: I think it is ok to leave `SingleNumericValue` for now, since it would make it easier to add new implementations. Though I am considering breaking these into individual class files, so maybe it can be done then.

src/main/java/com/yelp/nrtsearch/server/luceneserver/doc/LoadedDocValues.java
aprudhomme: Trimming resizes the internal array to match the current number of elements. This will result in more arrays being created and data copy operations. I think it is best to avoid this. It may hold a little more active memory at times, but no more than the size of the largest array.

sarthakn7 (approved):

PR #691, opened 2024-08-15
Add terminateAfterMaxRecallCount to track total hits

swethakann:
Add a new field to track total hits until a max threshold value. 

1. terminateAfter is not set

- No effect

2. terminateAfter =< terminateAfterMaxRecallCount (beyond terminateAfter, only counting and not scoring)
totalHits = terminateAfterMaxRecallCount (regardless of recall size, the totalHits should be capped at terminateAfterMaxRecallCount)
terminatedEarly = true

3. terminateAfter > terminateAfterMaxRecallCount (scoring + counting)
totalHits = terminateAfter
terminatedEarly = true

4. If timeout
Timeout exception.


src/main/java/com/yelp/nrtsearch/server/luceneserver/search/collectors/DocCollector.java
```diff
@@ -186,8 +186,14 @@ <C extends Collector> CollectorManager<? extends Collector, SearcherResult> wrap
...
             : indexState.getDefaultTerminateAfter();
+    int terminateAfterMaxRecallCount =
+        request.getTerminateAfterMaxRecallCount() > 0
+            ? request.getTerminateAfterMaxRecallCount()
+            : 0;
     if (terminateAfter > 0) {
-      wrapped = new TerminateAfterWrapper<>(wrapped, terminateAfter, () -> terminatedEarly = true);
+      wrapped =
+          new TerminateAfterWrapper<>(
+              wrapped, terminateAfter, terminateAfterMaxRecallCount, () -> terminatedEarly = true);
```
aprudhomme, line 196: When both terminate after and search timeout are used, the terminate after wraps the timeout. Do you want the timeout to always have priority? If so, this order may need to be reversed.

src/main/java/com/yelp/nrtsearch/server/luceneserver/search/TerminateAfterWrapper.java
```diff
@@ -77,7 +77,14 @@ public SearcherResult reduce(Collection<TerminateAfterCollectorWrapper> collecto
...
     }
-    return in.reduce(innerCollectors);
+    SearcherResult searcherResult = in.reduce(innerCollectors);
+    int totalHits = 0;
+    for (TerminateAfterCollectorWrapper collector : collectors) {
+      totalHits += collector.docCount;
+    }
+    searcherResult.getTopDocs().totalHits =
+        new TotalHits(totalHits, TotalHits.Relation.GREATER_THAN_OR_EQUAL_TO);
+    return searcherResult;
```
aprudhomme, line 87: If there was not an early termination, you can skip doing the total hits rewrite.

aprudhomme (approved):

PR #695, opened 2024-08-16
Support HitsLoggerPlugin

fragosoluana:
Adding a new plugin interface: HitsLoggerPlugin which will enable NRTSearch to log hits during the FetchTask phase (Next step: add implementation for hitsCount which allows the option to return a smaller amount of hits compared to the number of hits being logged)

General request flow is very similar to the HighligherPlugin

- added a new creator - HitsLoggerCreator to initialize the plugins at the startup time and register them, so that later on, the search request processor can select the corresponding HitsLogger out of the registry
- defined the HitsLogger interface and the HitsLoggerPlugin interface to provide the custom hits logger. Each hits logger plugin must implement getHitsLogger() to log hits


src/main/java/com/yelp/nrtsearch/server/grpc/LuceneServer.java
```diff
@@ -363,6 +364,7 @@ private void initExtendableComponents(
       RescorerCreator.initialize(configuration, plugins);
       ScriptService.initialize(configuration, plugins);
       SimilarityCreator.initialize(configuration, plugins);
+      HitsLoggerCreator.initialize(configuration, plugins);
```
aprudhomme, line 367: nit: this block should be in alphabetical order

src/main/java/com/yelp/nrtsearch/server/luceneserver/logging/HitsLoggerCreator.java
```diff
@@ -0,0 +1,78 @@
...
+   * Constructor.
+   *
+   * @param configuration server configuration
+   */
+  public HitsLoggerCreator(LuceneServerConfiguration configuration) {}
+
+  private void register(HitsLoggerProvider<?> hitsLoggerProvider) {
+    if (this.hitsLoggerProvider != null) {
+      throw new IllegalArgumentException("Hits logger already exists");
+    }
```
aprudhomme, line 39: This limitation would mean only one implementation of a logger can be active on a cluster at one time. That is pretty restrictive.

Is there a reason not use a mapping on name -> implementation, like we do for most the other extendable components? The query would then specify the implementation to use.

src/main/java/com/yelp/nrtsearch/server/luceneserver/logging/HitsLoggerFetchTask.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.List;
+
+/**
+ * Implementation of {@link FetchTask} which holds the required context to be able to log hits for a
+ * search request.
+ */
+public class HitsLoggerFetchTask implements FetchTask {
+  private final HitsLogger hitsLogger;
+
+  public HitsLoggerFetchTask(HitsLoggerCreator hitsLoggerCreator, LoggingHits loggingHits) {
```
aprudhomme, line 32: Is there a reason to pass the `HitsLoggerCreator`, it is a static singleton

clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -1065,6 +1067,14 @@ message FilterResult {
     map<string, CollectorResult> nestedCollectorResults = 2;
 }
 
+message LoggingHits {
+    // number of hits to log. For example, totalHits = 20 and hitsCount = 50 means that 50 documents will be logged
+    // while 20 documents will be returned in the Search Response
+    int32 hitsCount = 1;
```
aprudhomme, line 1073: I don't think I see this value being used anywhere. Also, what happens when `hitsCount` is less than `topHits`? Is there a default value?

fragosoluana: I added here to have LoggingHits already defined but I will use hitsCount in an upcoming PR. In a second thought, it is better to include hitsCount in the following PR with tests.

src/main/java/com/yelp/nrtsearch/server/luceneserver/logging/HitsLoggerFetchTask.java
fragosoluana: Good catch! I was following what has been done in [HighlightFetchTask here](https://sourcegraph.yelpcorp.com/Yelp/nrtsearch@ef18628bb63e04617cda088cda7531178c8d9236/-/blob/src/main/java/com/yelp/nrtsearch/server/luceneserver/highlights/HighlightFetchTask.java?L49) and missed that I can just call HitsLoggerCreator.getInstance().createHitsLogger(loggingHits) instead

src/main/java/com/yelp/nrtsearch/server/luceneserver/logging/HitsLoggerCreator.java
fragosoluana: Good question. I actually thought about the two solutions, but thinking about it again, I don't think being very restrictive here makes sense anymore. I thought that a single logger active on a cluster would be enough but being flexible and allowing multiple logger implementations is more beneficial, especially for clusters with multiple index. I will update the solution to support multiple loggers.

src/main/java/com/yelp/nrtsearch/server/plugins/HitsLoggerPlugin.java
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Plugin interface for providing custom logging. The plugins will be loaded at the startup time of
+ * the {@link com.yelp.nrtsearch.server.grpc.LuceneServer}. The hits logger instances provided from
+ * the getHitsLoggers() will be responsible for handling the corresponding hits logger task.
+ * Therefore, do not alter the instance objects for any requests.
+ */
+public interface HitsLoggerPlugin {
+  default Map<String, HitsLoggerProvider<? extends HitsLogger>> getHitsLoggers() {
+    return null;
```
aprudhomme, line 30: It would be better to return an empty map instead

clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -614,6 +614,8 @@ message SearchRequest {
     map<string, InnerHit> inner_hits = 26;
     // Defines runtime fields for this query
     repeated RuntimeField runtimeFields = 27;
+    // Any custom logging that should log hits after ranking
+    LoggingHits loggingHits = 28;
```
aprudhomme, line 618: Can you make this field id `29`, 28 is used in the v1 branch

aprudhomme (commented):
Looks good

src/main/java/com/yelp/nrtsearch/server/plugins/HitsLoggerPlugin.java
fragosoluana: Totally agree. I missed this change, thanks for pointing this out!

taoyyu (commented):
Looks good to me.
Just one minor comment for future work.
LoggingHits is defined in nrtsearch but it seems it will only work with plugins now. It might be good to have a simple default HitLogger which logs response to console. But this does not affect the ongoing task and could be in future prs

fragosoluana:
@taoyyu I like your suggestion! I will keep it in mind and potentially add in an upcoming PR.

aprudhomme (approved):

taoyyu (approved):

PR #696, opened 2024-08-20
Use yyyyMMddHHmmssSSS formatted datetime instead of UUID for index id

sarthakn7:
RP-9696

Use yyyyMMddHHmmssSSS formatted datetime instead of UUID for index id so that we can make sense of the order of ids both programmatically and also when manually looking at files in S3.

I have replaced the index id in the commands instead of moving to legacy as I don't think we need backward compatibility here. When needing to operate in old index id, we would use old nrtsearch version anyway.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/utils/IndexIdUtil.java
+++ b/src/main/java/com/yelp/nrtsearch/server/utils/IndexIdUtil.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2024 Yelp Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.yelp.nrtsearch.server.utils;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+public class IndexIdUtil {
+
+  private static final DateTimeFormatter INDEX_ID_FORMATTER =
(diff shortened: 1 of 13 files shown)


src/main/java/com/yelp/nrtsearch/tools/nrt_utils/legacy/incremental/RestoreIncrementalCommand.java
```diff
@@ -27,6 +27,7 @@
 import com.amazonaws.services.s3.transfer.TransferManagerBuilder;
```
aprudhomme, line 27: The `RestoreIncrementalCommand` is a frozen legacy command. It should not be updated so that is continues to work with v0 clusters. We will create new versions of there commands after we finish the backend structural changes.

Note this only applies the legacy commands, the state command below is fine to update.

src/main/java/com/yelp/nrtsearch/tools/nrt_utils/state/UpdateGlobalIndexStateCommand.java
```diff
@@ -70,9 +71,10 @@ public class UpdateGlobalIndexStateCommand implements Callable<Integer> {
   private String credsProfile;
 
   @CommandLine.Option(
-      names = {"--setUUID"},
-      description = "If specified, update index UUID to this value")
-  private String uuid;
+      names = {"--setDateTime"},
```
aprudhomme, line 74: Could we maybe make this a more general option name, like `--setId`

src/main/java/com/yelp/nrtsearch/server/utils/IndexIdUtil.java
```diff
@@ -0,0 +1,40 @@
...
+import java.time.format.DateTimeFormatter;
+
+public class IndexIdUtil {
+
+  private static final DateTimeFormatter INDEX_ID_FORMATTER =
+      DateTimeFormatter.ofPattern("yyyyMMddHHmmssSSS");
+
+  /** Generate a unique index id based on the current time formatted as yyyyMMddHHmmssSSS. */
+  public static String generateIndexId() {
+    return INDEX_ID_FORMATTER.format(LocalDateTime.now());
```
aprudhomme, line 28: I assume this would have variable output depending on the local timezone. It might be better to standardize everything to utc.

aprudhomme (approved):

PR #699, opened 2024-08-22
Search After Support for NRTSearch

taoyyu:
Added LastHitInfo in request and response search state(did not remove duplicate fields in searchState for backward compatibility). Passed to collector as searchAfter FieldDoc.
Only added to SortFieldCollector.
Added test with docId(primary use case), virtualField and LatLon.


src/main/java/com/yelp/nrtsearch/server/luceneserver/search/sort/SortParser.java
```diff
@@ -125,6 +129,78 @@ public static Map<String, CompositeFieldValue> getAllSortedValues(
...
+      case STRING:
+      case STRING_VAL:
+        sortValue = stringValue;
+        break;
+      case CUSTOM:
+        if (fd instanceof VirtualFieldDef || fd instanceof LatLonFieldDef) {
+          sortValue = Double.valueOf(stringValue);
+        } else {
+          sortValue = stringValue;
+        }
```
aprudhomme, line 196: It might be good to add something to the `Sortable` interface to let the fields parse the custom values

aprudhomme (approved):

PR #709, opened 2024-09-06
Add float vector script param

aprudhomme:
Add a script parameter type for a float vector. This is used to provide a `float[]` to the script. The advantage is that this type can be used for vector based computations more efficiently than a `List<Float>`.

### DIFF
--- a/clientlib/src/main/java/com/yelp/nrtsearch/server/utils/ScriptParamsUtils.java
+++ b/clientlib/src/main/java/com/yelp/nrtsearch/server/utils/ScriptParamsUtils.java
@@ -73,6 +73,21 @@ public class ScriptParamsUtils {
   }
 
   /**
+   * Decode a {@link com.yelp.nrtsearch.server.grpc.Script.ParamFloatVectorValue} message into an
+   * equivalent float array.
+   *
+   * @param vectorValue vector message
+   * @return float array
+   */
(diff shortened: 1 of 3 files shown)


clientlib/src/main/java/com/yelp/nrtsearch/server/utils/ScriptParamsUtils.java
```diff
@@ -72,6 +72,21 @@ public static List<Object> decodeList(Script.ParamListValue listValue) {
...
+   * equivalent float array.
+   *
+   * @param vectorValue vector message
+   * @return float array
+   */
+  public static float[] decodeFloatVector(Script.ParamFloatVectorValue vectorValue) {
+    float[] vector = new float[vectorValue.getValuesCount()];
+    for (int i = 0; i < vectorValue.getValuesCount(); i++) {
+      vector[i] = vectorValue.getValues(i);
+    }
```
sarthakn7, line 86: Can we use the toArray method?

aprudhomme: I don't believe that will work. It would give you a `Float[]`, but we want to create the primitive version.

sarthakn7 (approved):

PR #735, opened 2024-09-27
add ignore_above support for nrtsearch

taoyyu:
Ouroboros cluster has long text which can reach the lucene limit 32766 and ingestion got stuck.
They use ignore_above in the legacy ES schema. Adding same support for NRT.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/field/TextBaseFieldDef.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/field/TextBaseFieldDef.java
@@ -77,7 +77,7 @@ public abstract class TextBaseFieldDef extends IndexableFieldDef
     indexAnalyzer = parseIndexAnalyzer(requestField);
     searchAnalyzer = parseSearchAnalyzer(requestField);
     eagerFieldGlobalOrdinals = requestField.getEagerFieldGlobalOrdinals();
-    ignoreAbove = requestField.getIgnoreAbove();
+    ignoreAbove = requestField.hasIgnoreAbove() ? requestField.getIgnoreAbove() : Integer.MAX_VALUE;
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


src/main/java/com/yelp/nrtsearch/server/luceneserver/field/TextBaseFieldDef.java
```diff
@@ -232,6 +234,9 @@ public void parseDocumentField(
 
     for (int i = 0; i < fieldValues.size(); i++) {
       String fieldStr = fieldValues.get(i);
+      if (ignoreAbove > 0 && fieldStr.length() > ignoreAbove) {
+        continue;
+      }
```
aprudhomme, line 239: Does this value still need to be retrievable somehow? This would keep it from being stored anywhere.

taoyyu: Goal of ignoreAbove is to avoid the lucene limit 32766. docValue also has this limit for Sorted and SortedSet. It will be only retrievable if it is Binary DocValue or Stored. So I just dropped it for all.

Do you think I should keep for Stored fields only?

aprudhomme: I see that for elasticsearch this setting keeps the value from being indexed or stored. Though for ES the _source would still contain the value.

I assume the client will want the value at some point (?). This could maybe be done in the field itself, though a child field could also be used potentially.

taoyyu: I'm not sure what would be the best approach to do it in the field itself. I'll store the binary doc value if the doc type is binary already(It should work with Orbs case here.  All large text field in orbs are single value text).

If it field is declared with Sorted or Sorted Set doc value, I cannot store it as Binary for large text, it will break sorting. Or if I store with the stored field, the fillDocTask will try to read from doc values since the field definition hasDocValues is true.  It would be better if we can force retrieve from doc value. or stored fields, then we can retrieve the text from stored values.

I'll add the logic to store binary doc value for the single text field first to make sure orbs can work. For further improvements, I think we can do it in a separate pr since the current master breaks for large text any way.

aprudhomme (approved):

PR #765, opened 2024-10-15
Add nodeInfo grpc endpoint

aprudhomme:
Added the grpc endpoint `nodeInfo`. This allows you to get the node name, service name, host name, nrtsearch version, and ephemeral id of a running node.

I added the utility method `handleUnaryRequest` to the `Handler` to try to standardize basic error handling and logging for grpc requests. The plan is to migrate more requests to use this in the (near) future.


src/main/java/com/yelp/nrtsearch/server/handler/Handler.java
```diff
@@ -59,6 +62,41 @@ public StreamObserver<T> handle(StreamObserver<S> responseObserver) {
...
+   * @param responseObserver observer to send response to
+   * @param handler handler to process the request
+   * @param <T> request type
+   * @param <S> response type
+   */
+  public static <T extends GeneratedMessageV3, S extends GeneratedMessageV3>
+      void handleUnaryRequest(
+          String name, T protoRequest, StreamObserver<S> responseObserver, Handler<T, S> handler) {
+    try {
+      handler.handle(protoRequest, responseObserver);
```
sarthakn7, line 79: If doing this, do you want to add another handle method that returns the response and doesn't need responseObserver? It's currently passed just for error handling.

src/test/java/com/yelp/nrtsearch/server/handler/NodeInfoHandlerTest.java
```diff
@@ -0,0 +1,56 @@
...
+    nodeInfoHandler.handle(NodeInfoRequest.newBuilder().build(), mockResponseObserver);
+
+    ArgumentCaptor<NodeInfoResponse> captor = ArgumentCaptor.forClass(NodeInfoResponse.class);
+    verify(mockResponseObserver, times(1)).onNext(captor.capture());
+    NodeInfoResponse response = captor.getValue();
+    assertEquals("nodeName", response.getNodeName());
+    assertEquals("serviceName", response.getServiceName());
+    assertEquals("hostName", response.getHostName());
+    assertEquals("ephemeralId", response.getEphemeralId());
+    assertFalse(response.getVersion().isEmpty());
```
sarthakn7, line 53: I think you can assert the version to be equal to `Version.CURRENT.toString()`

sarthakn7 (approved):

PR #773, opened 2024-10-24
Expose more server endpoint through cli commands

aprudhomme:
Added cli commands missing for endpoints:
- DeleteByQuery
- NodeInfo
- GlobalState
- Metrics
- Custom

I skipped the `state` endpoint for now, since I intend to replace it with a better version. I also skipped the snapshot endpoints, since they likely need to be reworked a bit.


src/main/java/com/yelp/nrtsearch/server/grpc/NrtsearchClient.java
```diff
@@ -400,6 +410,28 @@ public List<String> getIndices() {
         .collect(Collectors.toList());
   }
 
+  public void nodeInfo() {
```
sarthakn7, line 413: Do we really need these functions in NrtsearchClient? Can we just to `client.getBlockingStub().<method>` directly in the Command classes? Most of these methods are not reused either.

sarthakn7 (approved):

PR #775, opened 2024-10-29
Add indexState endpoint

aprudhomme:
Adds the `indexState` endpoint to retrieve the index state as the `IndexStateInfo` protobuf message. Also, adds the corresponding cli command.


src/main/java/com/yelp/nrtsearch/server/grpc/NrtsearchServer.java
```diff
@@ -672,6 +675,12 @@ public void state(StateRequest request, StreamObserver<StateResponse> responseOb
       Handler.handleUnaryRequest("state", request, responseObserver, getStateHandler);
     }
 
+    @Override
+    public void indexState(
```
sarthakn7, line 679: Do you want to add a test that calls this endpoint?

sarthakn7 (approved):

PR #777, opened 2024-10-30
Clean up some style warnings

aprudhomme:
Went through and fixed a bunch of style warnings from my IDE.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/handler/AddDocumentHandler.java
+++ b/src/main/java/com/yelp/nrtsearch/server/handler/AddDocumentHandler.java
@@ -61,12 +61,12 @@ public class AddDocumentHandler extends Handler<AddDocumentRequest, AddDocumentR
   public StreamObserver<AddDocumentRequest> handle(
       StreamObserver<AddDocumentResponse> responseObserver) {
     return new StreamObserver<>() {
-      Multimap<String, Future<Long>> futures = HashMultimap.create();
+      final Multimap<String, Future<Long>> futures = HashMultimap.create();
       // Map of {indexName: addDocumentRequestQueue}
-      Map<String, ArrayBlockingQueue<AddDocumentRequest>> addDocumentRequestQueueMap =
+      final Map<String, ArrayBlockingQueue<AddDocumentRequest>> addDocumentRequestQueueMap =
           new ConcurrentHashMap<>();
       // Map of {indexName: count}
-      Map<String, Long> countMap = new ConcurrentHashMap<>();
+      final Map<String, Long> countMap = new ConcurrentHashMap<>();
 
       private int getAddDocumentsMaxBufferLen(String indexName) {
         try {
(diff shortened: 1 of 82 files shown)


src/main/java/com/yelp/nrtsearch/server/handler/AddDocumentHandler.java
```diff
@@ -61,12 +61,12 @@ public AddDocumentHandler(GlobalState globalState) {
   public StreamObserver<AddDocumentRequest> handle(
       StreamObserver<AddDocumentResponse> responseObserver) {
     return new StreamObserver<>() {
-      Multimap<String, Future<Long>> futures = HashMultimap.create();
+      final Multimap<String, Future<Long>> futures = HashMultimap.create();
```
sarthakn7, line 64: Why is it that these variables are made final but some other variables in this same file are not?
I also prefer to avoid making function-scope variables final - I feel like it adds verbosity but not much value in most cases.

src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
```diff
@@ -810,60 +809,61 @@ private CompositeFieldValue getFieldForHit(
...
+          RuntimeScript values = segmentFactory.newInstance(leaf);
+          int docID = hit.getLuceneDocId() - leaf.docBase;
+          // Check if the value is available for the current document
+          if (values != null) {
+            values.setDocId(docID);
+            Object obj = values.execute();
+            ObjectToCompositeFieldTransformer.enrichCompositeField(obj, compositeFieldValue);
+          }
+        }
+        case IndexableFieldDef<?> fieldDef when fieldDef.hasDocValues() -> {
```
sarthakn7, line 845: TIL switch expression doesn't need explicit instanceof, and also the matching with `when`. I'm finding these large case blocks a bit odd, but I guess I can get used to these.

src/main/java/com/yelp/nrtsearch/server/nrt/NRTReplicaNode.java
```diff
@@ -194,13 +194,7 @@ private static CopyState readCopyState(com.yelp.nrtsearch.server.grpc.CopyState
...
     Map<String, FileMetaData> files = readFilesMetaData(copyState.getFilesMetadata());
 
-    int count = copyState.getCompletedMergeFilesCount();
-    assert count == copyState.getCompletedMergeFilesCount();
-
-    Set<String> completedMergeFiles = new HashSet<>();
-    for (String completedMergeFile : copyState.getCompletedMergeFilesList()) {
-      completedMergeFiles.add(completedMergeFile);
-    }
+    Set<String> completedMergeFiles = new HashSet<>(copyState.getCompletedMergeFilesList());
```
sarthakn7, line 197: Some of these are pretty good, like this one might be better performant too (if the JIT was not optimizing it already).

sarthakn7 (approved):

src/main/java/com/yelp/nrtsearch/server/handler/AddDocumentHandler.java
aprudhomme: These are not really function-scope variables, they are class variables for the `StreamObserver` instance.

PR #778, opened 2024-11-01
Use FieldDefCreatorContext for field creation

aprudhomme:
Add the context class `FieldDefCreatorContext` to be used by `FieldDef` implementations during creation. The context currently only contains `NrtsearchConfig`, but more items can be added as needed.


sarthakn7 (approved):
What else are we planning to add in this context in the future?

aprudhomme:
> What else are we planning to add in this context in the future?

I am going to need some changes to the `ExecutorFactory` so that it will need to be passed to the vector field through this context. We may also need to pass other services in the future, if we move away from static instances.

PR #787, opened 2024-11-15
Add default value for terminateAfterMaxRecallCount

swethakann:
Adding a defaultTerminateAfterMaxRecallCount param so that we can use this for benchmarking.


aprudhomme:
This branch will need to be updated to resolve conflicts with main

PR #789, opened 2024-11-15
Support hits to log count

fragosoluana:
This PR introduces a new feature in NRTSearch for HitsLoggerPlugin that allows for logging additional hits beyond what is returned to the client. 

- The system can now log more hits than are returned to the client, allowing the client to log a broader set of data without adding too much information in the search response.
- Users can configure the number of hits to be returned separately from the number of hits to be logged. 
- Existing configurations will continue to function as before unless explicitly updated to use hitsToLog in LoggingHits message.

Testing
- Unit tests have been added to verify the correct behavior of the response size reduction functionalities.
- Added examples to demonstrate how to set up and utilize the new feature


src/main/java/com/yelp/nrtsearch/server/search/collectors/DocCollector.java
```diff
@@ -67,8 +67,11 @@ public DocCollector(
   }
 
   public static int computeNumHitsToCollect(SearchRequest request) {
-    // determine how many hits to collect based on request, facets and rescore window
+    // determine how many hits to collect based on request, facets, rescore window and hits to log
     int collectHits = request.getTopHits();
+    if (request.hasLoggingHits()) {
+      collectHits = request.getLoggingHits().getHitsToLog();
+    }
```
aprudhomme, line 74: What if `hitsToLog` is less than `topHits`? This might be better as a `max` operation

fragosoluana: ++ I missed adding `Max` operation here

src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
```diff
@@ -271,14 +271,25 @@ public SearchResponse handle(IndexState indexState, SearchRequest searchRequest)
 
       long t0 = System.nanoTime();
 
-      hits = getHitsFromOffset(hits, searchContext.getStartHit(), searchContext.getTopHits());
+      // hits to be logged also need to have their fields fetched
+      hits =
+          getHitsFromOffset(
+              hits,
+              searchContext.getStartHit(),
+              Math.max(searchContext.getTopHits(), searchContext.getHitsToLog()));
```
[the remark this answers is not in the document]
fragosoluana: Thanks for explaining this! My understanding was that `getHitsFromOffset` would take care of `startHit`. I am basically treating `hitsToLog` as the `topHits` if it's bigger than `topHits`. I then remove any extra hits in the new method`setResponseTopHits`. Not sure if I am missing something or if there is a better approach. I added a unit test to test what you mentioned to make sure this code is working as expected and the test is passing.

src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
```diff
@@ -271,14 +271,29 @@ public SearchResponse handle(IndexState indexState, SearchRequest searchRequest)
 
       long t0 = System.nanoTime();
 
-      hits = getHitsFromOffset(hits, searchContext.getStartHit(), searchContext.getTopHits());
+      if (searchContext.getFetchTasks().getHitsLoggerFetchTask() != null) {
```
fragosoluana, line 274: I wasn't too sure if I should create a new method or just modify `getHitsFromOffset` to require hitsToLog as another parameter. I decided to create a new method to make it more clear that the logic is a bit different when there are hits to log. Also, modifying the `getHitsFromOffset` would need to update its call in other two classes: `InnerHitFetchTask` and `TopHitsCollectorManager`. 

I don't have a strong opinion on which approach to choose so let me know if you prefer one over the other or if you have other suggestions.

src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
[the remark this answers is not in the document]
fragosoluana: Thanks a lot for this explanation! I added some unit tests to make sure the code is working as expected now.

src/test/java/com/yelp/nrtsearch/server/logging/HitsLoggerTest.java
```diff
@@ -185,11 +224,103 @@ public void testCustomHitsLoggerWithoutParam() {
...
+                            .setField("vendor_name")
+                            .setTextValue("vendor")
+                            .build())
+                    .build())
+            .setLoggingHits(
+                LoggingHits.newBuilder().setName("custom_logger").setHitsToLog(6).build())
+            .build();
+    SearchResponse response = getGrpcServer().getBlockingStub().search(request);
+    String expectedLogMessage =
+        "LOGGED doc_id: 4, doc_id: 5, doc_id: 6, doc_id: 7, doc_id: 8, doc_id: 9, ";
```
aprudhomme, line 293: Is this really the logging behavior you want? The response hits start with 5, but the logging hits start with 4. Should this be logging docs 5-11?

If `hitsToLog` is expected to be relative to `startHit`, it will need to have `startHit` added when being compared to `topHits` (since topHits is currently defined as startHit + hits_to_retrieve).

aprudhomme (approved):

PR #790, opened 2024-11-18
Add default terminate after max recall count

swethakann:


aprudhomme:
I think this is missing the usage of the setting.

aprudhomme (approved):

PR #792, opened 2024-12-04
update README

waziqi89:
RP-11906
Update v0.x README with the latest examples and rest commands


vim345 (commented):
I think Sarthak or asp would know more about this change. The main branch for this cluster seems to have up to date info. I'm not too sure how important it is to update the readme for the v0.x branch.

PR #795, opened 2024-12-06
Backward compatibility of #794

fragosoluana:
Also support https://github.com/Yelp/nrtsearch/pull/794 in 0.x versions


clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -766,6 +766,7 @@ message SearchResponse {
         double rescoreTimeMs = 10;
         map<string, double> rescorersTimeMs = 11;
         map<string, Diagnostics> innerHitsDiagnostics = 12;
+        double loggingHitsTimeMs = 13;
```
aprudhomme, line 769: This needs to use the same field number as the main branch, to keep the clientlib compatible

fragosoluana: thanks for pointing this out!

aprudhomme (approved):

PR #799, opened 2025-01-02
fvh adaptor to highlight the top boost phrase only

waziqi89:
RP-12281
Add `top_boost_only` toggle to highlight the found highest boosted phrase. This is achieved by adding an adaptor to the FragmentBuilder.

The test case is self-explanatory. Only the multiple occurances of the best match will be highlighted.


waziqi89:
```"This pizza place was one of my favorites in downtown, and as promised by other users, the <em>pepperoni pizza</em> is the best pepperoni pizza in town."```
when `pizza` is boost 2, `pepperoni pizza` is boost 3

File: src/main/java/com/yelp/nrtsearch/server/luceneserver/highlights/TopBoostOnlyFragmentsBuilderAdaptor.java [hunk 1]
```diff
@@ -0,0 +1,89 @@
...
+        getFragmentSourceMSO(
+            buffer, index, values, s, fragInfo.getEndOffset(), modifiedStartOffset);
+    int srcIndex = 0;
+    double topBoostValue =
+        fragInfo.getSubInfos().stream().map(SubInfo::getBoost).max(Float::compare).orElse(0f);
+    for (SubInfo subInfo : fragInfo.getSubInfos()) {
+      if (subInfo.getBoost() < topBoostValue) {
+        continue;
+      }
+      for (Toffs to : subInfo.getTermsOffsets()) {
```
waziqi89, line 71: Those are the only added lines from the baseFragmentsBuilder's default method.

src/main/java/com/yelp/nrtsearch/server/luceneserver/highlights/TopBoostOnlyFragmentsBuilderAdaptor.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+    StringBuilder fragment = new StringBuilder();
+    final int s = fragInfo.getStartOffset();
+    int[] modifiedStartOffset = {s};
+    String src =
+        getFragmentSourceMSO(
+            buffer, index, values, s, fragInfo.getEndOffset(), modifiedStartOffset);
+    int srcIndex = 0;
+    double topBoostValue =
+        fragInfo.getSubInfos().stream().map(SubInfo::getBoost).max(Float::compare).orElse(0f);
```
sarthakn7, line 66: Please use for loop instead of stream to avoid extra object creation. This code is called per document, so it can be called 1000s of times per second and we don't want it to be slow.
You can actually also store the subInfos with the highest boost along with the topBoostValue, and then you won't need to iterate over all of the subInfos again.

src/main/java/com/yelp/nrtsearch/server/luceneserver/highlights/TopBoostOnlyFragmentsBuilderAdaptor.java
```diff
@@ -0,0 +1,89 @@
...
+    for (SubInfo subInfo : fragInfo.getSubInfos()) {
+      if (subInfo.getBoost() < topBoostValue) {
+        continue;
+      }
```
sarthakn7, line 70: Does the `modifiedStartOffset` still track the correct offset even if some of the subInfos are skipped?

src/main/java/com/yelp/nrtsearch/server/luceneserver/highlights/TopBoostOnlyFragmentsBuilderAdaptor.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.List;
+import org.apache.lucene.document.Field;
+import org.apache.lucene.search.highlight.Encoder;
+import org.apache.lucene.search.vectorhighlight.BaseFragmentsBuilder;
+import org.apache.lucene.search.vectorhighlight.BoundaryScanner;
+import org.apache.lucene.search.vectorhighlight.FieldFragList.WeightedFragInfo;
+import org.apache.lucene.search.vectorhighlight.FieldFragList.WeightedFragInfo.SubInfo;
+import org.apache.lucene.search.vectorhighlight.FieldPhraseList.WeightedPhraseInfo.Toffs;
+
+public class TopBoostOnlyFragmentsBuilderAdaptor extends BaseFragmentsBuilder {
```
sarthakn7, line 27: Please add some javadocs to the class and to the makeFragment method

sarthakn7 (commented):
Can you please update the highlighting doc and also first send the PR in main branch before backporting to v0? This would ensure that the proto field numbers are correct.

src/main/java/com/yelp/nrtsearch/server/luceneserver/highlights/TopBoostOnlyFragmentsBuilderAdaptor.java
waziqi89: Yes.
This method uses pointer to concatenate the "highlighted" and the other parts. Skipping some of the phrases will still create a complete fragment.

src/main/java/com/yelp/nrtsearch/server/luceneserver/highlights/TopBoostOnlyFragmentsBuilderAdaptor.java
waziqi89: This is called once for each fragment-to-be-created. We have to calculated the topBoostValue for each fragment as not all fragments contains all the desired phrases.

sarthakn7 (approved):

PR #816, opened 2025-01-29
Add support for index_prefix

manav113:
The PR adds support for index_prefixes mapping. 

What's changed:
1. Added a message for indexPrefixes which has min_chars and max_chars property
2. Added PrefixFieldDef which is responsible for prefix indexing and searching.
3. Added PrefixWrappedAnalyzer which wraps the EdgeNGramTokenFilter to generate the prefixes
4. Changed TextFieldDef to incorporate the new field
5. getPrefixQuery is changed to incorporate the prefix query search logic in case index_prefixes is used


src/main/java/com/yelp/nrtsearch/server/field/TextFieldDef.java
```diff
@@ -15,15 +15,47 @@
...
+
   public TextFieldDef(
       String name, Field requestField, FieldDefCreator.FieldDefCreatorContext context) {
     super(name, requestField, context);
+    if (requestField.hasIndexPrefixes()) {
+      if (!requestField.getSearch()) {
+        throw new IllegalArgumentException(
+            "Cannot set index_prefixes on unindexed field [" + name + "]");
+      }
+      this.prefixFieldDef = new PrefixFieldDef(getName(), requestField, context);
```
aprudhomme, line 46: Instead of passing `requestField`, it might be better to use it to build a simplified `Field` with just the properties you want.

src/main/java/com/yelp/nrtsearch/server/query/QueryNodeMapper.java
```diff
@@ -580,9 +576,12 @@ private static Query getPrefixQuery(PrefixQuery prefixQuery, IndexState state) {
       throw new IllegalArgumentException(
           "Field \"" + prefixQuery.getField() + "\" is not indexed with terms");
     }
-
     MultiTermQuery.RewriteMethod rewriteMethod =
         getRewriteMethod(prefixQuery.getRewrite(), prefixQuery.getRewriteTopTermsSize());
+    if (fieldDef instanceof PrefixQueryable prefixQueryable) {
+      return prefixQueryable.getPrefixQuery(prefixQuery, rewriteMethod);
+    }
+
```
aprudhomme, line 584: I would suggest rewriting this method to only use the `PrefixQueryable` interface. This means that `AtomFieldDef` would also need to implement it.

src/test/java/com/yelp/nrtsearch/server/field/PrefixFieldDefTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+    PrefixQuery prefixQuery =
+        PrefixQuery.newBuilder().setField("test_field").setPrefix("test").build();
+    Query query = fieldDef.getPrefixQuery(prefixQuery, null);
+    assertNotNull(query);
+    assertTrue(query instanceof ConstantScoreQuery);
+  }
+
+  @Test
+  public void testPrefixQueryBooleanQuery() {
```
aprudhomme, line 161: It would be good to also verify the queries built at the `TextFieldDef` level for all different possibilities. i.e., << min, min - 1, between min and max, > max

src/main/java/com/yelp/nrtsearch/server/field/PrefixFieldDef.java
```diff
@@ -0,0 +1,156 @@
...
+    fieldType.setOmitNorms(true);
+    fieldType.setIndexOptions(IndexOptions.DOCS);
+    fieldType.setTokenized(true);
+  }
+
+  protected Analyzer parseIndexAnalyzer(Field requestField) {
+    Analyzer baseAnalyzer = super.parseIndexAnalyzer(requestField);
+    if (baseAnalyzer == null) {
+      return null;
+    }
```
aprudhomme, line 78: I think it would be good to throw an exception if there is no analyzer

src/main/java/com/yelp/nrtsearch/server/field/PrefixFieldDef.java
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  @Override
+  protected void setSearchProperties(FieldType fieldType, Field requestField) {
+    fieldType.setOmitNorms(true);
+    fieldType.setIndexOptions(IndexOptions.DOCS);
+    fieldType.setTokenized(true);
+  }
+
+  protected Analyzer parseIndexAnalyzer(Field requestField) {
```
aprudhomme, line 74: This method should be flagged with `@Override`

aprudhomme (commented):
This is most the way there

src/main/java/com/yelp/nrtsearch/server/analysis/PrefixWrappedAnalyzer.java
```diff
@@ -0,0 +1,47 @@
...
+ * limitations under the License.
+ */
+package com.yelp.nrtsearch.server.analysis;
+
+import org.apache.lucene.analysis.Analyzer;
+import org.apache.lucene.analysis.AnalyzerWrapper;
+import org.apache.lucene.analysis.TokenFilter;
+import org.apache.lucene.analysis.ngram.EdgeNGramTokenFilter;
+
+public class PrefixWrappedAnalyzer extends AnalyzerWrapper {
```
aprudhomme, line 23: Can you add doc strings for this class

src/main/java/com/yelp/nrtsearch/server/query/QueryNodeMapper.java
```diff
@@ -580,11 +576,15 @@ private static Query getPrefixQuery(PrefixQuery prefixQuery, IndexState state) {
       throw new IllegalArgumentException(
           "Field \"" + prefixQuery.getField() + "\" is not indexed with terms");
     }
```
aprudhomme, line 578: You can get rid of the checks above, as they are not needed when using the new interface. Any needed checks can be done in the interface implementation

src/main/java/com/yelp/nrtsearch/server/field/AtomFieldDef.java
```diff
@@ -150,4 +150,10 @@ public Query getRangeQuery(RangeQuery rangeQuery) {
           "Only SORTED or SORTED_SET doc values are supported for range queries: " + getName());
     }
   }
+
+  @Override
+  public Query getPrefixQuery(PrefixQuery prefixQuery, MultiTermQuery.RewriteMethod rewriteMethod) {
+    return new org.apache.lucene.search.PrefixQuery(
+        new Term(prefixQuery.getField(), prefixQuery.getPrefix()), rewriteMethod);
+  }
```
aprudhomme, line 158: You can add a call to `self.verifySearchable()` to replace the check in the query node mapper

src/main/java/com/yelp/nrtsearch/server/field/TextFieldDef.java
```diff
@@ -15,15 +15,72 @@
...
+  private static final int DEFAULT_MAX_CHARS = 5;
+
   public TextFieldDef(
       String name, Field requestField, FieldDefCreator.FieldDefCreatorContext context) {
     super(name, requestField, context);
+    if (requestField.hasIndexPrefixes()) {
+      if (!requestField.getSearch()) {
+        throw new IllegalArgumentException(
+            "Cannot set index_prefixes on unindexed field [" + name + "]");
+      }
```
aprudhomme, line 48: This could also be replaced with `verifySearchable`, for uniformity

src/main/java/com/yelp/nrtsearch/server/field/TextFieldDef.java
```diff
@@ -51,4 +108,50 @@ protected void setSearchProperties(FieldType fieldType, Field requestField) {
...
+    super.parseDocumentField(document, fieldValues, facetHierarchyPaths);
+
+    if (hasPrefix() && !fieldValues.isEmpty()) {
+      prefixFieldDef.parseDocumentField(document, fieldValues, facetHierarchyPaths);
+    }
+  }
+
+  @Override
+  public Query getPrefixQuery(PrefixQuery prefixQuery, MultiTermQuery.RewriteMethod rewriteMethod) {
+    if (prefixFieldDef != null && prefixFieldDef.accept(prefixQuery.getPrefix().length())) {
```
aprudhomme, line 132: `prefixFieldDef != null` could be replaced with `hasPrefix`

aprudhomme, line 132: `verifySearchable` would also go here

src/main/java/com/yelp/nrtsearch/server/field/TextFieldDef.java
```diff
@@ -51,4 +108,50 @@ protected void setSearchProperties(FieldType fieldType, Field requestField) {
...
+  }
+
+  public void validatePrefix(int minChars, int maxChars) {
+    if (minChars > maxChars) {
+      throw new IllegalArgumentException(
+          "min_chars [" + minChars + "] must be less than max_chars [" + maxChars + "]");
+    }
+    if (minChars < 2) {
+      throw new IllegalArgumentException("min_chars [" + minChars + "] must be greater than zero");
+    }
```
aprudhomme, line 152: Should the condition be `minChar <= 0`?

src/main/java/com/yelp/nrtsearch/server/field/TextFieldDef.java
```diff
@@ -15,15 +15,72 @@
...
+          new PrefixFieldDef(
+              getName(),
+              Field.newBuilder()
+                  .setSearch(true)
+                  .setIndexPrefixes(
+                      IndexPrefixes.newBuilder()
+                          .setMinChars(minChars)
+                          .setMaxChars(maxChars)
+                          .build())
+                  .build(),
```
aprudhomme, line 69: This field will also need the index analyzer defined in the parent field. Check if the parent has `analyzer` or `indexAnalyzer` and set them as needed

src/main/java/com/yelp/nrtsearch/server/field/properties/PrefixQueryable.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.yelp.nrtsearch.server.field.properties;
+
+import com.yelp.nrtsearch.server.grpc.PrefixQuery;
+import org.apache.lucene.search.MultiTermQuery;
+import org.apache.lucene.search.Query;
+
+public interface PrefixQueryable {
```
aprudhomme, line 22: Could you add doc strings for this interface

src/main/java/com/yelp/nrtsearch/server/field/PrefixFieldDef.java
```diff
@@ -0,0 +1,108 @@
...
+    this.parentField = parentName;
+  }
+
+  @Override
+  protected void setSearchProperties(FieldType fieldType, Field requestField) {
+    fieldType.setOmitNorms(true);
+    fieldType.setTokenized(true);
+    if (requestField.getSearch()) {
+      setIndexOptions(requestField.getIndexOptions(), fieldType, IndexOptions.DOCS);
+    }
```
aprudhomme, line 53: You know this field will have `search=true` and no `IndexOptions` set, so you can probably just set the value to `DOCS`

src/main/java/com/yelp/nrtsearch/server/field/TextFieldDef.java
```diff
@@ -15,15 +15,71 @@
...
+              : DEFAULT_MAX_CHARS;
+      validatePrefix(minChars, maxChars);
+
+      this.prefixFieldDef =
+          new PrefixFieldDef(
+              getName(),
+              Field.newBuilder()
+                  .setSearch(true)
+                  .setAnalyzer(requestField.getAnalyzer())
+                  .setIndexAnalyzer(requestField.getIndexAnalyzer())
```
aprudhomme, line 62: What happens if the analyzers in the `requestField` are unset, will the new field values be counted as set? If so, it would need to use the `has` methods to only set the analyzers if they exist in the request.

aprudhomme (dismissed):

aprudhomme (approved):

PR #821, opened 2025-02-06
U/raviraj/partial update

RavirajDataEngineering:
POC for partial update of RTB fields


src/main/java/com/yelp/nrtsearch/server/handler/AddDocumentHandler.java
```diff
@@ -400,15 +411,40 @@ public long runIndexingJob() throws Exception {
       IndexState indexState;
       ShardState shardState;
       IdFieldDef idFieldDef;
-
+      String ad_bid_id = "";
+      Set<String> partialUpdateFields =
+          Set.of("ad_bid_floor", "ad_bid_floor_USD", "ad_bid_value", "ad_bid_value_USD");
```
aprudhomme, line 416: I see the schema has these as multi valued fields. The lucene dv update api only supports single value numerics and binary types (as far as I can tell). You could try confirming this with tests.

src/main/java/com/yelp/nrtsearch/server/handler/AddDocumentHandler.java
```diff
@@ -106,6 +111,7 @@ private void incrementCount(String indexName) {
       @Override
       public void onNext(AddDocumentRequest addDocumentRequest) {
         String indexName = addDocumentRequest.getIndexName();
+        boolean isPartialUpdateRequest = isPartialUpdate(addDocumentRequest);
```
aprudhomme, line 114: Is this used for anything?

src/main/java/com/yelp/nrtsearch/server/handler/AddDocumentHandler.java
RavirajDataEngineering: I think you are right. I just set the `multivalued: true` on the test and it fails.

src/main/java/com/yelp/nrtsearch/server/handler/AddDocumentHandler.java
RavirajDataEngineering: Thank you. No its not used

PR #823, opened 2025-02-09
[Backport] Added files for implementing index_prefix

manav113:
Backport of [816]( https://github.com/Yelp/nrtsearch/pull/816)


src/main/java/com/yelp/nrtsearch/server/luceneserver/field/AtomFieldDef.java
```diff
@@ -145,4 +150,22 @@ public SortField getSortField(SortType type) {
...
+    query.setRewriteMethod(rewriteMethod);
+    return query;
+  }
+
+  protected void verifySearchable(String featureName) {
+    if (!isSearchable()) {
+      throw new IllegalStateException(
+          featureName + " requires field to be searchable: " + getName());
+    }
+  }
```
aprudhomme, line 170: It might be better to add this to `IndexableFieldDef` instead of duplicating it in each field

aprudhomme:
Looks like there is a spotless failure

aprudhomme (approved):

PR #826, opened 2025-02-13
Trigger log with no hits

fragosoluana:
Logging with no hits can still be useful, so let's not skip it and let plugin implementation to decide what to log or not in this situation


src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
```diff
@@ -387,6 +388,12 @@ public SearchResponse handle(IndexState indexState, SearchRequest searchRequest)
   private void fetchFields(SearchContext searchContext)
       throws IOException, ExecutionException, InterruptedException {
     if (searchContext.getResponseBuilder().getHitsBuilderList().isEmpty()) {
+      if (searchContext.getFetchTasks().getHitsLoggerFetchTask() != null) {
+        searchContext
```
taoyyu, line 392: can you leave a comment here that we may want to log request even if there is no hits t o return

taoyyu (dismissed):

taoyyu (approved):

PR #831, opened 2025-02-21
bootstrap metrics for v1

waziqi89:
RP-12622
Add bootstrap metrics for v1. The timer try-with-resource is recommended by prometheus timerApi docstring

Tested with modified configs from nova talk_v2.
Will adjust v0 PR accordingly

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,7 +26,7 @@ java {
 }
 
 allprojects {
-    version = '1.0.0-beta.4'
+    version = '1.0.0-beta.5'
     group = 'com.yelp.nrtsearch'
 }
 
--- a/src/main/java/com/yelp/nrtsearch/server/grpc/NrtsearchServer.java
+++ b/src/main/java/com/yelp/nrtsearch/server/grpc/NrtsearchServer.java
@@ -111,7 +111,6 @@ import io.grpc.ServerInterceptors;
 import io.grpc.netty.shaded.io.grpc.netty.NettyServerBuilder;
 import io.grpc.protobuf.services.ProtoReflectionService;
 import io.grpc.stub.StreamObserver;
-import io.prometheus.metrics.core.datapoints.Timer;
 import io.prometheus.metrics.instrumentation.jvm.JvmMetrics;
 import io.prometheus.metrics.model.registry.PrometheusRegistry;
 import java.io.File;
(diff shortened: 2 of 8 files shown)


src/main/java/com/yelp/nrtsearch/server/grpc/NrtsearchServer.java
```diff
@@ -148,74 +150,77 @@ public NrtsearchServer(
...
-    } else {
-      replicationServer =
-          ServerBuilder.forPort(configuration.getReplicationPort())
-              .addService(
-                  new ReplicationServerImpl(
-                      globalState, configuration.getVerifyReplicationIndexId()))
-              .executor(
-                  ExecutorFactory.getInstance()
-                      .getExecutor(ExecutorFactory.ExecutorType.REPLICATIONSERVER))
+    try (Timer _timer = BootstrapMetrics.nrtsearchBootstrapTimer.startTimer()) {
```
sarthakn7, line 153: Let's just get the time manually to avoid multiple levels of indentation. It's also useful if we need to log the time.

src/main/java/com/yelp/nrtsearch/server/nrt/NrtDataManager.java
```diff
@@ -178,27 +182,33 @@ public void restoreIfNeeded(Path shardDataDir) throws IOException {
     if (restoreIndex.getDeleteExistingData()) {
       FileUtils.deleteAllFilesInDir(shardDataDir);
     }
+    try (Timer _timer =
```
sarthakn7, line 185: Here the time is being calculated manually anyway, can re-use that.

src/main/java/com/yelp/nrtsearch/server/state/BackendGlobalState.java
```diff
@@ -96,6 +97,18 @@ public static String getUniqueIndexName(String indexName, String id) {
...
+   * Return the base index name from the unique index identifier.
+   *
+   * @param uniqueIndexName index identifier
+   * @return indexName index name
+   * @throws NullPointerException if uniqueIndexName is null
+   */
+  public static String getBaseIndexName(String uniqueIndexName) {
+    Objects.requireNonNull(uniqueIndexName);
+    return StringUtils.substringBefore(uniqueIndexName, "-");
+  }
```
sarthakn7, line 110: This won't work correctly if an index name as "-" in the name. Even if we don't have indices right now like that, still doesn't seem right to have to avoid "-" in name. I think it's okay to just add the index identifier instead of index name in the labels to avoid this. Or alternatively, add indexName as a parameter to the method.

build.gradle
```diff
@@ -26,7 +26,7 @@ java {
 }
 
 allprojects {
-    version = '1.0.0-beta.4'
+    version = '1.0.0-beta.5'
```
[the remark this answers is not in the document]
waziqi89: Shall we do a quick test on talk-v2 after merge? I think we shouldn't leave the code untested in prod for a long time.

src/main/java/com/yelp/nrtsearch/server/state/BackendGlobalState.java
waziqi89: Agree with "-" in index, but based on the original "reverse" method, this shall work if we change to reverse spilt from the last "-" as there's no hyphen in uuid.

https://github.com/Yelp/nrtsearch/blob/2bab53c0bc584ab518a1ed7f16c50770f9e8bed3/src/main/java/com/yelp/nrtsearch/server/state/BackendGlobalState.java#L94

waziqi89: `    // Unique identifier for index (yyyyMMddHHmmssSSS formatted date-time string)`

src/main/java/com/yelp/nrtsearch/server/nrt/NrtDataManager.java
```diff
@@ -65,7 +68,7 @@ public class NrtDataManager implements Closeable {
   private Path shardDataDir;
   private UploadManagerThread uploadManagerThread = null;
 
-  // Set during restoreIfNeeded, then only accessed by UploadManagerThread
+  // Set during restoreIfNeeded, then only accessed by UploadManagerThreadindexIdentifier
```
sarthakn7, line 71: Typo? I believe this was already correct and doesn't require changing

src/main/java/com/yelp/nrtsearch/server/state/BackendGlobalState.java
```diff
@@ -96,6 +97,19 @@ public static String getUniqueIndexName(String indexName, String id) {
...
+  /**
+   * Return the base index name from the unique index identifier.
+   *
+   * @param uniqueIndexName index identifier
+   * @return indexName index name
+   * @throws NullPointerException if uniqueIndexName is null
+   */
+  public static String getBaseIndexName(String uniqueIndexName) {
+    Objects.requireNonNull(uniqueIndexName);
+    // unique identifier doesn't contain "-"
```
sarthakn7, line 109: Not sure what is meant by this comment since the unique index identifier does contain hyphen.

sarthakn7 (commented):
Code lgtm, can you please correct the comments?

src/main/java/com/yelp/nrtsearch/server/state/BackendGlobalState.java
sarthakn7: Ah I guess you meant the yyyyMMdd part. Would be good to clarify here.

sarthakn7 (approved):

PR #836, opened 2025-03-14
Improve field documentation

aprudhomme:
Tried to make some improvements to our field docs by making a new page for each field type. There is more info that can be added, but this is a start.


docs/field_types/numeric.rst
```diff
@@ -0,0 +1,36 @@
+Numeric
+=======
+The field type is used to store numeric values.
```
sarthakn7, line 3:
```suggestion
This field type is used to store numeric values.
```

docs/field_types/id.rst
```diff
@@ -0,0 +1,31 @@
+ID Field
+========
+This field serves as the primary term for the document. All document are required to have a value for this field. When updating a document, the _ID field is used to atomically replace the previous document with the new version. It is highly recommended for all indies to have an _ID field.
```
sarthakn7, line 3: Typo
```suggestion
This field serves as the primary term for the document. All documents are required to have a value for this field. When updating a document, the _ID field is used to atomically replace the previous document with the new version. It is highly recommended for all indices to have an _ID field.
```

sarthakn7: Also slightly confusing - all documents are _required_ to have a value for this field, but it's _recommended_ for all indices to have an _ID field. IMO we can just say that this is required, since it's not possible to update documents without _ID field.
```suggestion
This field serves as the primary term for the document. All indices are required to have an _ID field, and all documents are required to have a value for this field. When updating a document, the _ID field is used to atomically replace the previous document with the new version.
```

sarthakn7 (approved):

PR #837, opened 2025-03-18
Added warming query stripping logic

fragosoluana:


src/main/java/com/yelp/nrtsearch/server/warming/WarmingQueryStripping.java
```diff
@@ -0,0 +1,139 @@
...
+
+        this.random = new Random();
+        this.strippedRescorerCount = 0;
+        this.strippedFunctionScoreScriptCount = 0;
+        this.strippedVirtualFieldsCount = 0;
+        this.strippedFacetsCount = 0;
+    }
+
+    public SearchRequest stripWarmingQuery(SearchRequest.Builder builder, int warmingCount) {
+        int probability = random.nextInt(99);
```
waziqi89, line 45: not sure about your intention, but sharing the same random result may lead the case multiple assertions are mutual inclusion.

waziqi89: You want to guarantee virtualField will be stripped when functionScore is stripped?

src/main/java/com/yelp/nrtsearch/server/warming/WarmingQueryStripping.java
```diff
@@ -0,0 +1,139 @@
...
+
+    private void stripRescorers(SearchRequest.Builder builder) {
+        builder.clearRescorers();
+        // also need to skip fetch tasks as they can depend on rescorers
+        builder.clearFetchTasks();
+        strippedRescorerCount++;
+    }
+
+    private void stripFunctionScoreScript(SearchRequest.Builder builder) {
+        if (builder.hasQuery() && builder.getQuery().hasFunctionScoreQuery()) {
```
waziqi89, line 95: `hasFunctionScoreQuery` looks too magical to me. Didn't look into details, but I am afraid this is a method provided by protoc, which only checks at the top level. It will miss the case when the functionScoreQuery is wrapped by booleanQuery

src/main/java/com/yelp/nrtsearch/server/warming/WarmingQueryStripping.java
```diff
@@ -0,0 +1,139 @@
...
+    private int strippedFacetsCount;
+
+    private int strippedQueryCount;
+
+    public WarmingQueryStripping(
+            int maxRescorerStrippingPerc,
+            int maxFunctionScoreScriptStrippingPerc,
+            int maxVirtualFieldsStrippingPerc,
+            int maxFacetsStrippingPerc) {
+        this.maxRescorerStrippingPerc = maxRescorerStrippingPerc;
```
waziqi89, line 32: Wonder what's the main purpose of separating the stripping into multiple subtasks? If it's used to analyze the different impact on each task, I get it. But in the "real" PR, I think it makes sense to "strip all when < maxStrippedWarmingQueriesPerc"

src/main/java/com/yelp/nrtsearch/server/warming/WarmingQueryStripping.java
fragosoluana: My understanding is that we don’t know exactly how many queries will be backed up—we only define a maximum using `maxWarmingQueries`. Because of this, I wanted to introduce a percentage-based configuration (e.g., 50% of queries should skip "rescorer") instead of relying on a fixed count, since we don’t have an exact number of backed-up queries.

Currently, we read the warming queries [line by line ](https://sourcegraph.yelpcorp.com/Yelp/nrtsearch/-/blob/src/main/java/com/yelp/nrtsearch/server/warming/Warmer.java?L144), which means we would need to read the entire file to determine the exact number of queries. One option would be to load the whole file at once to count the lines, but I assumed the current approach was chosen to optimize memory usage.

To address this, I implemented a solution where queries are stripped probabilistically based on the new maxPerc configurations, maintaining the existing readLine(). Let me know your thoughts.

src/main/java/com/yelp/nrtsearch/server/warming/WarmingQueryStripping.java
fragosoluana: Oh true, I need to cover the other use cases. I will double check if this will get too complex. If so, I could skip this for now.

src/main/java/com/yelp/nrtsearch/server/warming/WarmingQueryStripping.java
waziqi89: probability stripping makes sense.
I am more curious about the logic on using the same `int probability ` for multiple assertions.
For example, if we have virtualField 50% functionScore 70%. Then if the probability is < 50, then it's guaranteed < 70. In other words, the query which strips functionScore always strips virtualField. Those ain't independent event.

src/main/java/com/yelp/nrtsearch/server/warming/WarmingQueryStripping.java
waziqi89: I think we can recursively check all boolean clauses (neglect other possibility for now) and replace the functionScoreQuery's script into the dummy one `{lang": "js", "source": "1.0"}`

src/main/java/com/yelp/nrtsearch/server/warming/WarmingQueryStripping.java
fragosoluana: oh got it, yeah, I agree with you. I will move it into the `shouldStripQuery` method so it is generated for every check and avoid sharing the probability

src/main/java/com/yelp/nrtsearch/server/warming/WarmingQueryStripping.java
fragosoluana: Yeah, mainly to first experiment and understand which stripping logic can actually help; but also, I believe that each use case might need different query stripping.

fragosoluana:
I will reopen another PR to properly branch out from profiling branch

waziqi89:
> I will reopen another PR to properly branch out from profiling branch

Pleas don't branch out from profiling itself. It is behind the latest v0. You may want to create a "profiling_backup" branch at current head, and just force push to update the profiling

PR #839, opened 2025-03-25
Add query per stage latency metrics

mathisnyp:
Working on publishing per stage latency metrics via prometheus.


src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
```diff
@@ -328,7 +328,8 @@ public SearchResponse handle(IndexState indexState, SearchRequest searchRequest)
                         InnerHitFetchTask::getDiagnostic)));
       }
       searchContext.getResponseBuilder().setDiagnostics(diagnostics);
-
+      // TODO: These are the diagnostics I want to publish to prometheus, I'll try to figure out
```
mathisnyp, line 331: I just left these comments here as a reference for myself.

src/main/java/com/yelp/nrtsearch/server/monitoring/SearchResponseCollector.java
```diff
@@ -120,6 +120,7 @@ public static void updateSearchResponseMetrics(
             .labelValues(index, "facet:" + entry.getKey())
             .observe(entry.getValue());
       }
+      searchStageLatencyMs.labelValues(index, "rescore").observe(diagnostics.getRescoreTimeMs());//adding extra rescore metric to avoid calculating average of all rescorers
       for (Map.Entry<String, Double> entry : diagnostics.getRescorersTimeMsMap().entrySet()) {
```
mathisnyp, line 124: For now searchStageLatencyMs only has the rescorer latency for each rescorer, but for an initial overview a general latency for all rescorers might also be useful.

src/main/java/com/yelp/nrtsearch/server/monitoring/SearchResponseCollector.java
```diff
@@ -151,7 +155,6 @@ public MetricSnapshots collect() {
       if (publishVerboseMetrics) {
```
mathisnyp, line 155: When would we set this to true?

src/main/java/com/yelp/nrtsearch/server/monitoring/SearchResponseCollector.java
```diff
@@ -139,6 +140,9 @@ public MetricSnapshots collect() {
     try {
       metrics.add(searchTimeoutCount.collect());
       metrics.add(searchTerminatedEarlyCount.collect());
+      metrics.add(searchStageLatencyMs.collect());// Just adding this here should mean it gets published to prometheus, is that what we want?
+      // when is publishVerboseMetrics set to true, I couldn't find this metric in the grafana shard without any filters?
+      // maybe it makes sense to add an extra parameter like publishSearchStageLatencyMs with default value true (?)
```
mathisnyp, line 145: Does it make sense to have a way of turning this on and off separately?

src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
```diff
@@ -328,7 +328,8 @@ public SearchResponse handle(IndexState indexState, SearchRequest searchRequest)
                         InnerHitFetchTask::getDiagnostic)));
       }
       searchContext.getResponseBuilder().setDiagnostics(diagnostics);
```
sarthakn7, line 330: I think the diagnostics object here contains all of the required metrics, including the rescorer metrics. Can you please confirm once?

src/main/java/com/yelp/nrtsearch/server/monitoring/SearchResponseCollector.java
sarthakn7: There is a live index setting which enables publishing verbose metrics

sarthakn7: I think if a metric is per-hit, it should come under the verbose metrics

src/main/java/com/yelp/nrtsearch/server/monitoring/SearchResponseCollector.java
sarthakn7: Only when we need to the extra metrics to investigate something

PR #842, opened 2025-03-31
Support null for sort field values

aprudhomme:
For certain field types (like `TEXT`), sort values will be `null` when a document has no value. This causes issues for our sort processing.

For the `LastHitInfo`, the gRPC interface does not make it easy to express `null` for sort field values. I propose using the string `NULL_SORT_VALUE`. This seems specific enough to avoid collision with actual data.

When populating the sort values in each `Hit`, an empty `CompositeFieldValue` is used. This is what you would see if you tried to retrieve the field.

The `searchAfter` parsing has been updated to accept the `NULL_SORT_VALUE` string value.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/search/sort/SortParser.java
+++ b/src/main/java/com/yelp/nrtsearch/server/search/sort/SortParser.java
@@ -39,7 +39,7 @@ import org.apache.lucene.util.BytesRef;
 public class SortParser {
   public static final BiFunction<SortField, Object, CompositeFieldValue>
       DEFAULT_SORT_VALUE_EXTRACTOR = (sortField, value) -> getValueForSortField(sortField, value);
-  public static final String NULL_SORT_VALUE = "NULL_SORT_VALUE";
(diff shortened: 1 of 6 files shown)


taoyyu (dismissed):

src/main/java/com/yelp/nrtsearch/server/search/sort/SortParser.java
```diff
@@ -39,6 +39,7 @@
 public class SortParser {
   public static final BiFunction<SortField, Object, CompositeFieldValue>
       DEFAULT_SORT_VALUE_EXTRACTOR = (sortField, value) -> getValueForSortField(sortField, value);
+  public static final String NULL_SORT_VALUE = "NULL_SORT_VALUE";
```
sarthakn7, line 42: Should we document this somewhere? Maybe under a section called "Reserved Values"?
Also should we make it more unlikely to be used, maybe wrap in %? So `%NULL_SORT_VALUE%`?

sarthakn7 (approved):

PR #843, opened 2025-04-01
add new params to warm lucene query only

waziqi89:
add `strippedWarmingQueryRate` settings to strip the warming query.

Derive from Luana's testing code. Tested with lucy and ads configs locally.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/luceneserver/SearchHandler.java
+++ b/src/main/java/com/yelp/nrtsearch/server/luceneserver/SearchHandler.java
@@ -323,48 +323,6 @@ public class SearchHandler implements Handler<SearchRequest, SearchResponse> {
     return searchResponse;
   }
 
-  public void handleStrippedWarmingQuery(IndexState indexState, SearchRequest searchRequest) {
-    SearchContext searchContext;
-    SearcherTaxonomyManager.SearcherAndTaxonomy s = null;
-    ShardState shardState = indexState.getShard(0);
-
-    try {
-      s =
-          getSearcherAndTaxonomy(
-              searchRequest,
-              indexState,
-              shardState,
-              SearchResponse.Diagnostics.newBuilder(),
-              indexState.getSearchThreadPoolExecutor());
-
-      searchContext =
-          SearchRequestProcessor.buildContextForWarmingQuery(
-              searchRequest, indexState, shardState, s);
-
-      s.searcher.search(searchContext.getQuery(), searchContext.getCollector().getWrappedManager());
-    } catch (InterruptedException | IOException e) {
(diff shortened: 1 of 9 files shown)


src/main/java/com/yelp/nrtsearch/server/luceneserver/warming/WarmerConfig.java
```diff
@@ -47,8 +54,12 @@ public static WarmerConfig fromConfig(YamlConfigReader configReader) {
         configReader.getInteger(CONFIG_PREFIX + "warmingParallelism", DEFAULT_WARMING_PARALLELISM);
     boolean warmOnStartup =
         configReader.getBoolean(CONFIG_PREFIX + "warmOnStartup", DEFAULT_WARM_ON_STARTUP);
+    float strippedWarmingQueryRate =
+        configReader.getFloat(
+            CONFIG_PREFIX + "strippedWarmingQueryRate", DEFAULT_STRIPPED_WARMING_QUERY_RATE);
```
sarthakn7, line 59: I suggest using `warmBasicQueryOnlyPerc` name for this parameter

src/main/java/com/yelp/nrtsearch/server/luceneserver/SearchHandler.java
```diff
@@ -323,6 +323,48 @@ public SearchResponse handle(IndexState indexState, SearchRequest searchRequest)
     return searchResponse;
   }
 
+  public void handleStrippedWarmingQuery(IndexState indexState, SearchRequest searchRequest) {
```
sarthakn7, line 326: It seems like this function is unused, and only `WarmingUtils.handleStrippedWarmingQuery` is called

src/main/java/com/yelp/nrtsearch/server/luceneserver/warming/WarmingUtils.java
```diff
@@ -0,0 +1,146 @@
...
+import com.yelp.nrtsearch.server.luceneserver.search.SearchRequestProcessor;
+import java.io.IOException;
+import org.apache.lucene.facet.taxonomy.SearcherTaxonomyManager;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class WarmingUtils {
+  private static final Logger logger = LoggerFactory.getLogger(WarmingUtils.class);
+
+  public static void handleStrippedWarmingQuery(
```
sarthakn7, line 40: Why do we need the stripped query to go through a different flow? If there is anything else remaining in the query that must be stripped, why can't we just remove those things and go through the same flow?

sarthakn7 (approved):

PR #844, opened 2025-04-08
add basic warming query configs

waziqi89:
port the v0 stripped warming query code into v1.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/warming/WarmingUtils.java
+++ b/src/main/java/com/yelp/nrtsearch/server/warming/WarmingUtils.java
@@ -23,11 +23,8 @@ import com.yelp.nrtsearch.server.grpc.MultiFunctionScoreQuery;
 import com.yelp.nrtsearch.server.grpc.NestedQuery;
 import com.yelp.nrtsearch.server.grpc.Query;
 import com.yelp.nrtsearch.server.grpc.SearchRequest;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 public class WarmingUtils {
-  private static final Logger logger = LoggerFactory.getLogger(WarmingUtils.class);
 
   public static SearchRequest simplifySearchRequestForWarming(
       SearchRequest.Builder searchRequestBuilder) {
--- a/src/main/java/com/yelp/nrtsearch/server/warming/WarmingUtils.java
+++ b/src/main/java/com/yelp/nrtsearch/server/warming/WarmingUtils.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2020 Yelp Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.yelp.nrtsearch.server.warming;
+
+import com.yelp.nrtsearch.server.grpc.BooleanClause;
+import com.yelp.nrtsearch.server.grpc.BooleanQuery;
+import com.yelp.nrtsearch.server.grpc.ConstantScoreQuery;
+import com.yelp.nrtsearch.server.grpc.DisjunctionMaxQuery;
+import com.yelp.nrtsearch.server.grpc.MultiFunctionScoreQuery;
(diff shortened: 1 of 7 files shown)


src/main/java/com/yelp/nrtsearch/server/warming/WarmingUtils.java
```diff
@@ -0,0 +1,115 @@
...
+import com.yelp.nrtsearch.server.grpc.DisjunctionMaxQuery;
+import com.yelp.nrtsearch.server.grpc.MultiFunctionScoreQuery;
+import com.yelp.nrtsearch.server.grpc.NestedQuery;
+import com.yelp.nrtsearch.server.grpc.Query;
+import com.yelp.nrtsearch.server.grpc.SearchRequest;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class WarmingUtils {
+  private static final Logger logger = LoggerFactory.getLogger(WarmingUtils.class);
```
fragosoluana, line 30: I think this is not being used

fragosoluana (commented):
I noticed that [WarmerTest](https://github.com/Yelp/nrtsearch/pull/843/files#diff-af43159f6b229d6580f0191e8e4a722b84e3c949af88b2b9b8f2bfea314bfd13) changes from v0.x weren't added in this PR, just double checking if this is intentional

src/main/java/com/yelp/nrtsearch/server/warming/WarmingUtils.java
waziqi89: Intentionally leave it here in case we want to log something later. This exists in many other classes without reference.

src/test/java/com/yelp/nrtsearch/server/warming/WarmerTest.java
```diff
@@ -167,17 +191,47 @@ private List<SearchRequest> getTestSearchRequests() {
...
+  private List<SearchRequest> getTestBasicSearchRequests() {
+    List<SearchRequest> testRequests = new ArrayList<>();
+    SearchRequest searchRequest =
+        SearchRequest.newBuilder()
+            .setIndexName(index)
+            .setQuery(Query.newBuilder().setTermQuery(TermQuery.newBuilder().setField("field0")))
+            .build();
+    testRequests.add(searchRequest);
+
+    searchRequest =
```
fragosoluana, line 215: should we add a comment just to make it clear that we are expecting the second request to not be warmed up due to the random seed? When I quickly read the test, I expected this request to not have a script query since the name of the method was getTest`Basic`SearchRequests

src/main/java/com/yelp/nrtsearch/server/warming/WarmingUtils.java
sarthakn7: > This exists in many other classes without reference.

It's only there because people forget to remove it. It's just one line, easy to add back if needed in the future.

src/test/java/com/yelp/nrtsearch/server/warming/WarmerTest.java
```diff
@@ -161,23 +186,61 @@ private byte[] getWarmingBytes(List<String> queryStrings) throws IOException {
...
+    }
+
+    return testRequests;
+  }
+
   private List<String> getTestSearchRequestsAsJsonStrings() {
     return List.of(
-        "{\"indexName\":\"test_index\",\"query\":{\"termQuery\":{\"field\":\"field0\"}}}",
-        "{\"indexName\":\"test_index\",\"query\":{\"termQuery\":{\"field\":\"field1\"}}}");
+        "{\"indexName\":\"test_index\",\"query\":{\"functionScoreQuery\":{\"query\":{\"termQuery\":{\"field\":\"field0\"}},\"script\":{\"lang\":\"js\",\"source\":\"3 * 5\"}}}}",
```
fragosoluana, line 241: Before we were testing the same query twice to test it being simplified or not. With the last commit change, we have 2 extra identical queries and I am just wondering what is being tested with these extra identical queries. Just wondering if we need them or if we can keep only two and simplify the test code by removing the extra for loops in getTestBasicSearchRequests.

src/test/java/com/yelp/nrtsearch/server/warming/WarmerTest.java
```diff
@@ -205,33 +206,41 @@ private List<SearchRequest> getTestSearchRequests() {
...
+      SearchRequest searchRequest =
+          SearchRequest.newBuilder()
+              .setIndexName(index)
+              .setQuery(
+                  Query.newBuilder().setTermQuery(TermQuery.newBuilder().setField("field" + i)))
+              .build();
+      testRequests.add(searchRequest);
+    }
+
+    for (; i < 4; i++) {
```
fragosoluana, line 220: minor suggestion: use `for (int i = 2; i < 4; i++)` to make it clear that this for loop will add 2 requests. When I first read it quickly, I thought it was adding 4.

fragosoluana (approved):

PR #847, opened 2025-04-14
Implementing document updates with updateDocValues API

RavirajDataEngineering:
This PR :

1. Implements updating document with the `updateDocValues` API.
2. Adds metrics we used to compare the POC performance, these metrics are 
     - updateDocValue qps
     - updteDocValue latency
     - individual addDocument qps
     - individual addDocument latency


src/main/java/com/yelp/nrtsearch/server/handler/AddDocumentHandler.java
```diff
@@ -460,6 +473,58 @@ public long runIndexingJob() throws Exception {
...
+            String idFieldName = indexState.getIdFieldDef().get().getName();
+            String idFieldValue = addDocumentRequest.getFieldsMap().get(idFieldName).getValue(0);
+
+            if (idFieldValue == null || idFieldValue.isEmpty()) {
+              throw new IllegalArgumentException(
+                  String.format("the _ID should have a value set to execute update DocValue"));
+            }
+            term = new Term(idFieldName, idFieldValue);
+            continue;
+          }
```
aprudhomme, line 503: It might help with readability to create a separate method to extract the id term, then skip it when processing the other fields

src/main/java/com/yelp/nrtsearch/server/handler/AddDocumentHandler.java
```diff
@@ -460,6 +473,58 @@ public long runIndexingJob() throws Exception {
...
+            addDocumentRequest.getFieldsMap().entrySet()) {
+          FieldDef field = indexState.getField(entry.getKey());
+          if (field == null) {
+            throw new IllegalArgumentException(
+                String.format("Field: %s is not registered", entry.getKey()));
+          }
+          if (field.getName().equals(indexState.getIdFieldDef().get().getName())) {
+
+            String idFieldName = indexState.getIdFieldDef().get().getName();
+            String idFieldValue = addDocumentRequest.getFieldsMap().get(idFieldName).getValue(0);
```
aprudhomme, line 495: Have we checked that this index has an id field

src/test/java/com/yelp/nrtsearch/server/grpc/UpdateDocValuesTest.java
```diff
@@ -0,0 +1,253 @@
...
+            .withAutoStartConfig(true, Mode.PRIMARY, 0, IndexDataLocationType.LOCAL)
+            .build();
+
+    primaryServer.createIndex(testPartialUpdateIndex);
+    primaryServer.registerFields(testPartialUpdateIndex, fields);
+    primaryServer.startPrimaryIndex(testPartialUpdateIndex, -1, null);
+
+    primaryServer.addDocs(buildAddDocRequest(3).stream());
+    primaryServer.commit(testPartialUpdateIndex);
+    Thread.sleep(1000);
```
aprudhomme, line 83: Are these sleeps necessary. If you are waiting for an index refresh, it can be triggered on `primaryServer`

src/main/java/com/yelp/nrtsearch/server/monitoring/IndexingMetrics.java
```diff
@@ -0,0 +1,66 @@
...
+
+  // counter for addDocument requests received for the index with the index name as the label value
+  public static final Counter addDocumentRequestsReceived =
+      Counter.builder()
+          .name(ADD_DOCUMENT_REQUESTS_RECEIVED)
+          .help("Number of requests received for the add document API ")
+          .labelNames("index")
+          .build();
+
+  public static final Gauge updateDocValuesLatency =
```
aprudhomme, line 45: A Gauge type metric will only hold the last value. This would be better as a histogram or summary type

RavirajDataEngineering:
@aprudhomme Thanks a lot for reviewing the PR , I have addressed those comments, please let me know if anything else needs a change. I also wanted to check if `indexName` label is enough for these metrics or we need anything more

RavirajDataEngineering:
@aprudhomme I realized that the request can have empty fields set when making changes to `getUpdatableDocValueField` . Added some checks for it and some unit tests to cover those scenarios.

src/main/java/com/yelp/nrtsearch/server/handler/AddDocumentHandler.java
```diff
@@ -488,7 +564,12 @@ private void updateNestedDocuments(
       }
 
       documents.add(rootDoc);
+      IndexingMetrics.addDocumentRequestsReceived.labelValues(indexName).inc();
+      long timeInMillis = System.currentTimeMillis();
```
aprudhomme, line 568: Sorry, I could have been more clear. I meant doing something like `(ns_end - ns_start) / 1000000.0` so you would record milliseconds as a double with a fractional component.

src/main/java/com/yelp/nrtsearch/server/handler/AddDocumentHandler.java
```diff
@@ -460,6 +470,72 @@ public long runIndexingJob() throws Exception {
...
+        Term term = null;
+        for (Map.Entry<String, MultiValuedField> entry :
+            addDocumentRequest.getFieldsMap().entrySet()) {
+          FieldDef field = indexState.getField(entry.getKey());
+          if (field == null) {
+            throw new IllegalArgumentException(
+                String.format("Field: %s is not registered", entry.getKey()));
+          }
+          if (field.getName().equals(indexState.getIdFieldDef().get().getName())) {
+            term = buildTermForDocValueUpdate(indexState, addDocumentRequest);
```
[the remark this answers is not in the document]
RavirajDataEngineering: Yes , that will be a lot cleaner. Thanks

aprudhomme (approved):

PR #849, opened 2025-04-16
Add sum aggregation support

aprudhomme:
I was giving the vscode agent mode a try, so this was written entirely by AI.

This adds support for sum aggregations of double values produced by a script. This aggregation also supports being used to order parent buckets.


clientlib/src/main/java/com/yelp/nrtsearch/server/collectors/BucketOrder.java
```diff
@@ -121,9 +121,10 @@ public static BucketOrder createBucketOrder(
     if (nestedCollector == null) {
       throw new IllegalArgumentException("Nested collector not found: " + grpcBucketOrder.getKey());
     }
-    // currently, max and min are the only aggregations that can be used for ordering
+    // collectors that can be used for ordering: max, min, and now sum
```
sarthakn7, line 124: Nit: 
```suggestion
    // collectors that can be used for ordering: max, min, and sum
```

sarthakn7 (dismissed):

PR #855, opened 2025-04-30
feat(plugin): add ingestion plugin framework and example implementation

umeshdangat:
- Introduced AbstractIngestionPlugin base class to support ingestion plugins
- Defined IngestionPlugin interface with start/stop lifecycle methods
- Updated NrtsearchServer to initialize ingestion plugin state
- Implemented ExamplePlugin as an ingestion + analysis plugin
- Added test coverage for ingestion using Awaitility for async validation
- Registered example plugin in build and settings
- Bumped version to 1.0.0-beta.11

### DIFF
--- a/example-plugin/build.gradle
+++ b/example-plugin/build.gradle
@@ -34,6 +34,7 @@ dependencies {
     testImplementation libs.grpc.testing
     testImplementation libs.junit
     testImplementation libs.protobuf.java
+    testImplementation("org.awaitility:awaitility:4.3.0")
 }
 
 distributions {
--- a/example-plugin/src/test/java/com/yelp/nrtsearch/plugins/example/ExamplePluginTest.java
+++ b/example-plugin/src/test/java/com/yelp/nrtsearch/plugins/example/ExamplePluginTest.java
@@ -15,7 +15,9 @@
  */
 package com.yelp.nrtsearch.plugins.example;
 
+import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
 
 import com.yelp.nrtsearch.server.ServerTestCase;
 import com.yelp.nrtsearch.server.config.NrtsearchConfig;
@@ -26,6 +28,7 @@ import com.yelp.nrtsearch.server.grpc.CustomResponse;
 import com.yelp.nrtsearch.server.grpc.FieldDefRequest;
 import com.yelp.nrtsearch.server.grpc.MatchQuery;
 import com.yelp.nrtsearch.server.grpc.Query;
+import com.yelp.nrtsearch.server.grpc.RefreshRequest;
 import com.yelp.nrtsearch.server.grpc.SearchRequest;
 import com.yelp.nrtsearch.server.grpc.SearchResponse;
 import com.yelp.nrtsearch.server.plugins.Plugin;
@@ -37,41 +40,81 @@ import java.util.stream.Stream;
 import org.apache.lucene.analysis.Analyzer;
 import org.apache.lucene.analysis.TokenStream;
 import org.apache.lucene.analysis.tokenattributes.CharTermAttribute;
+import org.junit.After;
+import org.junit.Before;
 import org.junit.ClassRule;
 import org.junit.Test;
 
 public class ExamplePluginTest extends ServerTestCase {
   @ClassRule public static final GrpcCleanupRule grpcCleanup = new GrpcCleanupRule();
+  private static final String INGESTION_TEST_INDEX = ExamplePlugin.INGESTION_TEST_INDEX;
 
-  private static final ExamplePlugin examplePlugin = new ExamplePlugin(getConfig());
+  private ExamplePlugin examplePlugin;
+  private static final String FIELD_1 = "field1";
+
+  @Override
+  protected List<String> getIndices() {
+    return List.of(DEFAULT_TEST_INDEX, INGESTION_TEST_INDEX);
+  }
 
   @Override
   protected List<Plugin> getPlugins(NrtsearchConfig configuration) {
+    examplePlugin = new ExamplePlugin(configuration);
     return List.of(examplePlugin);
   }
 
   @Override
   protected FieldDefRequest getIndexDef(String name) throws IOException {
-    return getFieldsFromResourceFile("/register_fields.json");
+    if (DEFAULT_TEST_INDEX.equals(name)) {
+      return getFieldsFromResourceFile("/register_fields.json");
+    } else if (INGESTION_TEST_INDEX.equals(name)) {
+      return getFieldsFromResourceFile("/register_fields_ingestion.json");
+    } else {
+      throw new IllegalArgumentException("Unknown index: " + name);
+    }
   }
 
   @Override
   protected void initIndex(String name) throws Exception {
-    AddDocumentRequest addDocumentRequest =
-        AddDocumentRequest.newBuilder()
-            .setIndexName(name)
-            .putFields(
-                "field1",
-                MultiValuedField.newBuilder().addValue("How to use Nrtsearch<br>").build())
-            .build();
-    AddDocumentRequest addDocumentRequest2 =
-        AddDocumentRequest.newBuilder()
-            .setIndexName(name)
-            .putFields(
-                "field1",
-                MultiValuedField.newBuilder().addValue("<head>How to create plugin</head>").build())
-            .build();
-    addDocuments(Stream.of(addDocumentRequest, addDocumentRequest2));
+    // Only initialize test docs for analysis test index
+    if (DEFAULT_TEST_INDEX.equals(name)) {
+      AddDocumentRequest addDocumentRequest =
+          AddDocumentRequest.newBuilder()
+              .setIndexName(name)
+              .putFields(
+                  "field1",
+                  MultiValuedField.newBuilder().addValue("How to use Nrtsearch<br>").build())
+              .build();
+      AddDocumentRequest addDocumentRequest2 =
+          AddDocumentRequest.newBuilder()
+              .setIndexName(name)
+              .putFields(
+                  "field1",
+                  MultiValuedField.newBuilder()
+                      .addValue("<head>How to create plugin</head>")
+                      .build())
+              .build();
+      addDocuments(Stream.of(addDocumentRequest, addDocumentRequest2));
+    }
+  }
+
+  @Before
+  public void ensurePluginInitialized() throws Exception {
+    if (examplePlugin == null) {
+      for (Plugin plugin : getPlugins(getConfig())) {
+        if (plugin instanceof ExamplePlugin) {
+          examplePlugin = (ExamplePlugin) plugin;
+          break;
+        }
+      }
+    }
+  }
+
+  @After
+  public void cleanupPlugin() throws Exception {
+    if (examplePlugin != null) {
+      examplePlugin.stopIngestion();
+    }
   }
 
   @Test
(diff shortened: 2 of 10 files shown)


example-plugin/build.gradle
```diff
@@ -34,6 +34,7 @@ dependencies {
     testImplementation libs.grpc.testing
     testImplementation libs.junit
     testImplementation libs.protobuf.java
+    testImplementation("org.awaitility:awaitility:4.3.0")
```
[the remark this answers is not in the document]
umeshdangat: fixed

src/main/java/com/yelp/nrtsearch/server/plugins/AbstractIngestionPlugin.java
```diff
@@ -0,0 +1,95 @@
...
+   * @throws IOException if there are commit errors
+   */
+  protected final void commit(String indexName) throws IOException {
+    verifyInitialized(indexName);
+    IndexState indexState = globalState.getIndexOrThrow(indexName);
+    ShardState shard = indexState.getShard(0);
+    if (shard == null) {
+      throw new IllegalStateException("No shard found for index");
+    }
+    shard.commit();
```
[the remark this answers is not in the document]
umeshdangat: fixed

src/main/java/com/yelp/nrtsearch/server/plugins/AbstractIngestionPlugin.java
```diff
@@ -0,0 +1,95 @@
...
+  protected AbstractIngestionPlugin(NrtsearchConfig config) {
+    this.config = config;
+  }
+
+  /**
+   * Initialize plugin state with index access. Called by NrtSearchServer when index is ready.
+   *
+   * @param globalState The global server state
+   */
+  public final void initializeState(GlobalState globalState) throws IOException {
```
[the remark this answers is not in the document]
umeshdangat: - Updated AbstractIngestionPlugin to launch startIngestion asynchronously using plugin-provided ExecutorService
   - Added getIngestionExecutor() abstract method for plugin-controlled threading
   - Updated NrtsearchServer to initialize ingestion plugins after startup

src/main/java/com/yelp/nrtsearch/server/plugins/IngestionPlugin.java
```diff
@@ -0,0 +1,38 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.yelp.nrtsearch.server.plugins;
+
+import java.io.IOException;
+
+/** Interface for ingestion plugins. Defines lifecycle methods that plugins must implement. */
+public interface IngestionPlugin {
```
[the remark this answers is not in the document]
umeshdangat: removed

src/main/java/com/yelp/nrtsearch/server/plugins/AbstractIngestionPlugin.java
```diff
@@ -0,0 +1,95 @@
...
+  /**
+   * Add documents. Available for plugin implementations to call during ingestion.
+   *
+   * @param addDocRequests List of document requests to add
+   * @return The sequence number of the indexing operation
+   * @throws IOException if there are indexing errors
+   */
+  protected final long addDocuments(List<AddDocumentRequest> addDocRequests, String indexName)
+      throws Exception {
+    return new AddDocumentHandler.DocumentIndexer(globalState, addDocRequests, indexName)
```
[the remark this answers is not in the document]
umeshdangat: now that plugins have control of their own executorService startIngestion itself is run in an a plugin controlled executorservice e.g. kafka plugin

```
@Override
public void startIngestion() throws IOException {
  logger.info("Starting parallel ingestion from Kafka partitions");

  int numPartitions = 3;
  for (int i = 0; i < numPartitions; i++) {
    final int partitionId = i;
    getIngestionExecutor().submit(() -> {
      try {
        ingestFromPartition(partitionId);
      } catch (Exception e) {
        logger.error("Error ingesting from partition {}", partitionId, e);
      }
    });
  }
}


private void ingestFromPartition(int partitionId) {
  logger.info("Ingesting from partition {}", partitionId);
  while (running.get()) {
    // Poll Kafka for partitionId
    // Convert messages to AddDocumentRequest
    // Call addDocuments(...)
    // Optionally call commit(...)
  }
}
```

src/main/java/com/yelp/nrtsearch/server/grpc/NrtsearchServer.java
```diff
@@ -158,6 +159,7 @@ public void start() throws IOException {
     GlobalState globalState = serverImpl.getGlobalState();
 
     registerMetrics(globalState);
+    initIngestionPlugin(globalState, plugins);
```
aprudhomme, line 162: I think the best place to do the initialization is https://github.com/Yelp/nrtsearch/blob/main/src/main/java/com/yelp/nrtsearch/server/grpc/NrtsearchServer.java#L387 after the global state is constructed. Doing this in `LuceneServerImpl` means it will work with the test server classes.

src/main/java/com/yelp/nrtsearch/server/plugins/AbstractIngestionPlugin.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.concurrent.ExecutorService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Base class for ingestion plugins that handles state management and indexing operations. Plugin
+ * implementations: 1. Must implement startIngestion/stopIngestion to handle source-specific logic
+ * 2. Should use addDocuments/commit methods to index data during ingestion
+ */
+public abstract class AbstractIngestionPlugin extends Plugin {
```
aprudhomme, line 34: I would advise against making the plugin an abstract class. The `Plugin` is designed to potentially provide a lot of different functionality. Forcing it to contain the ingestion implementation would be odd.

I would instead suggest something more like:
```
interface IngestionPlugin extends Plugin {
  AbstractIngestor getIngestor()
}
```

And putting the abstract class in `com.yelp.nrtsearch.server.ingestion`

You could also use an `Ingestor` interface instead, and make `AbstractIngestor` implement that.

src/main/java/com/yelp/nrtsearch/server/config/NrtsearchConfig.java
```diff
@@ -434,4 +434,25 @@ private static List<String> getPluginSearchPath(Object o) {
...
+      Object raw = configReader.get("pluginConfigs.ingestion", obj -> obj);
+      if (raw instanceof Map<?, ?> outerMap) {
+        Map<String, Map<String, Object>> result = new HashMap<>();
+        for (Map.Entry<?, ?> entry : outerMap.entrySet()) {
+          if (entry.getKey() instanceof String pluginName
+              && entry.getValue() instanceof Map<?, ?> pluginConfig) {
+            result.put(pluginName, (Map<String, Object>) pluginConfig);
+          }
+        }
+        return result;
```
aprudhomme, line 450: You could simply this code a bit using a jackson annotated class. You can load it like https://github.com/Yelp/nrtsearch/blob/main/src/main/java/com/yelp/nrtsearch/server/config/ThreadPoolConfiguration.java#L173

example-plugin/src/test/java/com/yelp/nrtsearch/plugins/example/ExamplePluginTest.java
```diff
@@ -37,41 +40,80 @@
 import org.apache.lucene.analysis.Analyzer;
 import org.apache.lucene.analysis.TokenStream;
 import org.apache.lucene.analysis.tokenattributes.CharTermAttribute;
+import org.junit.After;
+import org.junit.Before;
 import org.junit.ClassRule;
 import org.junit.Test;
 
 public class ExamplePluginTest extends ServerTestCase {
```
aprudhomme, line 48: Is there a reason you are putting the tests in the `example-plugin` instead of the main testing?

src/main/java/com/yelp/nrtsearch/server/plugins/AbstractIngestionPlugin.java
umeshdangat: > I would advise against making the plugin an abstract class. 
Not sure I understand. I have not changed anything in the existing Plugin class? 

> Forcing it to contain the ingestion implementation would be odd.
I created a new `AbstractIngestionPlugin ` that extends the existing Plugin class. This is similar to what you suggesting? 
`interface IngestionPlugin extends Plugin` 
vs
`abstract class AbstractIngestionPlugin extends Plugin` ?

aprudhomme: Using an interface makes code organization cleaner. Otherwise, the ingestion implementation is forced to reside in the main plugin class.

This is also a pattern that cannot be repeated for other plugin types, since a class can implement many interfaces, but extend only one base class. Using abstract classes would unnecessarily limit what functionality could be colocated in a single plugin. For flexibility, having all plugins types being interfaces is better.

umeshdangat: Plugin class is already an abstract class NOT an interface: https://github.com/Yelp/nrtsearch/blob/main/src/main/java/com/yelp/nrtsearch/server/plugins/Plugin.java#L22

To do what you suggest
```
interface IngestionPlugin extends Plugin {      **//interface cannot extend abstract class**
  AbstractIngestor getIngestor()
}
```

I am going to have to change existing Plugin abstract class to an interface with something like

```
public interface Plugin extends Closeable {
  @Override
  default void close() throws IOException {}
}

```

This then means I have to change current implementations of Plugins  (including test plugins)

> from extends Plugin to implements Plugin ([places of usage ](https://github.com/search?q=repo%3AYelp%2Fnrtsearch%20%22extends%20Plugin%22&type=code))

Let me know if this is what you suggesting? It should be search and replace but will make files changed in this PR a lot more.

umeshdangat: Alternatively, I can just not change any existing `Plugin` behavior and do something like so...

```
public interface IngestionPlugin {
  Ingestor getIngestor();
}
```


`public abstract class AbstractIngestor implements Ingestor `


`public class ExampleIngestionPlugin extends Plugin implements IngestionPlugin`

By doing the above I think we still manage to keep `Plugin` class untouched and introduce ingestion functionality in `com.yelp.nrtsearch.server.ingestion`

example-plugin/src/test/java/com/yelp/nrtsearch/plugins/example/ExamplePluginTest.java
umeshdangat: fixed. Added relevant tests `IngestionPluginUtilsTest`,  `AbstractIngestorTest`

src/main/java/com/yelp/nrtsearch/server/plugins/AbstractIngestionPlugin.java
umeshdangat: added the above ^.

src/main/java/com/yelp/nrtsearch/server/grpc/NrtsearchServer.java
umeshdangat: done

src/main/java/com/yelp/nrtsearch/server/config/NrtsearchConfig.java
umeshdangat: done

src/main/java/com/yelp/nrtsearch/server/ingestion/IngestionPluginUtils.java
```diff
@@ -0,0 +1,41 @@
...
+      abstractIngestor.initialize(globalState);
+    }
+
+    ExecutorService executor = plugin.getIngestionExecutor();
+    executor.submit(
+        () -> {
+          try {
+            ingestor.start();
+          } catch (IOException e) {
+            e.printStackTrace();
```
sarthakn7: You can log this error with slf4j logger

sarthakn7 (commented):
Lgtm, just left a small comment

sarthakn7 (approved):

PR #856, opened 2025-05-01
Add option to use separate executor to process commit tasks

aprudhomme:
Adds the option `useSeparateCommitExecutor` to the nrtsearch config file, default false. When true, commit tasks use a separate executor for processing.

We sometimes see issues where commit tasks get backed up behind indexing tasks in the indexing executor queue. This can lead to commit requests timing out before they are processed, which keeps indexers from making progress. Using a separate pool would let the commit operation start regardless of what other indexing tasks are in progress. This should be safe, since the indexer only intends to commit completed indexing operations.

I set the default thread count/queue size for the executor to 5. In practice, there should be very few commits sent concurrently.

This code was written mostly by AI.

### DIFF
--- a/docs/server_configuration.rst
+++ b/docs/server_configuration.rst
@@ -111,6 +111,11 @@ Example server configuration
      - int
      - Maximum number of clauses in a query
      - 1024
+   
+   * - useSeparateCommitExecutor
+     - bool
+     - If enabled, the server will use a separate executor for commit operations instead of using the indexing executor.
+     - false
 
 .. list-table:: `Threadpool Configuration <https://github.com/Yelp/nrtsearch/blob/master/src/main/java/com/yelp/nrtsearch/server/config/ThreadPoolConfiguration.java>`_ (``threadPoolConfiguration.*``)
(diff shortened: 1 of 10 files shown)


docs/server_configuration.rst
```diff
@@ -241,6 +246,21 @@ Example server configuration
      - Name prefix for threads created by vector merge threadpool executor
      - VectorMergeExecutor
 
+   * - commit.maxThreads
+     - int
+     - Size of commit threadpool executor
+     - 5
```
sarthakn7, line 252: Should the default number be 1 since only one commit can happen at a time for an index?

aprudhomme: The synchronized scope for commit is reduced in v1 https://github.com/Yelp/nrtsearch/blob/main/src/main/java/com/yelp/nrtsearch/server/index/ShardState.java#L358 this allows multiple pending commits to join the same upload task. So it is good for them to run concurrently.

sarthakn7 (approved):

PR #868, opened 2025-07-14
dhruva_RP-13223-Parent_field_access_from_nested_query

dhruva-r:
## Jira Ticket

RP-13223
## Summary of Changes

Added support for nested queries to be able to access parent document fields. The `docLookup` object will now use a new `NESTED_DOCUMENT_OFFSET` meta field to calculate the docId of the the parent document and will be able to retrieve fields from the parent document if they do not exist in the child document.
## Tests

Unit tests and an end to end test created provide sufficient code coverage.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
+++ b/src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
@@ -185,21 +185,11 @@ public class SegmentDocLookup implements Map<String, LoadedDocValues<?>> {
    * @throws IllegalArgumentException if a parent docId cannot be found or does not exist
    */
   private int getParentDocId() {
-    FieldDef offsetFieldDef;
-    try {
-      offsetFieldDef = IndexState.getMetaField(IndexState.NESTED_DOCUMENT_OFFSET);
-    } catch (IllegalArgumentException e) {
-      throw new IllegalArgumentException(
-          "Document is not a nested document - no parent document available", e);
-    }
-
-    if (!(offsetFieldDef instanceof IndexableFieldDef<?> offsetIndexableFieldDef)) {
-      throw new IllegalArgumentException("NESTED_DOCUMENT_OFFSET field is not indexable");
-    }
-
+    IndexableFieldDef<?> offsetFieldDef =
+        (IndexableFieldDef<?>) IndexState.getMetaField(IndexState.NESTED_DOCUMENT_OFFSET);
     LoadedDocValues<?> offsetDocValues;
     try {
-      offsetDocValues = offsetIndexableFieldDef.getDocValues(context);
+      offsetDocValues = offsetFieldDef.getDocValues(context);
       offsetDocValues.setDocId(docId);
     } catch (IOException e) {
       throw new IllegalArgumentException("Could not load nested document offset values", e);


src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
```diff
@@ -96,6 +97,9 @@ public boolean containsValue(Object value) {
    * Get the {@link LoadedDocValues} for a given field. Creates a new instance or uses one from the
    * cache. The data is loaded for the current set document id.
    *
+   * <p>If the field is not found in the current document and it is a nested document, it will
```
waziqi89: The client shall be aware of using the parent field, and using some notation for the parent field. For example, "_PARENT.biz_feature_a"

src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
```diff
@@ -128,9 +132,92 @@ public LoadedDocValues<?> get(Object key) {
...
+      return null;
+    }
+
+    Object offsetValue = offsetDocValues.getFirst();
+    int offset = ((Number) offsetValue).intValue();
+
+    // The offset represents the exact number of documents to jump forward to reach the parent
+    int parentDocId = docId + offset;
+
+    FieldDef fieldDef = fieldDefLookup.apply(fieldName);
```
waziqi89: fieldDef can be passed as an argument and then you don't have to lookup again here

dhruva-r: The initialization of fieldDef in the get() method occurs inside an if statement, I could either move this initialization outside where it may not be safe or modify when I call the tryGetFromParentDocument method to be inside the scope of the if statement. Is it safe to just move the initialization outside like this?:

```
    FieldDef fieldDef = fieldDefLookup.apply(fieldName);
    if (docValues == null) {
      
      if (fieldDef == null) {
        throw new IllegalArgumentException("Field does not exist: " + fieldName);
      }
```
Edit - nevermind this just ruins the cache check that we do in the previous line

src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
dhruva-r: I agree that this is a much more efficient flow. From my understanding when the client calls the docLookup it does not use this notation currently? I believe it simply searches biz_feature_a.

dhruva-r: Hey @waziqi89 to implement this notation, would the "_PARENT" portion of the string be called "_PARENT" consistently or would it take the name of the parent document in that portion of the string?

waziqi89: I think you got what I mean.

src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
```diff
@@ -79,6 +82,17 @@ public boolean containsKey(Object key) {
       return false;
     }
     String fieldName = key.toString();
+
+    if (fieldName.startsWith("_PARENT.")) {
+      String parentFieldName = fieldName.substring("_PARENT.".length());
```
waziqi89: nit: looks like you can just substring in-place to update the fileName, and then you don't have to duplicate the other codes

src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
```diff
@@ -96,6 +110,13 @@ public boolean containsValue(Object value) {
    * Get the {@link LoadedDocValues} for a given field. Creates a new instance or uses one from the
    * cache. The data is loaded for the current set document id.
    *
+   * <p>If the field is not found in the current document and it is a nested document, it will
+   * attempt to find the field in the parent document using the NESTED_DOCUMENT_OFFSET field.
```
waziqi89: Is this fallback attemption necessary?
Wonder what if I have a schema with same feature names:
```
{
feature_A: 1.0
children: [{
   feature_A: null
}]
}
```

dhruva-r: This is a good point, I think the motivation behind having the fallback is so that we are not forced to change lookup calls made by the clients. I think it is likely better to remove the fallback altogether because of this case. @sarthakn7 do you have any thoughts?

src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
```diff
@@ -128,9 +158,84 @@ public LoadedDocValues<?> get(Object key) {
...
+  /**
+   * Attempt to retrieve the field from the parent document using NESTED_DOCUMENT_OFFSET.
+   *
+   * @param fieldName the name of the field to retrieve
+   * @param fieldDef the definition of the field to retrieve
+   * @return LoadedDocValues from parent document, or null if not found or not a nested document
+   * @throws IllegalArgumentException if there are issues accessing the offset field or parent
+   *     document
+   */
+  private LoadedDocValues<?> tryGetFromParentDocument(String fieldName, FieldDef fieldDef) {
```
waziqi89: Let's get consensus on whether we want to get rid of the fallback mech. If so, we can simplify the code significantly

src/main/java/com/yelp/nrtsearch/server/field/ObjectFieldDef.java
```diff
@@ -74,27 +73,36 @@ public void parseFieldWithChildren(
...
-              .collect(Collectors.toList());
+      int totalDocs = fieldValueMaps.size();
+      List<Document> childDocuments = new ArrayList<>(totalDocs);
+
+      for (int i = 0; i < totalDocs; i++) {
+        // Calculate offset as n-i (total docs minus current index)
+        int offset = totalDocs - i;
+        childDocuments.add(createChildDocument(fieldValueMaps.get(i), facetHierarchyPaths, offset));
+      }
+
```
aprudhomme: I don't think this will work if there are multiple nested documents. This could maybe be done during https://github.com/Yelp/nrtsearch/blob/main/src/main/java/com/yelp/nrtsearch/server/handler/AddDocumentHandler.java#L567 instead, in the `updateNestedDocuments`/`addNestedDocuments`. At that point, all the global set of documents to add are known.

There is also the issue of handling multiple levels of nested document (which I don't think is implemented properly right now).

src/main/java/com/yelp/nrtsearch/server/index/IndexState.java
```diff
@@ -564,6 +565,17 @@ private static Map<String, FieldDef> getPredefinedMetaFields(GlobalState globalS
...
                     .build(),
+                FieldDefCreator.createContext(globalState)),
+        NESTED_DOCUMENT_OFFSET,
+        FieldDefCreator.getInstance()
+            .createFieldDef(
+                NESTED_DOCUMENT_OFFSET,
+                Field.newBuilder()
+                    .setName(NESTED_DOCUMENT_OFFSET)
+                    .setType(FieldType.INT)
+                    .setSearch(true)
```
aprudhomme: Does this need to be searchable, or is it only going to be used to load the doc value?

src/main/java/com/yelp/nrtsearch/server/index/IndexState.java
```diff
@@ -83,6 +83,7 @@ public abstract class IndexState implements Closeable {
   public static final String NESTED_PATH = "_nested_path";
   public static final String ROOT = "_root";
   public static final String FIELD_NAMES = "_field_names";
+  public static final String NESTED_DOCUMENT_OFFSET = "_offset";
```
aprudhomme: Maybe something like `_parent_offset` would be more descriptive?

src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
```diff
@@ -106,6 +118,24 @@ public boolean containsValue(Object value) {
...
+      LoadedDocValues<?> parentDocValues =
+          tryGetFromParentDocument(parentFieldName, parentFieldDef);
+      if (parentDocValues == null) {
+        throw new IllegalArgumentException(
+            "Could not access parent field: "
+                + parentFieldName
+                + " (document may not be nested or parent field may not exist)");
+      }
+      return parentDocValues;
+    }
```
aprudhomme: This may be able to be simplified by doing a recursive lookup. This class could hold a (lazy initialized) `SegmentDocLookup` for its parent. If the `fieldName` starts with `_PARENT.`, strip the prefix and use the parent lookup to load the suffix name. 

This has some advantages, like built in `LoadedDocValue` caching for all parents in a segment. It would also support multiple nested doc levels like `_PARENT._PARENT.field_name`.

src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
```diff
@@ -106,6 +118,24 @@ public boolean containsValue(Object value) {
   public LoadedDocValues<?> get(Object key) {
     Objects.requireNonNull(key);
     String fieldName = key.toString();
+
+    if (fieldName.startsWith("_PARENT.")) {
```
sarthakn7: I'm wondering if we need the `.PARENT` notation. Can we do something like this:
```
if (!"".equals(nestedPath) && !fieldName.startsWith(nestedPath + ".")) {
  getFieldFromParentDoc()
} else {
  getFieldFromCurrentDoc()
}
```
This would require passing the nestedPath from the search request here. It can maybe be passed into DocLookup when it's being created in the search context.

waziqi89: This would only work if we have 1 layer. When more than one layer is used, the middle nestedPath is unprovided, and split/prefix-match has to be done anyways.

And we've already introduced the "_parent" term in the metafield anyways

src/main/java/com/yelp/nrtsearch/server/field/ObjectFieldDef.java
waziqi89: Hi @aprudhomme ! Do you think we need a mini-rework on indexing for nested? I think https://jira.yelpcorp.com/browse/RP-8578 can be easily implemented if we choose so, too

dhruva-r: I will need to verify if this doesn't work for multiple nested documents. I believe I talked about this case with @sarthakn7 initially and I ended up going ahead with setting the offset here, but I will double check.

dhruva-r: As for handling multiple levels of nested documents the intention is to not handle that during this ticket but to make sure that a future change can be made with lower impedance.

src/main/java/com/yelp/nrtsearch/server/index/IndexState.java
dhruva-r: This is only going to be used to load the doc value, I was under the impression initially it needed to be searchable to do so for some reason.

src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
sarthakn7: > the middle nestedPath is unprovided

Wouldn't the nested path have to include other layers too? E.g. one level nesting would be `delivery_area`, two levels would be `delivery_area.partner_details`

waziqi89: I see @sarthakn7 's point now.
For a nested query, we normally have childFields name with prefix, saying `reviews.rating`. If prepending a "_parent" to denote the parent field, the result looks a bit nontrivial - `_PARENT.reviews.rating` (where "_parent" and "reviews" can be cancelled out).

Instead, if we pass the nestedPath to the lookup method, we now have 1. current location (nestedPath) 2. absolute field path (eg, "biz_name"). We can implement a util function to resolve the relative route from 1 to 2. Then we know we need to step one step back without any probing. This applies to 3 layer nested, when 1 == "reviews.version" 2 =="biz_name", the util shall know there are 2-backstep to retrieve the field.

All in all, looks like the _parent prefix is unncessary.

dhruva-r: I have decided to push an update to the PR without this change for now as I ran into some issues trying to implement it this way, so I wanted to have a working version with the recursive solution on the branch

src/main/java/com/yelp/nrtsearch/server/field/ObjectFieldDef.java
dhruva-r: After a bit of investigating I have decided to move this functionality to the AddDocumentHandler.

dhruva-r:
I need to add a test for multiple nested documents (and checking the offset values), a test for multi levelled nested documents

src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
```diff
@@ -122,16 +124,27 @@ public LoadedDocValues<?> get(Object key) {
...
+      if (parentLevels > 0) {
+        SegmentDocLookup currentLookup = getParentLookup();
+        if (currentLookup == null) {
+          throw new IllegalArgumentException(
+              "Could not access parent field: "
+                  + fieldName
+                  + " (document may not be nested or parent field may not exist)");
+        }
+
+        for (int i = 1; i < parentLevels; i++) {
```
dhruva-r: In a previous commit when I was working with the `_PARENT.` prefix I used a better formatted recursive approach that @aprudhomme provided. That approach was not as intuitive to me when I removed the prefix so I decided to use this method instead.

src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
```diff
@@ -128,9 +169,104 @@ public LoadedDocValues<?> get(Object key) {
...
+    int minLength = Math.min(currentPathParts.length, targetPathParts.length);
+    for (int i = 0; i < minLength; i++) {
+      if (currentPathParts[i].equals(targetPathParts[i])) {
+        commonPrefixLength++;
+      } else {
+        break;
+      }
+    }
+
+    int levelsUp = currentPathParts.length - commonPrefixLength;
```
aprudhomme: I think an issue here might be that the path components do not always denote a nested document level. It is also used to access child fields of indexable fields and fields within non-nested objects. This could cause an incorrect level result in some cases.

src/main/java/com/yelp/nrtsearch/server/search/SearchRequestProcessor.java
```diff
@@ -134,32 +134,36 @@ public static SearchContext buildContextForRequest(
         .setExplain(searchRequest.getExplain())
         .setWarming(warming);
 
-    Map<String, FieldDef> queryVirtualFields = getVirtualFields(indexState, searchRequest);
-    Map<String, FieldDef> queryRuntimeFields = getRuntimeFields(indexState, searchRequest);
+    String rootQueryNestedPath =
+        indexState.resolveQueryNestedPath(searchRequest.getQueryNestedPath());
```
aprudhomme: The lookup is being set to use the query nested path. Would this have issues being used in a `NestedQuery`?

src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
```diff
@@ -54,6 +65,8 @@ public SegmentDocLookup(Function<String, FieldDef> fieldDefLookup, LeafReaderCon
    */
   public void setDocId(int docId) {
     this.docId = docId;
+    // Reset parent lookup when document changes
+    this.parentLookup = null;
```
aprudhomme: This will cause a new lookup to be created per document. It would be better to create the lookup once and reset the id per document. This could be done by maintaining a `parentDocId` value that is reset when the `docId` is set.

src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
```diff
@@ -106,6 +123,30 @@ public boolean containsValue(Object value) {
   public LoadedDocValues<?> get(Object key) {
     Objects.requireNonNull(key);
     String fieldName = key.toString();
+
+    if (queryNestedPath != null && !queryNestedPath.isEmpty() && !"_root".equals(queryNestedPath)) {
+      int parentLevels = resolveParentLevels(queryNestedPath, fieldName);
+      if (parentLevels > 0) {
+        SegmentDocLookup currentLookup = getParentLookup();
+        if (currentLookup == null) {
```
aprudhomme: One thing to consider is that this code is used in a very hot loop for query execution (per field access, per recalled document). You need to be careful about how much work is being done.

One thought, does the parent determination really need to be automatic. What if the `ScoreScript` provided a `getParentDocLookup` method that took an optional level value. This would allow the lookup to remain simple, as it could alway assume it is in the right document scope.

dhruva-r: I like your approach because it allows the segment doclookup to remain clean. I'll have to think about making the doclookup not automatic. In the previous solution the client would pass "_PARENT" prefix to indicate the level they wanted to access and this worked well because it was easy to navigate to the correct document with negligible processing time. I think that passing the level as an integer though seems a little weird and unintuitive in my head.

I think eventually this will mean that people will have to essentially use this same loop to determine the level but not at lookup time.

src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
dhruva-r: Does this mean we cannot use the nestedPath to determine level? Or could I have an example string of a nested path that would not work?

src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
```diff
@@ -106,6 +123,30 @@ public boolean containsValue(Object value) {
   public LoadedDocValues<?> get(Object key) {
     Objects.requireNonNull(key);
     String fieldName = key.toString();
+
+    if (queryNestedPath != null && !queryNestedPath.isEmpty() && !"_root".equals(queryNestedPath)) {
```
sarthakn7: You can use `IndexState.ROOT` instead of `"_root"`. And you can just check for the last condition (`!IndexState.ROOT.equals(queryNestedPath)`). If the queryNestedPath is null or empty, this condition will still be false.

src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
sarthakn7: `queryNestedPath` should only be used with nested documents though. The function `resolveQueryNestedPath` is called in SearchRequestProcessor, and it would throw an error if the path does not belong to a field with nested documents. I think it should still be okay to use it here, but will let @aprudhomme confirm.

src/main/java/com/yelp/nrtsearch/server/search/SearchRequestProcessor.java
sarthakn7: If my understanding is correct, nested query matches the parent docs of the nested documents. With the way we have implemented it, it would only work properly with one level of nesting.
Providing nestedQueryPath matches only child documents. I'm not sure if both of them can/should be used together.

src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
sarthakn7: Would it help if we cache the return value of `resolveParentLevels(queryNestedPath, fieldName)`? It would have to computed only once for every queryNestedPath and fieldName during the lifetime of a replica.

src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
aprudhomme: I still think this ends up being a problem when a nested object is a child field. 
For a nested path `field1.field2.nested_object`, where neither `field1` or `field2` are nested. Looking up `field1.field3.field4` would compute to 2 levels up, but it is actually only 1.

src/main/java/com/yelp/nrtsearch/server/search/SearchRequestProcessor.java
aprudhomme: But what if one of the queries against the nested documents needs to use the lookup (ScoreScript, etc). Would it be forced into the query nested scope?

src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
aprudhomme: The thought would be that the `ScoreScript` would provide the parent lookups with something like:
`Map<String, LoadedDocValue<?>> getParentLookup()` and/or
`Map<String, LoadedDocValue<?>> getParentLookup(int level)`
The resulting lookups would be set the the given nested scope. The parent lookups could still be created/cached within the `SegmentDocLookup` recursively.

dhruva-r: @aprudhomme To prevent the scope of this PR from increasing, I am removing multi layer nesting from this PR as it was originally out of scope. This is due to the complexity of the testing effort, more info [here](https://yelp.slack.com/archives/C08USG27679/p1753194319307779). 
In a future PR when we eventually enable multi layer nesting we can open up this discussion again as I agree, the doclookup should remain simple.

src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
dhruva-r: @aprudhomme Because I am restricting the scope of the PR to only support 1 level of nesting, is it reasonable to just return 1 from this function. This way the case that you are talking about will still perform the search on the direct parent document. The other option is to revert the PR to use the `_PARENT` prefix like before but @sarthakn7 had concerns about doing this. Currently if a level above 1 is detected the doclookup will throw an exception.

src/main/java/com/yelp/nrtsearch/server/search/SearchRequestProcessor.java
dhruva-r: @sarthakn7 I am highly considering reverting to use the `_PARENT` prefix since it would avoid the concerns that Andrew has here.

src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
```diff
@@ -128,9 +154,67 @@ public LoadedDocValues<?> get(Object key) {
...
+    }
+
+    parentLookup.setDocId(parentDocId);
+    return parentLookup;
+  }
+
+  /**
+   * Calculates the parent document ID using NESTED_DOCUMENT_OFFSET.
+   *
+   * @return parent document ID, or -1 if not found or not a nested document
```
waziqi89: why couldn't this be an explicit exception?
The request shall fail if the _parent notation is used wrongly.
Internally in nrtsearch itself, I don't see a reason why we shall possibly get parent when it doesn't exist

src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
```diff
@@ -68,7 +73,7 @@ public boolean isEmpty() {
 
   /**
    * Check if a given field name is capable of having doc values. This does not mean there is data
-   * present, just that there can be.
+   * present, just that there can be. Handles "_PARENT." prefix for parent field access.
```
waziqi89, line 77: maybe create it as a static final var

src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
```diff
@@ -54,6 +57,8 @@ public SegmentDocLookup(Function<String, FieldDef> fieldDefLookup, LeafReaderCon
    */
   public void setDocId(int docId) {
     this.docId = docId;
+    this.parentDocId = -1;
```
waziqi89, line 61: Why do we store parentDocId? It's not used anywhere independently, and it's private so inference cannot used it, either.

dhruva-r: @waziqi89 I made a mistake here: `this.parentLookup` should not be set to `null`. Instead the old parent lookup object is reused so that we are not creating an extra lookup per document. Instead I would create the lookup once and reset the id per document, then when the parent lookup is called, it will set the doc id of the parent and then return the lookup. This behaviour is already in `getParentLookup()`

src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
dhruva-r: I changed it to just return exceptions and wrapped it upstream to the get() method

src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
dhruva-r: This mistake was fixed on latest commit

waziqi89 (dismissed):

src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
```diff
@@ -128,9 +154,68 @@ public LoadedDocValues<?> get(Object key) {
...
+   * @throws IllegalArgumentException if a parent docId cannot be found or does not exist
+   */
+  private int getParentDocId() {
+    FieldDef offsetFieldDef;
+    try {
+      offsetFieldDef = IndexState.getMetaField(IndexState.NESTED_DOCUMENT_OFFSET);
+    } catch (IllegalArgumentException e) {
+      throw new IllegalArgumentException(
+          "Document is not a nested document - no parent document available", e);
+    }
```
sarthakn7: This is just getting the FieldDef, I don't think try catch is required here. You are always creating this field, not just when you have nested docs.

src/main/java/com/yelp/nrtsearch/server/doc/SegmentDocLookup.java
```diff
@@ -128,9 +154,68 @@ public LoadedDocValues<?> get(Object key) {
...
+
+    if (!(offsetFieldDef instanceof IndexableFieldDef<?> offsetIndexableFieldDef)) {
+      throw new IllegalArgumentException("NESTED_DOCUMENT_OFFSET field is not indexable");
+    }
```
sarthakn7: This check also seems redundant since the field is not created by a user but is inside nrtsearch and we know that it's an instance of IndexableFieldDef. You can directly get the field and cast it to IndexableFieldDef.

PR #883, opened 2025-09-16
Add acorn filter strategy

aprudhomme:
This is for the `development` branch. Allow the option to set the strategy to use for vector search when a filter is specified. There are two options:
- `FANOUT`: All neighbor vectors are check, only those that pass the filter are kept
- `ACORN`: Only neighbors that pass the filter are checked

Note the threshold value of 60 is the same as used by ES. Also, some tests were generated by AI.


src/main/java/com/yelp/nrtsearch/server/field/VectorFieldDef.java
```diff
@@ -243,6 +245,15 @@ private static KnnVectorsFormat getHnswBinaryQuantizedVectorsFormat(
         m, efConstruction, mergeWorkers, executorService);
   }
 
+  @VisibleForTesting
+  static KnnSearchStrategy createKnnSearchStrategy(KnnQuery.FilterStrategy filterStrategy) {
+    return switch (filterStrategy) {
+      case DEFAULT -> KnnSearchStrategy.Hnsw.DEFAULT;
+      case ACORN -> new KnnSearchStrategy.Hnsw(60);
```
sarthakn7: `KnnSearchStrategy.Hnsw.DEFAULT` is basically `new KnnSearchStrategy.Hnsw(60)`:
https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/knn/KnnSearchStrategy.java#L30C27-L30C60

Is this expected?

aprudhomme: It is actually 0 in 10.2.2 https://github.com/apache/lucene/blob/releases/lucene/10.2.2/lucene/core/src/java/org/apache/lucene/search/knn/KnnSearchStrategy.java#L30 Given that they are changing the default, I will make it explicitly 0.

sarthakn7: Hmm weird that it's different in the main branch

sarthakn7 (approved):

PR #891, opened 2025-10-02
Add delete support for ingestion plugins

umeshdangat:
- Add deleteByQuery() method to Ingestor interface and AbstractIngestor
- Add getIdFieldDef() helper to expose index ID field metadata to plugins
- Support TermInSetQuery for type-safe deletes (LONG, INT, TEXT fields)
- Fix S3Mock to use file-based backend instead of in-memory for large plugin tests

Tested this on paimon-ingestion plugin

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/ingestion/AbstractIngestor.java
+++ b/src/main/java/com/yelp/nrtsearch/server/ingestion/AbstractIngestor.java
@@ -16,12 +16,17 @@
 package com.yelp.nrtsearch.server.ingestion;
 
 import com.yelp.nrtsearch.server.config.NrtsearchConfig;
+import com.yelp.nrtsearch.server.field.IdFieldDef;
 import com.yelp.nrtsearch.server.grpc.AddDocumentRequest;
+import com.yelp.nrtsearch.server.grpc.Query;
 import com.yelp.nrtsearch.server.handler.AddDocumentHandler;
 import com.yelp.nrtsearch.server.index.IndexState;
+import com.yelp.nrtsearch.server.index.ShardState;
+import com.yelp.nrtsearch.server.query.QueryNodeMapper;
 import com.yelp.nrtsearch.server.state.GlobalState;
 import java.io.IOException;
 import java.util.List;
+import java.util.Optional;
 
 /**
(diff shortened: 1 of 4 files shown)


src/main/java/com/yelp/nrtsearch/server/ingestion/AbstractIngestor.java
```diff
@@ -61,6 +66,59 @@ public long addDocuments(List<AddDocumentRequest> addDocRequests, String indexNa
...
+   * before any subsequent addDocuments calls. This ordering is critical for UPDATE operations where
+   * the old document must be deleted before the new version is added.
+   *
+   * @param queries list of queries to match documents for deletion
+   * @param indexName target index name
+   * @return sequence number of the delete operation
+   * @throws Exception if delete fails
+   */
+  @Override
+  public long deleteByQuery(List<Query> queries, String indexName) throws Exception {
```
sarthakn7, line 86: Does an ingestion plugin need the flexibility to delete by query? Can we just provide a list of ids to delete and build the query here?

src/main/java/com/yelp/nrtsearch/server/ingestion/AbstractIngestor.java
```diff
@@ -61,6 +66,59 @@ public long addDocuments(List<AddDocumentRequest> addDocRequests, String indexNa
...
+    verifyInitialized(indexName);
+    IndexState indexState = globalState.getIndexOrThrow(indexName);
+    ShardState shardState = indexState.getShard(0);
```
AshishKhatkar, line 89: Why always 0? Should this not be based on what shard is currently active?

sarthakn7: Nrtsearch works with only one lucene shard per cluster and parallelism is achieved using parallel segment search and by sharding across clusters.

src/main/java/com/yelp/nrtsearch/server/ingestion/AbstractIngestor.java
umeshdangat: this would be nicer since it allows custom deletion in case of adhoc delete requests to be configured and at the same time supports delete by ids as well

PR #892, opened 2025-10-03
Add rate limiting option for s3 downloading

aprudhomme:
Adds the ability to throttle S3 downloading to a target rate. This is done by wrapping the `InputStream` and using a global rate limiter that looks at the amount of data read within a given time window.

I have restructured the configuration, and moved the 's3Metrics' options. The s3 backend options are new defined like:
```
remoteConfig:
  s3:
    metrics: true
    rateLimitPerSecond: 10MB
    rateLimitWindowSeconds: 5
``` 

`rateLimitPerSecond`: may be an absolute value of bytes, or use a kb/mb/gb suffix (default: 0, no limit)
`rateLimitWindowSeconds`: size of the window to enforce the limit over (default: 1)

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/utils/GlobalThrottledInputStream.java
+++ b/src/main/java/com/yelp/nrtsearch/server/utils/GlobalThrottledInputStream.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2025 Yelp Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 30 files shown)


src/main/java/com/yelp/nrtsearch/server/utils/GlobalThrottledInputStream.java
```diff
@@ -0,0 +1,46 @@
...
+   * @param in the InputStream to be wrapped
+   * @param limiter the GlobalWindowRateLimiter to use for throttling
+   */
+  public GlobalThrottledInputStream(InputStream in, GlobalWindowRateLimiter limiter) {
+    super(in);
+    this.limiter = limiter;
+  }
+
+  @Override
+  protected void afterRead(int n) {
```
sarthakn7: Nit: rename n to `numBytes`

sarthakn7 (dismissed):

PR #946, opened 2026-03-30
Ranking Parent Documents with Nested Features

dhruva-r:
## Summary of Changes
  - Add _CHILDREN. prefix support in SegmentDocLookup, allowing scoring plugins running on parent documents to access aggregated doc values from their child (nested) documents (e.g., _CHILDREN.appointments.price returns all appointment prices across a parent's children)
  - Use parentBitSet.prevSetBit(parentDocId - 1) to find child document ranges instead of storing a per-document _nested_child_count field - this avoids per-parent doc values allocation and is ~4.5x faster for child range detection (benchmarked)
  - Cache BitSetProducer instances on IndexState (singleton for root, ConcurrentHashMap for per-path) so they're reused across requests
  - Filter children by nested path using BitSetProducer, so indexes with multiple nested types (e.g., appointments and reviews) return only values from the target path
  - Add ChildFilter proto message on SearchRequest (field 34), enabling query-time filtering of which children are included in _CHILDREN. aggregation
  - Use ConcurrentHashMap for per-request path filter cache to avoid ConcurrentModificationException under Lucene's parallel search slices

## Tests


src/main/java/com/yelp/nrtsearch/server/doc/ChildAggregatedDocValues.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    // A child path filter was provided but produced no matches in this segment.
+    // This happens when QueryBitSetProducer returns null for an empty result set.
+    // In this case, no children should be included.
+    if (hasChildPathFilter && childPathBitSet == null) {
+      return;
+    }
+
+    LoadedDocValues<?> childCountDocValues = childCountFieldDef.getDocValues(leafContext);
```
vim345: getDocValues(leafContext) is called on every setDocId invocation rather than once in the constructor (as is done for childFieldDocValues). This allocates a new accessor on every parent document scored. It should be initialized once in the constructor and reused like childFieldDocValues.

Is it possible for you to do some benchmarking to see what the performance looks like with or without the above suggestion?

vim345 (commented):
Overall LGTM. I just had one comment.

Also, AI mentioned the following about missing test coverage:

```
 What's missing:

  1. No test for the issue #1 bug path — there's no test that verifies the _nested_child_count doc values accessor is reused across multiple documents in the same segment (i.e. no performance regression test, and the bug wouldn't cause a test failure — just extra allocations).
  2. No test for a parent with zero children — the childCount <= 0 early-return path in setDocId is untested.
  3. No test for documents without nested fields — i.e. calling _CHILDREN. on a non-parent document in a mixed index.
  4. No test for the lastParentDocId caching guard — the early-return if (parentDocId == lastParentDocId) return; is never exercised by a test that calls setDocId twice with the same doc ID.
  5. ChildFilterIntegrationTest uses scores to encode results rather than fetching actual field values, which makes assertions somewhat fragile and hard to read — though this is a style concern rather than a correctness gap.
```

Is it possible to address these as well?

src/main/java/com/yelp/nrtsearch/server/doc/ChildAggregatedDocValues.java
dhruva-r: Yes I can run some benchmarks to see what the performance looks like

dhruva-r: I was also able to add the additional testing coverage that was requested

clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -830,6 +830,21 @@ message SearchRequest {
     map<string, string> additionalOptions = 32;
     // When Multi-Retriever is set, top level query, queryText and knn fields should not be set and will error
     MultiRetrieverRequest multiRetriever = 33;
+    // Filters applied to child documents when accessed via _CHILDREN. prefix in scripts.
```
waziqi89, line 833: when updating the protobuf, I think you need to run ./gradlew grpcGateWay to update the gateway in the PR

src/main/java/com/yelp/nrtsearch/server/index/IndexState.java
```diff
@@ -95,6 +101,10 @@ public abstract class IndexState implements Closeable {
   private final ExecutorService searchExecutor;
   private Warmer warmer = null;
 
+  private volatile BitSetProducer cachedParentBitSetProducer;
```
waziqi89, line 104: According to CC, BitSetProducer will automatically handle the lifetime of segment add and merge. Cool.

waziqi89 (commented):
I am not fully aware of the use case of this proj, but feature wise LGTM.

Since now we have the parent child bitset cache, the symmetric implementation of childAggregatedDocValues (right-join) shall be able to replace the current nested document lookup using child count (left-join). Do we want to follow up on that? https://fluffy.yelpcorp.com/i/hV613lrc2fvNxfMn9k5x00H1dLbKgfnC.html

clientlib/src/main/proto/yelp/nrtsearch/search.proto
dhruva-r: I forgot to run this, I'll do that now

dhruva-r:
Just waiting for the 🟢 and will squash and merge to development branch

waziqi89 (approved):

PR #958, opened 2026-04-14
Add blender framework for multi-retriever search

waziqi89:
[RP-14960](https://jira.yelpcorp.com/browse/RP-14960) - blender interface
[RP-14962](https://jira.yelpcorp.com/browse/RP-14962) - RRF impl

Introduces the blender layer that merges per-retriever TopDocs into a single ranked result set:
- Create dedicated multiretriever packages to include all multiretriever related classes under search
- Create BlenderPlugin and BlenderOperation to support extensible blenders.
- Create builtin weightRRF Blender Operation

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/plugins/BlenderPlugin.java
+++ b/src/main/java/com/yelp/nrtsearch/server/plugins/BlenderPlugin.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Yelp Inc.
+ * Copyright 2026 Yelp Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/main/java/com/yelp/nrtsearch/server/plugins/BlenderPlugin.java
+++ b/src/main/java/com/yelp/nrtsearch/server/plugins/BlenderPlugin.java
@@ -16,6 +16,7 @@
 package com.yelp.nrtsearch.server.plugins;
 
 import com.yelp.nrtsearch.server.search.multiretriever.blender.BlenderOperation;
+import com.yelp.nrtsearch.server.search.multiretriever.blender.BlenderProvider;
 import java.util.Collections;
 import java.util.Map;
 
@@ -25,14 +26,10 @@ import java.util.Map;
  * com.yelp.nrtsearch.server.grpc.PluginBlender} message as part of the {@link
  * com.yelp.nrtsearch.server.grpc.Blender} in a {@link
  * com.yelp.nrtsearch.server.grpc.MultiRetrieverRequest}.
- *
- * <p>Implementations should be stateless; any per-request parameters are available at execution
- * time via the {@link com.yelp.nrtsearch.server.grpc.Blender} proto passed to {@link
- * BlenderOperation#mergeHits}.
  */
 public interface BlenderPlugin {
-  /** Get map of blender name to {@link BlenderOperation} instance for registration. */
-  default Map<String, BlenderOperation> getBlenders() {
+  /** Get map of blender name to {@link BlenderProvider} for registration. */
+  default Map<String, BlenderProvider<? extends BlenderOperation>> getBlenders() {
     return Collections.emptyMap();
   }
 }
--- a/src/main/java/com/yelp/nrtsearch/server/plugins/BlenderPlugin.java
+++ b/src/main/java/com/yelp/nrtsearch/server/plugins/BlenderPlugin.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2020 Yelp Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.yelp.nrtsearch.server.plugins;
+
+import com.yelp.nrtsearch.server.search.multiretriever.blender.BlenderOperation;
+import java.util.Collections;
+import java.util.Map;
+
+/**
+ * Plugin interface for providing custom {@link BlenderOperation}s. Provides info for registration
+ * of blenders by name. These blenders can be invoked by sending a {@link
+ * com.yelp.nrtsearch.server.grpc.PluginBlender} message as part of the {@link
+ * com.yelp.nrtsearch.server.grpc.Blender} in a {@link
+ * com.yelp.nrtsearch.server.grpc.MultiRetrieverRequest}.
+ *
+ * <p>Implementations should be stateless; any per-request parameters are available at execution
+ * time via the {@link com.yelp.nrtsearch.server.grpc.Blender} proto passed to {@link
+ * BlenderOperation#mergeHits}.
+ */
+public interface BlenderPlugin {
+  /** Get map of blender name to {@link BlenderOperation} instance for registration. */
+  default Map<String, BlenderOperation> getBlenders() {
+    return Collections.emptyMap();
+  }
+}
--- a/src/main/java/com/yelp/nrtsearch/server/search/multiretriever/blender/BlenderOperation.java
+++ b/src/main/java/com/yelp/nrtsearch/server/search/multiretriever/blender/BlenderOperation.java
@@ -1,0 +1,150 @@
+/*
+ * Copyright 2020 Yelp Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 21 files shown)


src/main/java/com/yelp/nrtsearch/server/search/multiretriever/blender/score/WeightedScoreDoc.java
```diff
@@ -0,0 +1,75 @@
...
+ * raw score is multiplied by its weight before being combined according to the configured {@link
+ * ScoreMode}:
+ *
+ * <ul>
+ *   <li>{@link ScoreMode#MAX} — highest weighted score across retrievers (default).
+ *   <li>{@link ScoreMode#SUM} — sum of all weighted retriever scores.
+ *   <li>{@link ScoreMode#AVG} — running average of all weighted retriever scores.
+ * </ul>
+ */
+public class WeightedScoreDoc extends BlendedScoreDoc {
```
waziqi89, line 31: This is used for test only until we have Score-Blender-Operation

src/main/java/com/yelp/nrtsearch/server/search/multiretriever/blender/BlenderOperation.java
```diff
@@ -0,0 +1,150 @@
...
+   */
+  static TopDocs sortAndPaginate(Collection<BlendedScoreDoc> merged, int startHit, int topHits) {
+    int total = merged.size();
+    if (topHits == 0) {
+      return new TopDocs(
+          new TotalHits(total, TotalHits.Relation.GREATER_THAN_OR_EQUAL_TO), new ScoreDoc[0]);
+    }
+    // Heap capacity is bounded by topHits to avoid materializing docs outside the window.
+    // `total` is preserved as the true deduplicated hit count for TotalHits reporting.
+    int capacity = (topHits > 0 && topHits < total) ? topHits : total;
```
swethakann: Does this work when startHit is > 0 though.

src/main/java/com/yelp/nrtsearch/server/search/multiretriever/blender/operation/WeightedRrfBlenderOperation.java
```diff
@@ -0,0 +1,79 @@
+/*
+ * Copyright 2020 Yelp Inc.
```
swethakann: Nit: Copyright 2026 here and in BlenderCreator.java

src/main/java/com/yelp/nrtsearch/server/search/multiretriever/blender/operation/WeightedRrfBlenderOperation.java
```diff
@@ -0,0 +1,79 @@
...
+    int k =
+        rrfConfig.getRankConstant() > 0
+            ? rrfConfig.getRankConstant()
+            : WeightedRRFScoreDoc.DEFAULT_K;
+
+    Map<Integer, BlendedScoreDoc> merged = new HashMap<>();
+
+    for (Map.Entry<String, TopDocs> entry : retrieverResults.entrySet()) {
+      ScoreDoc[] scoreDocs = entry.getValue().scoreDocs;
+      float boost = retrieverContexts.get(entry.getKey()).getBoost();
```
swethakann, line 61: maybe add a check for the key here and throw exception if it doesn't exist?

waziqi89: Do you mean `retrieverContexts.get` might get keyError?
I would image in proper implementation: retrieverResults' keys is a subset of retrieverContexts', if is not equal

src/main/java/com/yelp/nrtsearch/server/search/multiretriever/blender/operation/WeightedRrfBlenderOperation.java
waziqi89: Yes, let me fix that. Claude does it arbitrarily..

src/main/java/com/yelp/nrtsearch/server/search/multiretriever/blender/BlenderOperation.java
waziqi89: updated this

src/main/java/com/yelp/nrtsearch/server/search/multiretriever/blender/BlenderOperation.java
```diff
@@ -0,0 +1,150 @@
...
+   */
+  static TopDocs sortAndPaginate(Collection<BlendedScoreDoc> merged, int startHit, int topHits) {
+    int total = merged.size();
+    if (startHit > topHits || topHits <= 0) {
+      return new TopDocs(
+          new TotalHits(total, TotalHits.Relation.GREATER_THAN_OR_EQUAL_TO), new ScoreDoc[0]);
+    }
+    // Heap capacity is bounded by topHits to avoid materializing docs outside the window.
+    // `total` is preserved as the true deduplicated hit count for TotalHits reporting.
+    int capacity = Math.min(topHits, total);
```
swethakann, line 120: Don't we need to set capacity to startHit + topHits so the heap captures enough docs after the skip? 

Other changes look good to me. You can merge after resolving any comments after asp's review + ensuring this has a good test case to check for this scenario.

waziqi89: https://github.com/Yelp/nrtsearch/blob/main/src/main/java/com/yelp/nrtsearch/server/search/collectors/DocCollector.java#L80

topHits is the cap regardless of the startHit

src/main/java/com/yelp/nrtsearch/server/search/multiretriever/blender/BlenderOperation.java
```diff
@@ -0,0 +1,150 @@
...
+   * @param topHits maximum number of hits to return; {@code 0} returns empty; negative values mean
+   *     no limit
+   * @return paginated {@link TopDocs}
+   */
+  static TopDocs sortAndPaginate(Collection<BlendedScoreDoc> merged, int startHit, int topHits) {
+    int total = merged.size();
+    if (startHit > topHits || topHits <= 0) {
+      return new TopDocs(
+          new TotalHits(total, TotalHits.Relation.GREATER_THAN_OR_EQUAL_TO), new ScoreDoc[0]);
+    }
```
aprudhomme: I feel like this validation should be happening a lot further up in the stack, to avoid unneeded work.

src/main/java/com/yelp/nrtsearch/server/search/multiretriever/blender/BlenderOperation.java
```diff
@@ -0,0 +1,150 @@
...
+   * paginated by the caller.
+   *
+   * @param retrieverResults per-retriever {@link TopDocs} in declaration order, keyed by name
+   * @param blender blender proto config
+   * @param retrieverContexts per-retriever contexts in declaration order, keyed by name
+   * @return merged hits, unsorted and unpaginated
+   */
+  Collection<BlendedScoreDoc> mergeHits(
+      LinkedHashMap<String, TopDocs> retrieverResults,
+      com.yelp.nrtsearch.server.grpc.Blender blender,
```
aprudhomme: I would suggest not passing the blender as a parameter. This information is static for the query, and could be used during construction to create a stateful instance per query. The advantage of this is that validation could be done up front, avoiding work when there is a failure.

src/main/java/com/yelp/nrtsearch/server/plugins/BlenderPlugin.java
```diff
@@ -0,0 +1,38 @@
...
+ * com.yelp.nrtsearch.server.grpc.Blender} in a {@link
+ * com.yelp.nrtsearch.server.grpc.MultiRetrieverRequest}.
+ *
+ * <p>Implementations should be stateless; any per-request parameters are available at execution
+ * time via the {@link com.yelp.nrtsearch.server.grpc.Blender} proto passed to {@link
+ * BlenderOperation#mergeHits}.
+ */
+public interface BlenderPlugin {
+  /** Get map of blender name to {@link BlenderOperation} instance for registration. */
+  default Map<String, BlenderOperation> getBlenders() {
```
aprudhomme: Going along with my other comment. The plugin could provide a factory for producing per query instances of the `BlenderOperation`. We do something similar for other components like https://github.com/Yelp/nrtsearch/blob/main/src/main/java/com/yelp/nrtsearch/server/rescore/RescorerProvider.java

waziqi89: yes. In my original design, a provider was provided. But RRF seems doesn't need it, so I removed it. Revert back

src/main/java/com/yelp/nrtsearch/server/search/multiretriever/blender/BlenderOperation.java
waziqi89: The main motivation is that the param for predefined blender like RRF are not under general params, but under it's own def.
Referred to the queryRescorer now, the param is processed and passed in Processor

src/main/java/com/yelp/nrtsearch/server/search/multiretriever/blender/BlenderOperation.java
waziqi89: Understand that this can be done earlier by skipping search execution on retrievers.

This is slightly tricky to check and skip as the knn rewrites happens in Processor and we need a rewritten knn query for union query generation.

The details can be finalized in the following up PRs for changes in searchHandler, I just moved this to blend() to make it slightly earlier for now.

waziqi89: Another piece of thought is that: topHits = 0 is actually a nonsense use case for multiretriever. We may just check and throw exceptions in Processor instead of returning empty. This is much simpler to implement.

swethakann (approved):

aprudhomme (approved):

PR #969, opened 2026-05-04
Add Search Handling for Multi-Retriever Requests

swethakann:
File changes:

**SearchRequestProcessor.java** — Removed the UnsupportedOperationException blocker. Added explicit validation rejecting facets and collectors with multi-retriever. This will be implemented in future PR.                                                                                                                                                    
                                                                                                                                                                                     
**SearchHandler.java**                                                                                                                        
  - Submits per-retriever searcher.search() calls to searchExecutor in parallel
  - Applies per-retriever L1 rescorers inside each callable                                                                                                                          
  - Computes a blend window large enough for downstream L2 rescorers and hits-to-log
  - Calls blenderOperation.blend() for all blender types (RRF, WeightedScoreOrder, ScorelessRawMerge, Plugin)                                                                        
  - Sets blended scores on Hit via direct ScoreDoc.score (bypassing the no-op HitCountCollector)                                                                                     
  - Populates Hit.retrieverScores from BlendedScoreDoc.scoreDocs per-retriever raw scores                                                                                            
  - Records per-retriever search timings and blender timing in MultiRetrieverDiagnostics                                                                                             
  - Calls maybeAddProfiling() on per-retriever collectors when profile=true                                                                                                            
                                                                                                                                                                                     
**RescoreTask.java** — Added getWindowSize() accessor needed by executeMultiRetriever.                                                                                                 
                                                                                                                                                                                     
**ServerTestCase.java** — Added getSearchRequestFromResourceFile() helper for JSON-driven tests.                                                                                                                                                                                                                                   

**MultiRetrieverSearchTest.java** — 15 tests covering validation, all three blender types, diagnostics, profiling, and a JSON-driven end-to-end test.

**MultiRetrieverRrfRequest.json** — JSON fixture representing a real client request.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
+++ b/src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
@@ -537,10 +537,9 @@ public class SearchHandler extends Handler<SearchRequest, SearchResponse> {
     LinkedHashMap<String, RetrieverContext> retrieverContexts =
         new LinkedHashMap<>(multiRetrieverContext.getRetrieverContextMap());
 
-    record RetrieverResult(
-        TopDocs topDocs, long totalHits, double searchTimeMs, double rescoreTimeMs) {}
+    record RetrieverMetrics(TopDocs topDocs, double searchTimeMs, double rescoreTimeMs) {}
 
-    ConcurrentHashMap<String, RetrieverResult> retrieverResults = new ConcurrentHashMap<>();
+    ConcurrentHashMap<String, RetrieverMetrics> retrieverResults = new ConcurrentHashMap<>();
     LinkedHashMap<String, Future<TopDocs>> retrieverFutures = new LinkedHashMap<>();
     for (Map.Entry<String, RetrieverContext> entry : retrieverContexts.entrySet()) {
       String name = entry.getKey();
@@ -556,7 +555,6 @@ public class SearchHandler extends Handler<SearchRequest, SearchResponse> {
                         retrieverContext.getDocCollector().getWrappedManager());
                 TopDocs topDocs = result.getTopDocs();
                 double searchTimeMs = (System.nanoTime() - searchStart) / 1_000_000.0;
-                long totalHits = topDocs.totalHits.value();
 
                 double rescoreTimeMs = 0;
                 if (retrieverContext.getRescoreTask() != null) {
@@ -565,7 +563,7 @@ public class SearchHandler extends Handler<SearchRequest, SearchResponse> {
                   rescoreTimeMs = (System.nanoTime() - rescoreStart) / 1_000_000.0;
                 }
                 retrieverResults.put(
-                    name, new RetrieverResult(topDocs, totalHits, searchTimeMs, rescoreTimeMs));
+                    name, new RetrieverMetrics(topDocs, searchTimeMs, rescoreTimeMs));
                 return topDocs;
               }));
     }
@@ -589,24 +587,25 @@ public class SearchHandler extends Handler<SearchRequest, SearchResponse> {
     // Populate per-retriever diagnostics
     SearchResponse.Diagnostics.MultiRetrieverDiagnostics.Builder multiRetrieverDiagnosticsBuilder =
         diagnostics.getMultiRetrieverDiagnosticsBuilder();
-    for (Map.Entry<String, RetrieverResult> entry : retrieverResults.entrySet()) {
+    for (Map.Entry<String, RetrieverMetrics> entry : retrieverResults.entrySet()) {
       String name = entry.getKey();
-      RetrieverResult retrieverResult = entry.getValue();
+      RetrieverMetrics retrieverResult = entry.getValue();
       RetrieverContext.RetrieverType type = retrieverContexts.get(name).getRetrieverType();
+      org.apache.lucene.search.TotalHits luceneTotalHits = retrieverResult.topDocs().totalHits;
       TotalHits totalHits =
           TotalHits.newBuilder()
-              .setRelation(
-                  TotalHits.Relation.valueOf(retrieverResult.topDocs().totalHits.relation().name()))
-              .setValue(retrieverResult.totalHits())
+              .setRelation(TotalHits.Relation.valueOf(luceneTotalHits.relation().name()))
+              .setValue(luceneTotalHits.value())
               .build();
       SearchResponse.Diagnostics.RetrieverDiagnostics.Builder retrieverDiagBuilder;
       if (type == RetrieverContext.RetrieverType.KNN) {
-        // Preserve vectorDiagnostics already set by SearchRequestProcessor
+        // Preserve vectorDiagnostics already set by SearchRequestProcessor, then add timing
         retrieverDiagBuilder =
             multiRetrieverDiagnosticsBuilder
                 .getRetrieverDiagnosticsOrDefault(
                     name, SearchResponse.Diagnostics.RetrieverDiagnostics.getDefaultInstance())
-                .toBuilder();
+                .toBuilder()
+                .setSearchTimeMs(retrieverResult.searchTimeMs());
       } else {
         retrieverDiagBuilder =
             SearchResponse.Diagnostics.RetrieverDiagnostics.newBuilder()
--- a/src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
+++ b/src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
@@ -50,6 +50,9 @@ import com.yelp.nrtsearch.server.search.SearchContext;
 import com.yelp.nrtsearch.server.search.SearchCutoffWrapper.CollectionTimeoutException;
 import com.yelp.nrtsearch.server.search.SearchRequestProcessor;
 import com.yelp.nrtsearch.server.search.SearcherResult;
+import com.yelp.nrtsearch.server.search.multiretriever.MultiRetrieverContext;
+import com.yelp.nrtsearch.server.search.multiretriever.RetrieverContext;
+import com.yelp.nrtsearch.server.search.multiretriever.blender.score.BlendedScoreDoc;
 import com.yelp.nrtsearch.server.state.GlobalState;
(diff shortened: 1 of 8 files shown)


src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
```diff
@@ -512,6 +522,104 @@ private void fetchFields(SearchContext searchContext)
...
+                long t0 = System.nanoTime();
+                SearcherResult result =
+                    searcher.search(
+                        retrieverContext.getQuery(),
+                        retrieverContext.getDocCollector().getWrappedManager());
+                TopDocs topDocs = result.getTopDocs();
+                if (retrieverContext.getRescoreTask() != null) {
+                  topDocs = retrieverContext.getRescoreTask().rescore(topDocs, searchContext);
+                }
+                retrieverTimingsMs.put(name, (System.nanoTime() - t0) / 1_000_000.0);
```
waziqi89: For retriever timing, I prefer to keep this sync-ed with the status quo: having firstPassSearchTime and RescorerTime separate.

src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
```diff
@@ -512,6 +522,104 @@ private void fetchFields(SearchContext searchContext)
...
+                return topDocs;
+              }));
+    }
+
+    // Compute the blend window to cover rescorer windows and hits
+    // L2 rescorer then trims to the final topHits window.
+    int blendTopHits = searchContext.getStartHit() + searchContext.getTopHits();
```
waziqi89: L2 rescorer is universal, and shall be handled by the [main search flow](https://github.com/Yelp/nrtsearch/blob/main/src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java#L262) **after** blending.

I don't expect any L2 rescorer appear in multiretriever-blender step

src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
```diff
@@ -512,6 +522,104 @@ private void fetchFields(SearchContext searchContext)
...
+    for (RescoreTask rescorer : searchContext.getRescorers()) {
+      int windowSize = rescorer.getWindowSize();
+      if (windowSize > blendTopHits) {
+        blendTopHits = windowSize;
+      }
+    }
+    int hitsToLog = searchContext.getHitsToLog();
```
waziqi89: Same for hitsToLog, I don't think we want to log hits before blender. This is no such request for now; Even if, we may handle them differently

src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
```diff
@@ -512,6 +522,104 @@ private void fetchFields(SearchContext searchContext)
...
+    double blenderTimeMs = (System.nanoTime() - blendStartTime) / 1_000_000.0;
+
+    // Merge per-retriever timings into diagnostics
+    SearchResponse.Diagnostics.MultiRetrieverDiagnostics.Builder multiRetrieverDiagnosticsBuilder =
+        diagnostics.getMultiRetrieverDiagnosticsBuilder();
+    for (Map.Entry<String, Double> timingEntry : retrieverTimingsMs.entrySet()) {
+      String name = timingEntry.getKey();
+      SearchResponse.Diagnostics.RetrieverDiagnostics existing =
+          multiRetrieverDiagnosticsBuilder.getRetrieverDiagnosticsOrDefault(name, null);
+      SearchResponse.Diagnostics.RetrieverDiagnostics.Builder retrieverDiagnosticsBuilder =
```
waziqi89: Why can it be nonnull?
I image two cases:
1. multiple retriever with the identical name: invalid case - can fail at processor
2. this is resolved KNN query - then shall we "add" instead of overwrite? (rewrite takes the most time, search here is basically read, and you can ignore or add)

src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
```diff
@@ -512,6 +522,104 @@ private void fetchFields(SearchContext searchContext)
...
+          existing != null
+              ? existing.toBuilder()
```
waziqi89: Another Nit-picking question: can we just put builder all the way from processor? Then we don't have to "toBuilder" here

src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
```diff
@@ -512,6 +522,104 @@ private void fetchFields(SearchContext searchContext)
...
+    multiRetrieverDiagnosticsBuilder.setBlenderTimeMs(blenderTimeMs);
+
+    // Add per-retriever collector profiling stats
+    if (profileResultBuilder != null) {
+      ProfileResult.MultiRetrieverProfileResult.Builder multiRetrieverProfileBuilder =
+          profileResultBuilder.getMultiRetrieverProfileResultBuilder();
+      for (Map.Entry<String, RetrieverContext> entry : retrieverContexts.entrySet()) {
+        String name = entry.getKey();
+        ProfileResult.Builder retrieverProfile =
+            multiRetrieverProfileBuilder.getRetrieverProfileResultsOrDefault(name, null) != null
```
waziqi89: Same as diagnostics, wonder why it could be existing

src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
waziqi89: you may rename firstPassSearchTime into retrieverSearchTime. And add retrieverRescoreSearchTime

src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
waziqi89: The relevant tests shall be deleted, or replaced by some tests to ensure the existing rescoerer work as before

swethakann: The L2 rescorer doesn't run inside executeMultiRetriever — it runs in the main handle() flow after the method returns as usual. The sizing logic (blendTopHits accounting for rescorer window sizes) is because  if we don't request enough hits upfront, the post-blend rescorer won't have its full window to work with. This is similar to DocCollector.computeNumHitsToCollect accounting for rescorer windows in the non-multi-retriever path.

waziqi89: I see what you mean here. You are absolutely right.

Now I see the root problem: I over-simplified the hitsToCollect in mind. It shall be more complicated than just startHit, topHits. (uncompleted work from https://jira.yelpcorp.com/browse/RP-14770)

waziqi89: I think the solution is to decouple `DocCollector.computeNumHitsToCollect` from its argument `searchRequest` to the detailed numbers so that we can reuse the same logic to get the numToCollect for blender. 

I would say: keep `computeNumHitsToCollect(searchRequest)`, and add `computeNumHitsToCollect(startHits, topHits, logHits=0, rescorerList=null/empty)`

swethakann: Seems like a good idea. Made the changes

src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
```diff
@@ -548,33 +549,35 @@ private TopDocs executeMultiRetriever(
...
+
+                double rescoreTimeMs = 0;
                 if (retrieverContext.getRescoreTask() != null) {
+                  long rescoreStart = System.nanoTime();
                   topDocs = retrieverContext.getRescoreTask().rescore(topDocs, searchContext);
+                  rescoreTimeMs = (System.nanoTime() - rescoreStart) / 1_000_000.0;
                 }
-                retrieverTimingsMs.put(name, (System.nanoTime() - t0) / 1_000_000.0);
+                retrieverResults.put(
+                    name, new RetrieverResult(topDocs, totalHits, searchTimeMs, rescoreTimeMs));
```
waziqi89: topDocs contains the totalHits object. You don't have to pass a totalHits-value.
totalHits object contains both relation and value, and you used both in the following diagnostic parts.
IMO, this Record can be renamed as RetrieverStatistics/Metrics, and containing TotalHits-object, searchTimeMs, rescoreTimeMs

waziqi89: I have a feeling that my interface design: ```default TopDocs blend(
      LinkedHashMap<String, Future<TopDocs>> retrieverFutures,``` makes your record a bit awkward to implement.
If you do want to keep the record as `retrieverResult(topDocs, searchTimeMs, rescoreTimeMs)`, you can change the interface and accept `Map<name, Future<retrieverResult>>` instead.

waziqi89 (approved):

src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
```diff
@@ -512,6 +523,120 @@ private void fetchFields(SearchContext searchContext)
...
+    record RetrieverMetrics(TopDocs topDocs, double searchTimeMs, double rescoreTimeMs) {}
+
+    ConcurrentHashMap<String, RetrieverMetrics> retrieverResults = new ConcurrentHashMap<>();
+    LinkedHashMap<String, Future<TopDocs>> retrieverFutures = new LinkedHashMap<>();
+    for (Map.Entry<String, RetrieverContext> entry : retrieverContexts.entrySet()) {
+      String name = entry.getKey();
+      RetrieverContext retrieverContext = entry.getValue();
+      retrieverFutures.put(
+          name,
+          searchExecutor.submit(
```
aprudhomme, line 549: It looks like this version of lucene should prevent deadlocking when using the search executor https://github.com/apache/lucene/blob/releases/lucene/10.1.0/lucene/core/src/java/org/apache/lucene/search/TaskExecutor.java#L105 so it should be ok.

aprudhomme (approved):

PR #970, opened 2026-05-05
Add normalizer option for ATOM fields

aprudhomme:
---
  Add normalizer support for ATOM field type

  ATOM fields are single-token keyword fields that previously stored and indexed values exactly as provided — meaning "Hello" and "hello" were treated as different terms. This adds
  support for a normalizer field option on ATOM fields, analogous to Elasticsearch's keyword normalizer.

  A normalizer applies text transformations (char filters + token filters, with an implied keyword tokenizer) consistently at both index time and query time, ensuring that
  case-insensitive or otherwise normalized matching works transparently.

  What's included

  Proto changes:
  - New Normalizer message in analysis.proto — oneof predefined (string) or custom (CustomNormalizer)
  - New CustomNormalizer message — char filters + token filters (keyword tokenizer implied)
  - New normalizer field (field 38) on the Field message in luceneserver.proto

  Built-in predefined normalizer:
  - "lowercase" — lowercases the entire field value using Lucene's LowerCaseFilter

  Custom normalizer support:
  - Specify any combination of char filters and token filters via CustomNormalizer

  Plugin extension:
  - AnalysisPlugin.getNormalizers() default method allows plugins to register additional named normalizers

  AtomFieldDef normalization applied consistently at:
  - Index time: doc values, stored fields, and facets (via parseDocumentField override)
  - Query time: term queries, term-in-set queries, range queries, and prefix queries

  Backward compatible: fields without a normalizer behave identically to before.

  Example usage

```
  {
    "name": "status",
    "type": "ATOM",
    "search": true,
    "storeDocValues": true,
    "normalizer": {
      "predefined": "lowercase"
    }
  }
```

  With this field, indexing "Active", "ACTIVE", or "active" all produce the same indexed term "active", and querying with any casing will match.

  Test plan

  - AnalyzerCreatorTest — unit tests for isNormalizerDefined, getNormalizer (predefined/custom/unknown/plugin), and duplicate registration guard
  - AtomFieldDefTest — unit tests asserting normalized output from term query construction
  - AtomFieldNormalizerTest — end-to-end integration tests covering: term query, term-in-set query, range query, prefix query, stored field retrieval, multi-value fields, custom
  normalizer, and backward compatibility with non-normalized fields


vim345 (approved):

sarthakn7 (approved):
Looks good, but would be good to update docs as well

PR #976, opened 2026-05-12
Add ScriptRescorer to re-score blended results

waziqi89:
RP-15336

Add support for script rescorer to enable static or dynamic tweak on scores.
This is enabled by adapting Lucene DoubleValuesSource (core of script) to the RescoreOperation.
Theoretically, any double value script can be used immediately.

Test locally with JS script:
test code: https://fluffy.yelpcorp.com/i/pVwrqJv3J7xPJWbJ13k1FGrqhWTGfdLn.html
result: https://fluffy.yelpcorp.com/i/MDwpfQmLxkJNktXzbfBPHf3vx6sKCJK0.html

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/rescore/ScriptRescore.java
+++ b/src/main/java/com/yelp/nrtsearch/server/rescore/ScriptRescore.java
@@ -1,0 +1,122 @@
+/*
+ * Copyright 2026 Yelp Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.yelp.nrtsearch.server.rescore;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.search.DoubleValues;
+import org.apache.lucene.search.DoubleValuesSource;
+import org.apache.lucene.search.IndexSearcher;
+import org.apache.lucene.search.ScoreDoc;
+import org.apache.lucene.search.TopDocs;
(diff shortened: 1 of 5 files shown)


src/main/java/com/yelp/nrtsearch/server/rescore/ScriptRescore.java
```diff
@@ -0,0 +1,122 @@
...
+        scriptValues = scriptSource.getValues(leaf, scoreValues);
+      }
+
+      int segmentDocId = scoreDoc.doc - leaves.get(leafIndex).docBase;
+      scriptValues.advanceExact(segmentDocId);
+      scoreDoc.score = (float) scriptValues.doubleValue();
+    }
+
+    // Re-sort by new score descending.
+    Arrays.sort(scoreDocs, (a, b) -> Float.compare(b.score, a.score));
```
aprudhomme: You usually want to also use the doc id as a tiebreaker when the scores are the same. Lucene might already expose this comparator somewhere, otherwise, you can probably pull it from the rescore plugin.

src/main/java/com/yelp/nrtsearch/server/rescore/ScriptRescore.java
```diff
@@ -0,0 +1,122 @@
...
+      // Advance to the leaf that contains this global doc id.
+      while (leafIndex < leaves.size() - 1 && leaves.get(leafIndex + 1).docBase <= scoreDoc.doc) {
+        leafIndex++;
+        scriptValues = null; // entering a new segment
+      }
+
+      if (scriptValues == null) {
+        LeafReaderContext leaf = leaves.get(leafIndex);
+        // TopDocsScoreValues exposes the previous-pass score as _score to the script.
+        scoreValues = new TopDocsScoreValues(scoreDocs, leaf.docBase);
```
aprudhomme: I think exposing _score can be simplified. You can create a single instance of https://github.com/Yelp/nrtsearch/blob/main/src/main/java/com/yelp/nrtsearch/server/query/QueryUtils.java#L47 and set the value prior to each script run.

waziqi89: Claude code suggested to implement similarly as this initially, but I thought creating a wrapper like object makes code a bit clearer.
But as long as we've already had this util code base, I will use it instead.

src/main/java/com/yelp/nrtsearch/server/rescore/ScriptRescore.java
waziqi89: Checked the code. It's in Topdocs but private. I will replicate the logic here locally

aprudhomme (approved):

PR #982, opened 2026-05-25
Create scriptFactoryContext to wrap all related scoreScript.factory params, including a new one - sharedDocCotext

waziqi89:
⏺ RP-15633 (https://jira.yelpcorp.com/browse/RP-15633) Add retriever score side values for L2 ScriptRescorer

*Breaking Change on All Script Plugins*

- Creating scriptFactoryContext to convey all necessary contexts for script factory to avoid future changes.
- Add the retriever_score and pass to script factory through sharedDocContext, allowing easy access from the rescorers

Test passed local manual test:  https://fluffy.yelpcorp.com/i/QlQZks5Twn7rmZB6kzktnbL6VsVZJ25m.html

Once merged. Need to update the core image with migration for [all](https://sourcegraph.yelpcorp.com/search?q=repo:%5Ejava-packages/nrtsearch-plugin+scoreScript.factory&patternType=keyword&sm=0) related plugins:
- painless (to be verified)
- inference
- lucyscoring
- queryscore
- maptype (test only)

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
+++ b/src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
@@ -20,6 +20,7 @@ import com.google.common.util.concurrent.MoreExecutors;
 import com.google.protobuf.InvalidProtocolBufferException;
 import com.google.protobuf.util.JsonFormat.Printer;
 import com.yelp.nrtsearch.server.doc.LoadedDocValues;
+import com.yelp.nrtsearch.server.doc.SharedDocContext;
 import com.yelp.nrtsearch.server.facet.DrillSidewaysImpl;
 import com.yelp.nrtsearch.server.facet.FacetTopDocs;
 import com.yelp.nrtsearch.server.field.DateTimeFieldDef;
@@ -183,6 +184,7 @@ public class SearchHandler extends Handler<SearchRequest, SearchResponse> {
       if (searchContext.getMultiRetrieverContext() != null) {
         hits =
             executeMultiRetriever(searchContext, s.searcher(), diagnostics, profileResultBuilder);
+        populateRetrieverScores(hits, searchContext.getSharedDocContext());
       } else {
         SearcherResult searcherResult;
         if (!searchRequest.getFacetsList().isEmpty()) {
@@ -646,6 +648,23 @@ public class SearchHandler extends Handler<SearchRequest, SearchResponse> {
     }
 
     return blendedHits;
+  }
+
+  /**
+   * Write per-retriever scores from blended hits into the shared doc context. For retriever {@code
+   * "text"}, the key {@code "retriever_text"} is set on each document's context map, accessible in
+   * JS scripts as {@code _shared_retriever_text} and in ScoreScript subclasses via {@code
+   * get_shared_double("retriever_text", defaultValue)}.
+   */
+  static void populateRetrieverScores(TopDocs hits, SharedDocContext sharedDocContext) {
+    for (ScoreDoc scoreDoc : hits.scoreDocs) {
+      if (scoreDoc instanceof BlendedScoreDoc blended) {
+        Map<String, Object> ctx = sharedDocContext.getContext(scoreDoc.doc);
+        for (Map.Entry<String, ScoreDoc> entry : blended.getScoreDocs().entrySet()) {
+          ctx.put("retriever_" + entry.getKey(), (double) entry.getValue().score);
+        }
+      }
+    }
   }
 
   /**
--- a/src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
+++ b/src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
@@ -86,6 +86,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class SearchHandler extends Handler<SearchRequest, SearchResponse> {
+  public static final String RETRIEVER_KEY_PREFIX = "retriever_";
   private static final ExecutorService DIRECT_EXECUTOR = MoreExecutors.newDirectExecutorService();
   private static final Printer protoMessagePrinter =
       ProtoMessagePrinter.omittingInsignificantWhitespace();
@@ -661,7 +662,7 @@ public class SearchHandler extends Handler<SearchRequest, SearchResponse> {
       if (scoreDoc instanceof BlendedScoreDoc blended) {
         Map<String, Object> ctx = sharedDocContext.getContext(scoreDoc.doc);
         for (Map.Entry<String, ScoreDoc> entry : blended.getScoreDocs().entrySet()) {
-          ctx.put("retriever_" + entry.getKey(), (double) entry.getValue().score);
+          ctx.put(RETRIEVER_KEY_PREFIX + entry.getKey(), (double) entry.getValue().score);
         }
       }
     }
--- a/src/main/java/com/yelp/nrtsearch/server/query/QueryNodeMapper.java
+++ b/src/main/java/com/yelp/nrtsearch/server/query/QueryNodeMapper.java
@@ -44,6 +44,7 @@ import com.yelp.nrtsearch.server.index.IndexState;
 import com.yelp.nrtsearch.server.index.ShardState;
 import com.yelp.nrtsearch.server.query.multifunction.MultiFunctionScoreQuery;
 import com.yelp.nrtsearch.server.script.ScoreScript;
+import com.yelp.nrtsearch.server.script.ScriptFactoryContext;
 import com.yelp.nrtsearch.server.script.ScriptService;
 import com.yelp.nrtsearch.server.utils.ScriptParamsUtils;
 import java.io.IOException;
@@ -298,7 +299,8 @@ public class QueryNodeMapper {
         ScriptParamsUtils.decodeParams(functionScoreQuery.getScript().getParamsMap());
     return new FunctionScoreQuery(
         getQuery(functionScoreQuery.getQuery(), context),
-        scriptFactory.newFactory(params, context.docLookup()));
+        scriptFactory.newFactory(
+            ScriptFactoryContext.builder(params, context.docLookup()).build()));
   }
 
   private FunctionMatchQuery getFunctionFilterQuery(
@@ -307,7 +309,9 @@ public class QueryNodeMapper {
         ScriptService.getInstance().compile(functionFilterQuery.getScript(), ScoreScript.CONTEXT);
     Map<String, Object> params =
         ScriptParamsUtils.decodeParams(functionFilterQuery.getScript().getParamsMap());
-    return new FunctionMatchQuery(scriptFactory.newFactory(params, docLookup), score -> score > 0);
+    return new FunctionMatchQuery(
+        scriptFactory.newFactory(ScriptFactoryContext.builder(params, docLookup).build()),
+        score -> score > 0);
   }
 
   private Query getTermQuery(
--- a/src/main/java/com/yelp/nrtsearch/server/query/QueryNodeMapper.java
+++ b/src/main/java/com/yelp/nrtsearch/server/query/QueryNodeMapper.java
@@ -114,11 +114,11 @@ public class QueryNodeMapper {
               MatchOperator.MUST, BooleanClause.Occur.MUST));
 
   public Query getQuery(com.yelp.nrtsearch.server.grpc.Query query, IndexState state) {
-    return getQuery(query, new QueryContext(state.docLookup, state.getGlobalState()));
+    return getQuery(query, new QueryContext(state.docLookup, state.getGlobalState(), null));
   }
 
   public Query getQuery(com.yelp.nrtsearch.server.grpc.Query query, DocLookup docLookup) {
-    return getQuery(query, new QueryContext(docLookup, null));
+    return getQuery(query, new QueryContext(docLookup, null, null));
   }
 
   public Query getQuery(com.yelp.nrtsearch.server.grpc.Query query, QueryContext context) {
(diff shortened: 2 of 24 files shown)


src/main/java/com/yelp/nrtsearch/server/script/js/JsScriptBindings.java
```diff
@@ -168,4 +204,83 @@ public boolean advanceExact(int doc) {
...
+    }
+
+    @Override
+    public double doubleValue() {
+      return currentValue;
+    }
+
+    @Override
+    public boolean advanceExact(int doc) {
+      currentValue = sideValues.getDouble(fieldName, leafDocBase + doc, 0.0);
```
swethakann: Should this return false when the doc is absent?

src/main/java/com/yelp/nrtsearch/server/script/SideValues.java
```diff
@@ -0,0 +1,91 @@
...
+ * Values are organized by field name and then by global Lucene document ID.
+ *
+ * <p>For multi-retriever queries, {@code SearchHandler} populates one field per retriever using the
+ * key {@code retriever_<name>} (e.g. {@code retriever_text}, {@code retriever_knn}). Scripts access
+ * these via {@code get_side_value_double("retriever_<name>", defaultValue)}.
+ *
+ * <p>Multiple pipeline steps may each contribute their own field names via {@link
+ * #merge(SideValues)}, which adds new entries without overwriting fields already present.
+ */
+public class SideValues {
```
aprudhomme: There is already a construct for sharing information between search operations. It is the `SharedDocContext` stored in the `SearchContext`. Is there a reason that cannot be used, instead of creating something new?

src/main/java/com/yelp/nrtsearch/server/script/ScoreScript.java
```diff
@@ -133,6 +189,69 @@ public interface Factory {
...
+     * SegmentFactory#newInstance(LeafReaderContext, DoubleValues, SideValues)}. Script engines that
+     * expose side values differently (e.g. as named JS expression variables) should override this
+     * method instead.
+     *
+     * @param params parameters from script request
+     * @param docLookup index level doc value lookup provider
+     * @param sideValues per-document side values accumulated by pipeline steps before this rescore
+     * @return {@link DoubleValuesSource} to evaluate script
+     */
+    default DoubleValuesSource newFactory(
```
aprudhomme: Having multiple `newFactory` methods is going to potentially cause issues for some `ScriptEngine`s. See https://github.com/Yelp/nrtsearch/blob/main/src/main/java/com/yelp/nrtsearch/server/script/ScriptContext.java for expectations on script definition structure.

Changing this interface will likely require scripting plugins to be rebuilt. When we want to do this, rollout will need to be coordinated. It might be better to remove the `DocLookup` and pass some kind of `ScriptFactoryContext` instead, which contains all resources needed. This would avoid incompatible changes the next time we want to add something.

src/main/java/com/yelp/nrtsearch/server/script/SideValues.java
waziqi89: SharedDocContext is heavily used by inference allowing clients to update or even remove it arbitrarily which may lead to some unexpected behaviors. 

This sideValues can be renamed to something else but the main idea is for internal core to write and expose no setter to clients

```
  SideValues protections:
  1. Instances are immutable — the internal map is Collections.unmodifiableMap; no code holding a reference can mutate it.
  2. merge uses putIfAbsent — whoever writes first wins; a later pipeline step (e.g., a second retriever or plugin) cannot overwrite an already-present field.
  3. Scripts have no write path — ScoreScript only exposes get_side_value_double; there's no setter that touches SideValues.
```

waziqi89: thought twice one this: SQ is privileged user can could possibly overwrite the retriever_score as they mentioned "recompute similarity and word_score" in rescorer. We may not want to restrict them from overwriting

src/main/java/com/yelp/nrtsearch/server/script/ScriptFactoryContext.java
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Bundles all request-level resources that a {@link ScoreScript.Factory} may need to produce a
+ * {@link org.apache.lucene.search.DoubleValuesSource}. Using a single value object instead of
+ * individual parameters makes the {@link ScoreScript.Factory#newFactory} signature stable — adding
+ * resources in the future does not require changing the interface.
+ */
+public final class ScriptFactoryContext {
+
+  private final Map<String, Object> params;
```
aprudhomme, line 30: I'm not sure if the params should be included in the context, or be left separate. The params are a script level property, but everything else are query level. Separating them could make interfaces cleaner by letting the context be passed in place of the doc lookup.

src/main/java/com/yelp/nrtsearch/server/search/SearchRequestProcessor.java
```diff
@@ -334,7 +335,9 @@ private static Map<String, FieldDef> getVirtualFields(
           ScriptService.getInstance().compile(vf.getScript(), ScoreScript.CONTEXT);
       Map<String, Object> params = ScriptParamsUtils.decodeParams(vf.getScript().getParamsMap());
       FieldDef virtualField =
-          new VirtualFieldDef(vf.getName(), factory.newFactory(params, docLookup));
+          new VirtualFieldDef(
+              vf.getName(),
+              factory.newFactory(ScriptFactoryContext.builder(params, docLookup).build()));
```
aprudhomme: It would be good if this script has access to the shared context. The shared context is created above in this class during query processing. It can be created earlier and passed wherever needed.

src/main/java/com/yelp/nrtsearch/server/search/SearchRequestProcessor.java
```diff
@@ -576,7 +579,7 @@ private static RescoreTask buildRescoreTask(
           ScriptService.getInstance().compile(script, ScoreScript.CONTEXT);
       Map<String, Object> params = ScriptParamsUtils.decodeParams(script.getParamsMap());
       DocLookup docLookup = indexState.docLookup;
-      rescoreOperation = new ScriptRescore(factory.newFactory(params, docLookup));
+      rescoreOperation = new ScriptRescore(factory, params, docLookup);
```
aprudhomme: It should be possible to provide the shared document accessor here, or even provide the script factory context directly (if params is removed).

src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
```diff
@@ -648,6 +650,23 @@ record RetrieverResult(TopDocs topDocs, double searchTimeMs, double rescoreTimeM
...
+   * "text"}, the key {@code "retriever_text"} is set on each document's context map, accessible in
+   * JS scripts as {@code _shared_retriever_text} and in ScoreScript subclasses via {@code
+   * get_shared_double("retriever_text", defaultValue)}.
+   */
+  static void populateRetrieverScores(TopDocs hits, SharedDocContext sharedDocContext) {
+    for (ScoreDoc scoreDoc : hits.scoreDocs) {
+      if (scoreDoc instanceof BlendedScoreDoc blended) {
+        Map<String, Object> ctx = sharedDocContext.getContext(scoreDoc.doc);
+        for (Map.Entry<String, ScoreDoc> entry : blended.getScoreDocs().entrySet()) {
+          ctx.put("retriever_" + entry.getKey(), (double) entry.getValue().score);
```
aprudhomme: Could the prefix be a static constant somewhere?

src/main/java/com/yelp/nrtsearch/server/query/QueryNodeMapper.java
```diff
@@ -298,7 +299,8 @@ private FunctionScoreQuery getFunctionScoreQuery(
         ScriptParamsUtils.decodeParams(functionScoreQuery.getScript().getParamsMap());
     return new FunctionScoreQuery(
         getQuery(functionScoreQuery.getQuery(), context),
-        scriptFactory.newFactory(params, context.docLookup()));
+        scriptFactory.newFactory(
+            ScriptFactoryContext.builder(params, context.docLookup()).build()));
```
aprudhomme: It might be good if the scripts here also have shared context access (when available). I'm not sure the best way to pass this. There is already a `QueryContext` that is used in some instances. We generally will need to clean this up sometime.

src/main/java/com/yelp/nrtsearch/server/script/ScoreScript.java
```diff
@@ -110,6 +128,27 @@ public double get_score() {
...
+   * Get a {@code double} value from the shared doc context for the current document.
+   *
+   * <p>For multi-retriever queries, each retriever's raw score is stored under {@code
+   * retriever_<name>} (e.g. {@code "retriever_text"}, {@code "retriever_knn"}).
+   *
+   * @param key shared doc context key
+   * @param defaultValue returned when no entry is present for this key and document
+   * @return stored value cast to {@code double}, or {@code defaultValue}
+   */
+  public double get_shared_double(String key, double defaultValue) {
```
aprudhomme, line 142: Is there a reason this method needs to only apply to double? The values in the shared context map can be of arbitrary type. This might even be able to be a generic method (would need testing with painless).

You could also just provide the map itself.

waziqi89: This is limited by the JS bindings where only doubleValueSource is supported
```
public abstract class Bindings {

  /** Sole constructor. (For invocation by subclass constructors, typically implicit.) */
  protected Bindings() {}

  /** Returns a DoubleValuesSource bound to the variable name */
  public abstract DoubleValuesSource getDoubleValuesSource(String name);
}
```

waziqi89: But we can just allow others like painless to do more. And marking this as a drawback of JS

src/main/java/com/yelp/nrtsearch/server/script/ScriptFactoryContext.java
waziqi89: https://yelp.slack.com/archives/C09PUNMQ1EV/p1780513329459539

src/main/java/com/yelp/nrtsearch/server/rescore/ScriptRescore.java
```diff
@@ -53,14 +60,18 @@ public class ScriptRescore implements RescoreOperation {
...
-  public ScriptRescore(DoubleValuesSource scriptSource) {
-    this.scriptSource = scriptSource;
+  public ScriptRescore(ScoreScript.Factory factory, ScriptFactoryContext scriptFactoryContext) {
+    this.factory = factory;
+    this.scriptFactoryContext = scriptFactoryContext;
   }
 
   @Override
   public TopDocs rescore(TopDocs hits, RescoreContext context) throws IOException {
+    DoubleValuesSource scriptSource = factory.newFactory(scriptFactoryContext);
```
waziqi89, line 73: scoreScript is both rescorer and script, it can access to the sharedDocContext either way. We choose to follow the scriptFactoryContext to set up eagerly

aprudhomme: I'm not sure I understand. Why does the factory need to be created here? How is this different than doing it at creation?

src/main/java/com/yelp/nrtsearch/server/script/ScoreScript.java
```diff
@@ -110,6 +129,42 @@ public double get_score() {
...
```
aprudhomme, line 142: Is there a reason this and the other method names are in snake case?

src/main/java/com/yelp/nrtsearch/server/script/ScoreScript.java
aprudhomme: You could still potentially do something like `<T> T getSharedValue(String key, T default, Class<T> clazz)` and have the js binding invoke with `getSharedValue("some_key", 100, Number.class).doubleValue()`, though I would need to see how painless would react.

src/main/java/com/yelp/nrtsearch/server/script/ScoreScript.java
```diff
@@ -110,6 +129,42 @@ public double get_score() {
...
+   * Get the full shared doc context map for the current document. Allows script plugins to access
+   * values of any type stored in the shared context.
+   *
+   * <p>Not available through JS expressions, which are limited to {@code double} values.
+   *
+   * @return shared doc context map for the current document, or an empty map if no context exists
+   */
+  public Map<String, Object> get_shared_doc_context_map() {
+    if (sharedDocContext == null) {
+      return Collections.emptyMap();
```
aprudhomme, line 163: nit: I think we have started using `Map.of()` instead

src/main/java/com/yelp/nrtsearch/server/script/ScoreScript.java
waziqi89: [get_score](https://github.com/Yelp/nrtsearch/blob/main/src/main/java/com/yelp/nrtsearch/server/script/ScoreScript.java#L101) is using snake case here to expose the api. But if there are some other reasons get_score must be in snake_case which doesn't apply to the new methods, we can convert back to camelCase

waziqi89: fixed

src/main/java/com/yelp/nrtsearch/server/script/ScoreScript.java
waziqi89: I kept the method to return ctx map and let the user to case the type and handling the default

waziqi89:
close as we target to development

PR #983, opened 2026-05-26
add min score query 

waziqi89:
RP-15663
publish the minScoreQuery wrapper to avoid confusion on using multifunctionScoreQuery


swethakann:
We currently support min_score in nrtsearch in a multi-function score query. Is there a use case needing this in the high level query as well? I am trying to understand if we want to consolidate or re-use the existing one.

waziqi89:
> We currently support min_score in nrtsearch in a multi-function score query. Is there a use case needing this in the high level query as well? I am trying to understand if we want to consolidate or re-use the existing one.

CCI is actually using the multiFuncScoreQuery now as the workaround. This is just a nice-to-have feature, as the wrapper is already added for vectorSearch, and we just publish API here

swethakann (dismissed):

clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -744,6 +744,9 @@ message Query {
         ExactVectorQuery exactVectorQuery = 27;
         CrossIndexQuery crossIndexQuery = 28;
     }
+    // Optional minimum score threshold. Documents with scores below this value are excluded.
+    // Defaults to 0 (no filtering). Negative values are invalid.
+    optional float min_score = 29;
```
aprudhomme: Rather than being an option, should this just be its own query type?

waziqi89: We've already set up [this](https://github.com/Yelp/nrtsearch/blob/main/clientlib/src/main/proto/yelp/nrtsearch/search.proto#L1685) the optional for knn. Adding this way instead of a standalone query is more consistent.

aprudhomme: A `KnnQuery` does not expose a query structure, so it needs to have this option as a parameter. I don't think this is necessary when you have full control of the query building. It is not really doing anything special, I think it could just be defined as its own query type to avoid a special api carveout.

waziqi89: fixed

swethakann (approved):

aprudhomme (approved):

PR #984, opened 2026-05-27
Aggregations and Facets Support for Multi-Retriever

swethakann:
Removes the validation that blocked facets and collectors on multi-retriever requests, and implements aggregation execution as a second pass against the union query after blending.

  - **SearchRequestProcessor:** Removed IllegalArgumentException guards for getFacetsCount() and getCollectorsCount() on the multi-retriever path. The union query is already built from all retriever queries; addDrillDowns wraps it in a DrillDownQuery when facets are present, covering both paths.
  - **SearchHandler:** After executeMultiRetriever returns blended hits, runs a second aggregation pass against the union query — via DrillSideways if facets are present (collecting facet counts and custom collectors in one pass), or a plain searcher.search if only collectors are requested. Blended hits are passed as the topDocsForSample source for sampleTopDocs facets so sampling reflects the final ranked order.
  - **Refactor:** Extracted runDrillSidewaysSearch and executeSearch helpers to eliminate duplicated DrillSideways construction, timeout unwrapping, and search logic between the multi-retriever and single-search paths. No behavior change to the single-search path.

### Tests

Added category field (ATOM, SORTED_SET_DOC_VALUES) to the multi-retriever test index and added four new tests: facets only, collectors only, facets + collectors together, and facets with sampleTopDocs.

### DIFF
--- a/clientlib/src/main/proto/yelp/nrtsearch/search.proto
+++ b/clientlib/src/main/proto/yelp/nrtsearch/search.proto
@@ -1120,7 +1120,7 @@ message SearchResponse {
             // Vector diagnostics, only populated for KNN retrievers
             VectorDiagnostics vectorDiagnostics = 4;
             // Whether collection for this retriever was cut short by a timeout
-            bool hadTimeout = 5;
+            bool hitTimeout = 5;
             // Whether collection for this retriever terminated early (terminateAfter reached)
             bool terminatedEarly = 6;
         }
--- a/clientlib/src/main/proto/yelp/nrtsearch/search.proto
+++ b/clientlib/src/main/proto/yelp/nrtsearch/search.proto
@@ -1119,6 +1119,10 @@ message SearchResponse {
             TotalHits totalHits = 3;
             // Vector diagnostics, only populated for KNN retrievers
             VectorDiagnostics vectorDiagnostics = 4;
+            // Whether collection for this retriever was cut short by a timeout
+            bool hadTimeout = 5;
+            // Whether collection for this retriever terminated early (terminateAfter reached)
+            bool terminatedEarly = 6;
         }
         // Diagnostics specific to multi-retriever search
         message MultiRetrieverDiagnostics {
@@ -1430,6 +1434,8 @@ message ProfileResult {
     message MultiRetrieverProfileResult {
         // Per-retriever profile results keyed by retriever name
         map<string, ProfileResult> retrieverProfileResults = 1;
+        // Profile result for the aggregation pass (facets + collectors) run after blending
+        ProfileResult aggregationProfileResult = 2;
     }
 
     // Stats for search phase
--- a/src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
+++ b/src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
@@ -93,6 +93,10 @@ public class SearchHandler extends Handler<SearchRequest, SearchResponse> {
   private final ExecutorService searchExecutor;
   private final boolean warming;
 
+  /** Return value of {@link #executeMultiRetriever}. */
+  private record MultiRetrieverResult(
+      TopDocs topDocs, boolean hadTimeout, boolean terminatedEarly) {}
+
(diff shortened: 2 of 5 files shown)


clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -1119,6 +1119,10 @@ message SearchResponse {
             TotalHits totalHits = 3;
             // Vector diagnostics, only populated for KNN retrievers
             VectorDiagnostics vectorDiagnostics = 4;
+            // Whether collection for this retriever was cut short by a timeout
+            bool hadTimeout = 5;
```
waziqi89: Shall be named as `hitTimeout` for consistency?

src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
```diff
@@ -93,6 +93,10 @@ public class SearchHandler extends Handler<SearchRequest, SearchResponse> {
   private final ExecutorService searchExecutor;
   private final boolean warming;
 
+  /** Return value of {@link #executeMultiRetriever}. */
+  private record MultiRetrieverResult(
```
waziqi89, line 97: Maybe we want to put those records in a separate class now as their complexity are growing

src/main/java/com/yelp/nrtsearch/server/search/SearchRequestProcessor.java
```diff
@@ -250,7 +250,9 @@ public static SearchContext buildContextForRequest(
       contextBuilder.setCollector(buildDocCollector(collectorCreatorContext));
     }
 
-    // Facets are applied on the union query (all matching docs) for multi-retriever requests.
+    // Facets are applied on the union query for multi-retriever requests. Counts reflect the full
```
waziqi89: Comments are a bit confusing as this line applies to both flow

waziqi89: I also see the difference of collector context between two is only the numOfHits. Shall we consolidate two branches into one place here?

src/test/java/com/yelp/nrtsearch/server/search/MultiRetrieverSearchTest.java
```diff
@@ -89,6 +89,9 @@ protected void initIndex(String name) throws Exception {
                   MultiValuedField.newBuilder()
                       .addValue(String.format("[%f, %f, %f]", i * 0.1f, i * 0.2f, i * 0.3f))
                       .build())
+              .putFields(
+                  "category",
+                  MultiValuedField.newBuilder().addValue(i <= 5 ? "odd" : "even").build())
```
waziqi89: NIT: this "def of even/odd" is a bit weird to understand

src/test/java/com/yelp/nrtsearch/server/search/MultiRetrieverSearchTest.java
```diff
@@ -227,31 +254,131 @@ public void testValidationErrors() {
...
-                "terms",
-                Collector.newBuilder()
-                    .setTerms(TermsCollector.newBuilder().setField("doc_id").setSize(5).build())
-                    .build())
-            .build(),
-        "Collectors are not supported with MultiRetriever requests");
+    assertEquals(1, response.getFacetResultCount());
+    assertEquals("category_facet", response.getFacetResult(0).getName());
+    // All 10 docs matched (union of both retrievers), split into two categories
+    assertEquals(2, response.getFacetResult(0).getLabelValuesCount());
```
waziqi89, line 270: Do we want to assert the each child count

src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
swethakann: I don't see any new field additions immediately here. In case that happens, we can move it to a class then. It might be an overkill for now since it is just used here.

waziqi89 (approved):

aprudhomme (approved):

PR #986, opened 2026-05-29
allow clients to drop the non-positive scored docs in score script

waziqi89:
In the score script rescorer, we don't have support to drop the unwanted docs.
After this, clients can set the score of those to 0.


clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -1362,6 +1362,9 @@ message QueryRescorer {
 message ScriptRescorer {
     // Script to execute per document; must return a double
     Script script = 1;
+    // When true, documents whose final script score is <= 0 are dropped from the result set.
+    // Defaults to false (non-positive scores are kept, matching previous behaviour).
+    bool filter_non_positive = 2;
```
aprudhomme, line 1367: I'm not sure if this filter should be 0 inclusive, since it is a valid document score

waziqi89: I thought 0-score means unmatch, thereby shall drop. But looks like it's not the case in Lucene, 0 is still a "match but no contribution".

waziqi89: I will close this PR for now as I don't know if clients want to drop the doc by setting a negative score

PR #995, opened 2026-06-15
Count accurate totalhits for aggregations

swethakann:
When facets or collectors are requested in a multi-retriever search, the search executes against the union query. Previously, totalHits in the response was taken from the blender output — bounded by the per-retriever recall windows and always GREATER_THAN_OR_EQUAL_TO. 

This change overrides it with the accurate count from the aggregation search result, which reflects true union-query total hits and carries the correct EQUAL_TO / GREATER_THAN_OR_EQUAL_TO relation based on totalHitsThreshold.


waziqi89 (approved):
you may want to backport this to 1.3 branch.
Main is 1.4 now

PR #998, opened 2026-07-08
Nrtsearch configurable s3 upload with refresh

taoyyu:
Created a new boolean config s3RefreshUpload with default false
when s3RefreshUpload is true, primary would upload to s3 in parallel with grpc data replication.
Replicas stay the same, only getting refresh updates from grpc for now.
Isolated replicas may get refresh updates during commits.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/nrt/NrtDataManager.java
+++ b/src/main/java/com/yelp/nrtsearch/server/nrt/NrtDataManager.java
@@ -65,6 +65,7 @@ public class NrtDataManager implements Closeable {
   private final RemoteBackend remoteBackend;
   private final RestoreIndex restoreIndex;
   private final boolean remoteCommit;
+  private final boolean s3RefreshUpload;
 
   // Set during startUploadManager
   private NRTPrimaryNode primaryNode;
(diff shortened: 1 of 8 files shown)


src/main/java/com/yelp/nrtsearch/server/nrt/NrtDataManager.java
```diff
@@ -346,8 +359,9 @@ public InputStream downloadIndexFile(String fileName, NrtFileMetaData fileMetaDa
    * @param watchers List of RefreshUploadFuture objects to notify when the upload is complete
    */
   public synchronized void enqueueUpload(CopyState copyState, List<RefreshUploadFuture> watchers) {
-    if (!remoteCommit) {
-      throw new IllegalStateException("Remote commit is not available for this configuration");
+    if (!remoteCommit && !s3RefreshUpload) {
```
aprudhomme: In what situation would the refresh upload be enabled, but remote commit would not?

taoyyu: oops, I'll fix this and add validator to make sure remoteCommit is true when s3RefreshUpload is enabled.

aprudhomme (approved):

PR #999, opened 2026-07-10
Add option to use adaptive concurrency for bootstrapping index download

aprudhomme:
# Problem
Infrastructure for a cluster is not always homogeneous. Due to differences in resource limits (network bandwidth, disk io, etc) it is hard to use a single value to tune the index download during bootstrapping. Additionally, resources are shared by other services running of the same host. Can we provide a means to dynamically adjust the download concurrency based on the resources available.

My hope is that this will avoid resource issues when using the crt async client.

# Solution

  Replaces the raw Semaphore in downloadBatch with a ConcurrencyLimiter interface, with two implementations:

  - StaticConcurrencyLimiter — wraps the existing Semaphore behavior; default when adaptiveConcurrency.enabled is false.
  - AdaptiveConcurrencyLimiter — dynamically adjusts concurrency based on observed throughput.

## Algorithm

  The adaptive limiter uses a dual-EMA throughput gradient over a sliding time window:

  1. Bytes transferred are fed continuously via a TransferListener attached to each download request. This gives an aggregate signal across all concurrent transfers and all CRT multipart parts — avoiding the per-file sampling bias where large files (split into many parts internally) would skew measurements.
  2. Every windowDurationMs (default 2s), the accumulated bytes are converted to a throughput sample (bytes/sec).
  3. Two exponential moving averages are updated: a fast short EMA (α=0.3) and a slow long EMA (α=0.1).
  4. The gradient shortEma / longEma drives concurrency adjustments:
    - ≥ 1.0 — throughput stable or improving → increase by 10% (additive)
    - < 0.85 — throughput degrading → decrease by 25% (multiplicative)
    - otherwise → hold
  5. Download errors trigger an immediate multiplicative decrease regardless of the window schedule.
  6. A configurable warmup period (default 3 windows) allows increases only, preventing cold-start penalties during connection pool warm-up.
  6. A configurable warmup period (default 3 windows) allows increases only, preventing cold-start penalties during connection pool warm-up.
  7. maxLimit (default 100) is a hard ceiling that acts as a safety net regardless of gradient signals.

## Changes

  - ConcurrencyLimiter — new interface (acquire, onSuccess, onError, recordBytes)
  - StaticConcurrencyLimiter — fixed-limit implementation
  - AdaptiveConcurrencyLimiter — throughput-gradient implementation
  - S3Backend.AdaptiveConcurrencyConfig — new nested config class at remoteConfig.s3.adaptiveConcurrency.*
  - S3Backend.downloadBatch — uses ConcurrencyLimiter + attaches a TransferListener to feed byte deltas to the limiter
  - docs/server_configuration.rst — documents the new remoteConfig.s3.adaptiveConcurrency.* settings and the previously undocumented retry settings

## Configuration
```
  remoteConfig:
    s3:
      adaptiveConcurrency:
        enabled: true
        initialLimit: 16   # starting concurrency
        minLimit: 1
        maxLimit: 100      # hard ceiling — prevents GC death spiral
        windowDurationMs: 2000
        warmupWindows: 3
```

  All other parameters (shortAlpha, longAlpha, decreaseThreshold, decreaseFactor) have sensible defaults and should not need tuning in most cases. The feature is opt-in; existing behavior is unchanged when
  enabled: false.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/remote/s3/AdaptiveConcurrencyLimiter.java
+++ b/src/main/java/com/yelp/nrtsearch/server/remote/s3/AdaptiveConcurrencyLimiter.java
@@ -290,17 +290,20 @@ public class AdaptiveConcurrencyLimiter implements ConcurrencyLimiter {
   /** Release a permit, absorbing it against outstanding debt if any exists. */
   private void releasePermit() {
     inflight.decrementAndGet();
-    while (true) {
-      int debt = permitDebt.get();
-      if (debt <= 0) {
-        semaphore.release();
-        return;
-      }
-      if (permitDebt.compareAndSet(debt, debt - 1)) {
+    // Fast path: no debt, release immediately without acquiring the lock.
+    if (permitDebt.get() <= 0) {
+      semaphore.release();
+      return;
+    }
+    // Slow path: acquire the lock so the absorption is atomic with respect to
+    // applyLimitChange, which also modifies permitDebt under the same lock.
+    synchronized (this) {
+      if (permitDebt.get() > 0) {
+        permitDebt.decrementAndGet();
         // Permit absorbed — do not release to the semaphore.
-        return;
+      } else {
+        semaphore.release();
       }
-      // CAS missed; retry.
     }
   }
 }
--- a/src/main/java/com/yelp/nrtsearch/server/remote/s3/AdaptiveConcurrencyLimiter.java
+++ b/src/main/java/com/yelp/nrtsearch/server/remote/s3/AdaptiveConcurrencyLimiter.java
@@ -1,0 +1,306 @@
+/*
+ * Copyright 2025 Yelp Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.yelp.nrtsearch.server.remote.s3;
+
+import java.util.concurrent.Semaphore;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A {@link ConcurrencyLimiter} that dynamically adjusts the concurrency limit using a
+ * throughput-gradient algorithm driven by a sliding time window.
+ *
+ * <p>Throughput is measured as total bytes received across <em>all</em> concurrent transfers in the
+ * current time window, fed continuously via {@link #recordBytes(long)} from a {@link
+ * software.amazon.awssdk.transfer.s3.progress.TransferListener}. This is intentionally aggregate
+ * rather than per-file: the CRT S3 client downloads large files as many parallel parts, so a
+ * per-file sample conflates internal part-level parallelism with actual IO saturation. The
+ * time-window approach averages cleanly across files of any size.
+ *
+ * <p>Every {@code windowDurationMs} a new throughput sample is computed (bytes / elapsed seconds)
+ * and fed into two EMAs:
+ *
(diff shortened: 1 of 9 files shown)


src/main/java/com/yelp/nrtsearch/server/remote/s3/AdaptiveConcurrencyLimiter.java
```diff
@@ -0,0 +1,306 @@
...
+  }
+
+  /**
+   * Apply a limit change. Must be called while holding {@code this} lock so that the limit and
+   * semaphore state are updated atomically with respect to other adjustments.
+   */
+  private void applyLimitChange(int oldLimit, int newLimit) {
+    currentLimit.set(newLimit);
+    int delta = newLimit - oldLimit;
+    if (delta > 0) {
```
vim345, line 274: AI is giving warning on this line:

applyLimitChange (under this lock) does a non-atomic read-then-modify of permitDebt, but releasePermit() CAS-decrements it concurrently without the lock. This can drive permitDebt negative, causing
  future permit releases to bypass debt absorption — temporarily allowing inflight operations to exceed currentLimit.

Fix: Use a CAS loop in applyLimitChange for the debt payoff, or make releasePermit also acquire the lock (at the cost of throughput on the completion path).

src/main/java/com/yelp/nrtsearch/server/remote/s3/S3Backend.java
```diff
@@ -867,6 +1019,41 @@ public void downloadIndexFiles(
...
+        long delta = prev == null ? current : current - prev;
+        if (delta > 0) {
+          limiter.recordBytes(delta);
+        }
+      }
+    };
+  }
+
+  @VisibleForTesting
+  ConcurrencyLimiter createConcurrencyLimiter(int maxConcurrency) {
```
vim345, line 1021: The retry loop in downloadIndexFiles halves maxConcurrency each round and passes it to downloadBatch → createConcurrencyLimiter. But when adaptive mode is enabled, a fresh limiter is created from
  config values, completely ignoring the reduced concurrency. The downloadRetryReduceConcurrency setting becomes a no-op.

  The RST docs mention this ("Has no effect when adaptiveConcurrency.enabled is true"), but the adaptive limiter doesn't benefit from the error signal either — each retry round creates a brand-new
  limiter that starts fresh from initialLimit, losing all previously-learned throughput state.

src/main/java/com/yelp/nrtsearch/server/remote/s3/S3Backend.java
```diff
@@ -867,6 +1019,41 @@ public void downloadIndexFiles(
     }
   }
 
+  private TransferListener createThroughputListener(ConcurrencyLimiter limiter) {
+    // lastSeenBytes tracks the last cumulative snapshot per request, same as
+    // S3ProgressListenerImpl,
+    // so we can compute the delta on each bytesTransferred callback.
+    ConcurrentHashMap<Object, Long> lastSeenBytes = new ConcurrentHashMap<>();
+    return new TransferListener() {
```
vim345: createThroughputListener uses the same ConcurrentHashMap<Object, Long> + cumulative-to-delta pattern as S3ProgressListenerImpl in the same package. Both are attached to the same download request
  simultaneously, so each bytesTransferred callback hits two ConcurrentHashMap lookups. A shared delta-computation helper or a callback hook on the existing listener would eliminate the duplication and
  halve the per-callback overhead.

vim345 (commented):
AI found some potential issues.

src/main/java/com/yelp/nrtsearch/server/remote/s3/AdaptiveConcurrencyLimiter.java
aprudhomme: Fixed. `applyLimitChange` held the lock but did a non-atomic read-then-modify of `permitDebt` while `releasePermit` CAS-decremented it concurrently without the lock, which could drive `permitDebt` negative and allow inflight to exceed `currentLimit`.

Fixed by making `releasePermit` acquire the lock when it needs to absorb debt. The fast path (no debt) stays lock-free to avoid contention on the common completion path.

src/main/java/com/yelp/nrtsearch/server/remote/s3/S3Backend.java
aprudhomme: I don't see an issue with this behavior. Adaptive concurrency would discover the limit based on the current conditions, which may change between the retries. Starting fresh still seems reasonable.
The error penalty is instant, as it immediately lowers the target concurrency.

vim345 (approved):

PR #1000, opened 2026-07-13
Fix L1 quality for multi-retriever

swethakann:
When a multi-retriever has an L1 rescorer with windowSize > topHits, expand collection to windowSize so the rescorer sees the full candidate pool, then truncate back to topHits after rescoring (keeping the best N by L1 score) before feeding to the blender.

  - RetrieverContext — added topHits field
  - SearchRequestProcessor — numHitsToCollect = max(topHits, rescorer.windowSize)
  - SearchHandler — truncate TopDocs to topHits after rescore

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
+++ b/src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
@@ -562,6 +562,12 @@ public class SearchHandler extends Handler<SearchRequest, SearchResponse> {
                   long rescoreStart = System.nanoTime();
                   topDocs = retrieverContext.getRescoreTask().rescore(topDocs, searchContext);
                   rescoreTimeMs = (System.nanoTime() - rescoreStart) / 1_000_000.0;
+                  int topHits = retrieverContext.getTopHits();
+                  if (topDocs.scoreDocs.length > topHits) {
+                    ScoreDoc[] truncated = new ScoreDoc[topHits];
+                    System.arraycopy(topDocs.scoreDocs, 0, truncated, 0, topHits);
+                    topDocs = new TopDocs(topDocs.totalHits, truncated);
+                  }
                 }
                 return new RetrieverResult(
                     topDocs,
--- a/src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
+++ b/src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
@@ -562,12 +562,7 @@ public class SearchHandler extends Handler<SearchRequest, SearchResponse> {
                   long rescoreStart = System.nanoTime();
                   topDocs = retrieverContext.getRescoreTask().rescore(topDocs, searchContext);
                   rescoreTimeMs = (System.nanoTime() - rescoreStart) / 1_000_000.0;
-                  int topHits = retrieverContext.getTopHits();
-                  if (topDocs.scoreDocs.length > topHits) {
-                    ScoreDoc[] truncated = new ScoreDoc[topHits];
-                    System.arraycopy(topDocs.scoreDocs, 0, truncated, 0, topHits);
-                    topDocs = new TopDocs(topDocs.totalHits, truncated);
-                  }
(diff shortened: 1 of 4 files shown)


aprudhomme (dismissed):

src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
```diff
@@ -562,6 +562,12 @@ record RetrieverResult(
                   long rescoreStart = System.nanoTime();
                   topDocs = retrieverContext.getRescoreTask().rescore(topDocs, searchContext);
                   rescoreTimeMs = (System.nanoTime() - rescoreStart) / 1_000_000.0;
+                  int topHits = retrieverContext.getTopHits();
```
waziqi89: Why do we add this truncation here? we don't do this in the default path, I don't see particular reason why this shall be different.

src/main/java/com/yelp/nrtsearch/server/search/multiretriever/RetrieverContext.java
```diff
@@ -34,6 +34,7 @@ public enum RetrieverType {
   private final Query query;
   private final DocCollector docCollector;
   private final RescoreTask rescoreTask;
+  private final int topHits;
```
waziqi89, line 37: I guess you add this for truncation.

src/test/java/com/yelp/nrtsearch/server/search/MultiRetrieverSearchTest.java
```diff
@@ -713,6 +713,72 @@ public void testProfilingPopulatesPerRetrieverResults() {
...
+
+    // Only 3 docs fed to blender (truncated after rescore), all with score 7.0
+    assertEquals(3, response.getHitsCount());
+    for (Hit hit : response.getHitsList()) {
+      assertEquals(7.0, hit.getScore(), SCORE_DELTA);
+    }
+  }
+
+  @Test
+  public void testL1RescorerTruncatesToTopHitsBeforeBlend() {
```
waziqi89, line 735: Ain't the two tests the same

waziqi89: we may to do set up a stricter test:
topHits 3, rescorer window 10. docs ranked 8-10 are rescored to the top so they end up showing the results. This cannot be the case if only 3 got rescored

src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
waziqi89: https://github.com/Yelp/nrtsearch/blob/57de2d09077947dd4b98146a297a88fa37bd5031/src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java#L286

I went through the code again. We do slice based on the topHits in default path. Then we can reuse this method `getHitsFromOffset(topDocs, 0, retriever.topHits)`

src/test/java/com/yelp/nrtsearch/server/search/MultiRetrieverSearchTest.java
swethakann: I have combined the tests in the follow up commit if that makes it easier for you. 

We don't need the quality check at the moment imo. Re-ranking at L1 is done by inference plugin in prod so we only need a functionality check. I had verified the inference code does score descending yesterday before this change. So that's good.

waziqi89: I think the gap missed testing is:
`all docs: big number - firstPassSearch(topHits:3) -> filter docs: 10 (verify this, not 3) -> rescorer(window: 10) -> blender&response 3`

swethakann: After checking the code, the change is pretty trivial but the test set up was complex. Instead captured it in a simple method that is easy to isolate for testing.

waziqi89 (approved):

PR #1008, opened 2026-08-03
replica pull from s3 during refresh with config

taoyyu:
Added config s3ReplicaRefreshDownload with default false
when s3ReplicaRefreshDownload is true
During refresh or commit
1. replica receives push notification from primary
2. downloads the latest point state from s3
3. create a RemoteCopy Job that downloads file from s3
4. set the latest point state

It is a little tricky for segment merge and I'd like to get feedback here. I haven't made any changes about the code here
So 
1. primary still tries to copy merged segment files to replicas and replicas will throw errors and got skipped
2. in the next refresh, replicas will download the files from s3.
3. Since our refresh is usually every second, this is not too bad maybe?

This is a little different from isolated replicas since isolated replicas does not receive push notifications from primary and primary does not need to know the replicas so primary won't try to copy data to them at all.

I'm planning to make changes to replica to direct return success when primary trying to copy merged data to replicas if the config is true. This will avoid making primaries throw errors and then skip.

Or I can try make it more complicated to make primary aware that these replicas are s3 based replicas and do not do pre copy merge via grpc streaming for these replicas.

One other idea would be not rely on push notifications from primary.

I want to get your thoughts about the precopy merge here @aprudhomme, thanks

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/nrt/NRTReplicaNode.java
+++ b/src/main/java/com/yelp/nrtsearch/server/nrt/NRTReplicaNode.java
@@ -23,6 +23,7 @@ import com.yelp.nrtsearch.server.grpc.ReplicationServerClient;
 import com.yelp.nrtsearch.server.monitoring.NrtMetrics;
 import com.yelp.nrtsearch.server.nrt.jobs.CopyJobManager;
 import com.yelp.nrtsearch.server.nrt.jobs.GrpcCopyJobManager;
+import com.yelp.nrtsearch.server.nrt.jobs.NrtPushRemoteCopyJobManager;
 import com.yelp.nrtsearch.server.nrt.jobs.RemoteCopyJobManager;
 import com.yelp.nrtsearch.server.nrt.jobs.SimpleCopyJob;
 import com.yelp.nrtsearch.server.utils.HostPort;
@@ -74,7 +75,8 @@ public class NRTReplicaNode extends ReplicaNode {
       boolean ackedCopy,
       boolean decInitialCommit,
       boolean filterIncompatibleSegmentReaders,
-      int lowPriorityCopyPercentage)
+      int lowPriorityCopyPercentage,
+      boolean s3ReplicaRefreshDownload)
       throws IOException {
     // the id is always 0, the nodeName is the identifier
     super(0, indexDir, searcherFactory, printStream);
@@ -99,6 +101,8 @@ public class NRTReplicaNode extends ReplicaNode {
               nrtDataManager,
               this,
               isolatedReplicaConfig);
+    } else if (s3ReplicaRefreshDownload) {
+      copyJobManager = new NrtPushRemoteCopyJobManager(nrtDataManager, this);
     } else {
       copyJobManager =
           new GrpcCopyJobManager(indexName, indexId, primaryAddress, ackedCopy, this, id);
--- a/src/main/java/com/yelp/nrtsearch/server/nrt/NRTReplicaNode.java
+++ b/src/main/java/com/yelp/nrtsearch/server/nrt/NRTReplicaNode.java
@@ -211,8 +211,13 @@ public class NRTReplicaNode extends ReplicaNode {
   }
 
   public CopyJob launchPreCopyFiles(
-      AtomicBoolean finished, long curPrimaryGen, Map<String, FileMetaData> files)
+      AtomicBoolean finished,
+      long curPrimaryGen,
+      Map<String, FileMetaData> files,
+      String primaryId,
+      String timeString)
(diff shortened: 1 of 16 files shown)


src/main/java/com/yelp/nrtsearch/server/nrt/jobs/NrtPushRemoteCopyJobManager.java
```diff
@@ -0,0 +1,101 @@
...
+      String reason,
+      Map<String, FileMetaData> files,
+      Map<String, FileMetaData> prevFiles,
+      boolean highPriority,
+      CopyJob.OnceDone onceDone)
+      throws IOException {
+    if (files != null) {
+      throw new IllegalArgumentException(
+          "NrtPushRemoteCopyJobManager does not support merge precopy");
+    }
```
aprudhomme, line 67: I think what will need to happen for the merge handling is that the primary would upload the merge data to s3 prior to notifying the replicas. These replicas would create a `RemoteCopyJob` to download the files from s3 here, and the merge precopy process would continue normally.

taoyyu: ok, so replica should be notified and download from s3 here instead of relying on the next refresh then, right? I'll change the code

taoyyu:
After this commit, the merge behavior is like
1. primary upload merged files to s3
2. primary notify replicas (added primary id and timestamp to the proto)
3. for s3 replicas, it will still stream back transfer status but download from s3 instead of copy from primary. Do you think it is still necessary to stream back status?
4. The primary would keep track of the uploaded merge files to avoid upload again in the next refresh.

Let me know if you have any thoughts

src/main/java/com/yelp/nrtsearch/server/nrt/NrtDataManager.java
```diff
@@ -541,11 +579,16 @@ private Map<String, NrtFileMetaData> uploadDiff(CopyState copyState) throws IOEx
         if (lastFileMetaData != null && isSameFile(fileMetaData, lastFileMetaData)) {
           currentPointFiles.put(fileName, lastFileMetaData);
         } else {
-          String timeString = TimeStringUtils.generateTimeStringSec();
-          NrtFileMetaData nrtFileMetaData =
-              new NrtFileMetaData(fileMetaData, ephemeralId, timeString);
-          currentPointFiles.put(fileName, nrtFileMetaData);
-          filesToUpload.put(fileName, nrtFileMetaData);
+          NrtFileMetaData mergeUploaded = mergePreCopyUploadedFiles.remove(fileName);
```
aprudhomme: If files are re-merged before a refresh, they will never show up in the nrt point file list. This could allow stale entries to build up in the map. We should take steps to make sure the size is not unbounded. I think guava might have some cache constructs to help with this.

src/main/java/com/yelp/nrtsearch/server/nrt/jobs/NrtPushRemoteCopyJobManager.java
```diff
@@ -0,0 +1,144 @@
...
+      Map<String, FileMetaData> files,
+      Map<String, FileMetaData> prevFiles,
+      boolean highPriority,
+      CopyJob.OnceDone onceDone)
+      throws IOException {
+    if (files != null) {
+      return newMergePreCopyJob(reason, files, highPriority, onceDone);
+    }
+
+    NrtDataManager.PointStateWithTimestamp targetPointStateWithTimestamp =
```
aprudhomme: It might be good to break the rest of this method into a separate method. Then `newCopyJob` can just delegate to the correct method.

src/main/java/com/yelp/nrtsearch/server/nrt/NRTReplicaNode.java
```diff
@@ -207,8 +211,13 @@ protected void sendNewReplica() throws IOException {
...
   public CopyJob launchPreCopyFiles(
-      AtomicBoolean finished, long curPrimaryGen, Map<String, FileMetaData> files)
+      AtomicBoolean finished,
+      long curPrimaryGen,
+      Map<String, FileMetaData> files,
+      String primaryId,
+      String timeString)
       throws IOException {
+    copyJobManager.setMergePreCopyMetadata(primaryId, timeString);
     return launchPreCopyMerge(finished, curPrimaryGen, files);
```
aprudhomme: These operations require synchronization. Making this method synchronized should be fine.

Also, it would be good to reset the metadata back to null after launching the merge job. A try/finally block would cover it.

aprudhomme (approved):

PR #1014, opened 2026-08-27
Add embedding plugin framework for automatic text-to-vector conversion

sarthakn7:
## Summary

Today, clients that want to do vector search on nrtsearch must generate embeddings externally before sending them in index/search requests. This adds complexity — clients need to either load the model to generate embeddings or call external services first. This especially adds a barrier for ad-hoc queries, since you'd need to first generate an embedding in some way and then put it in the query.
                                                                                                                                                                                                                                                                                                                            
This PR provides embedding generation in nrtsearch itself via a plugin. Clients can now send plain text and let the server convert it to vectors transparently, at both index time and search time. The text format is up to the plugin - it can be a simple sentence, or it can be a complex json. The plugin architecture provides flexibility to either generate embeddings using local models or call external APIs.
                                                                                               
**Indexing:** If a vector field has embeddingProvider: "some provider" configured, you can send "great food and friendly service" instead of "[0.12, -0.34, ...]". Pre-computed vectors still work — the server auto-detects which format you're sending. Anything not matching a vector is treated as text to be embedded.

**Searching:** Use query_text in a KNN query instead of query_vector:                                                                                                                                                                                                                                                         
```
{               
  "knn": [{                                                                                                                                                                                                                                                                                                               
    "field": "review_embedding",
    "k": 10,                                                                                                                                                                                                                                                                                                              
    "num_candidates": 100,
    "query_text": "delicious Italian food"                                                                                                                                                                                                                                                                                
  }]                                      
}
```                                                                                                                                                                                                                                                                                                                         
                  
You can also override the provider per-query with embedding_provider if you want to use a different model at search time than what the field was indexed with.                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                            
## Writing a new plugin
                                                                                                                                                                                                                                                                                                                            
1. Implement EmbeddingPlugin on your plugin class and return a map of type name → factory
2. Implement EmbeddingProviderFactory to parse your config and instantiate your provider
3. Extend EmbeddingProvider and implement doEmbed(String text) (must be thread-safe)
4. For byte vector field support, also override doEmbedBytes(String text)                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                            
The included nrtsearch-onnx-embedding module is a working reference — it uses ONNX Runtime for inference and DJL's HuggingFace Tokenizer for text preprocessing, and works with any sentence-transformer model exported to ONNX format.                                                                                   

## Tests
                                                                                                                                                                                                                                                                                                                            
- Unit tests for provider registration, text detection heuristic, dimension validation                                                                                                                                                                                                                                    
- Integration test with a mock provider covering full index → search round-trip
- ONNX plugin tests against real all-MiniLM-L6-v2 model (determinism, correct dimensions, different inputs produce different vectors)

### DIFF
--- a/clientlib/src/main/proto/yelp/nrtsearch/luceneserver.proto
+++ b/clientlib/src/main/proto/yelp/nrtsearch/luceneserver.proto
@@ -679,6 +679,12 @@ message Field {
     // Normalizer for ATOM field types. Applies text normalization (e.g. lowercase) consistently
     // at both index time and query time. Predefined options: "lowercase".
     Normalizer normalizer = 38;
+  // Name of a configured embedding provider. When set on a VECTOR field, text strings sent
+  // during indexing are automatically converted to vectors using this provider.
+  // For float vector fields, the provider's embed() method is used.
+  // For byte vector fields, the provider's embedBytes() method is used.
+  // Note: values starting with '[' are always parsed as JSON vector arrays, not as text to embed.
+  string embeddingProvider = 39;
 }
 
 // Options for including IndexPrefixes for field
--- a/src/main/java/com/yelp/nrtsearch/server/embedding/EmbeddingProvider.java
+++ b/src/main/java/com/yelp/nrtsearch/server/embedding/EmbeddingProvider.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright 2026 Yelp Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.yelp.nrtsearch.server.embedding;
+
+/**
+ * Converts text into a float vector embedding. Implementations must be thread-safe, as {@link
+ * #embed(String)} may be called concurrently from multiple index and search threads.
+ *
+ * <p>Subclasses implement {@link #doEmbed(String)} (and optionally {@link #doEmbedBytes(String)})
+ * to produce raw vectors. The public {@link #embed(String)} and {@link #embedBytes(String)} methods
+ * are final and automatically validate that the returned vector length matches {@link
+ * #dimensions()}.
+ */
+public abstract class EmbeddingProvider implements AutoCloseable {
+
+  /** Returns the output vector dimensions of this provider. */
+  public abstract int dimensions();
+
+  /**
+   * Produce a float vector embedding for the given text. Implementations must be thread-safe.
+   *
+   * @param text input text to embed
+   * @return float array of embedding values; length must equal {@link #dimensions()}
+   */
+  protected abstract float[] doEmbed(String text);
+
+  /**
+   * Produce a byte vector embedding for the given text. Implementations must be thread-safe. The
+   * default implementation throws {@link UnsupportedOperationException}; providers that support
+   * byte vector fields must override this method.
+   *
+   * @param text input text to embed
+   * @return byte array of embedding values; length must equal {@link #dimensions()}
+   * @throws UnsupportedOperationException if this provider does not support byte embeddings
+   */
+  protected byte[] doEmbedBytes(String text) {
+    throw new UnsupportedOperationException(
+        "This embedding provider does not support byte vector embeddings");
+  }
+
+  /**
+   * Convert text to a float vector embedding, validating that the result length matches {@link
+   * #dimensions()}.
+   *
+   * @param text input text to embed
+   * @return float array of embedding values
+   * @throws IllegalArgumentException if the returned vector length does not match {@link
+   *     #dimensions()}
+   */
+  public final float[] embed(String text) {
+    float[] result = doEmbed(text);
+    if (result.length != dimensions()) {
+      throw new IllegalArgumentException(
+          "Embedding provider returned vector of size "
+              + result.length
+              + " but provider dimensions() is "
+              + dimensions());
+    }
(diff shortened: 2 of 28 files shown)


clientlib/src/main/proto/yelp/nrtsearch/luceneserver.proto
```diff
@@ -679,6 +679,12 @@ message Field {
     // Normalizer for ATOM field types. Applies text normalization (e.g. lowercase) consistently
     // at both index time and query time. Predefined options: "lowercase".
     Normalizer normalizer = 38;
+  // Name of a configured embedding provider. When set on a VECTOR field, text strings sent
+  // during indexing are automatically converted to vectors using this provider.
+  // For float vector fields, the provider's embed() method is used.
+  // For byte vector fields, the provider's embedBytes() method is used.
+  // Note: values starting with '[' are always parsed as JSON vector arrays, not as text to embed.
+  string embeddingProvider = 39;
```
aprudhomme, line 687: nit: indentation does not match the rest of this block

src/main/java/com/yelp/nrtsearch/server/embedding/EmbeddingProviderFactory.java
```diff
@@ -0,0 +1,29 @@
...
+
+/** Factory for creating {@link EmbeddingProvider} instances from configuration. */
+public interface EmbeddingProviderFactory {
+  /**
+   * Create a named provider instance from config.
+   *
+   * @param config the provider-specific config map from server YAML
+   * @return configured embedding provider
+   */
+  EmbeddingProvider create(Map<String, Object> config);
```
aprudhomme, line 28: If the config is from the server yaml file, it is essentially static and does not need to be passed dynamically as part of the interface. Also, since these objects are created by plugins, there will already be direct access to the server config class during initialization. Do we even want these values in the config file? If they are changed, it will not update any of the index data, which may cause issues.

This parameter would make more sense if configuration was also present in the FieldDef itself. Otherwise, the factory is not really necessary at all and you could keep track of only the initialized providers.

src/main/java/com/yelp/nrtsearch/server/embedding/EmbeddingProvider.java
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * <p>Subclasses implement {@link #doEmbed(String)} (and optionally {@link #doEmbedBytes(String)})
+ * to produce raw vectors. The public {@link #embed(String)} and {@link #embedBytes(String)} methods
+ * are final and automatically validate that the returned vector length matches {@link
+ * #dimensions()}.
+ */
+public abstract class EmbeddingProvider implements AutoCloseable {
+
+  /** Returns the output vector dimensions of this provider. */
+  public abstract int dimensions();
```
aprudhomme, line 30: It might be good to give a little more flexibility here. An embedding provider may be capable of supporting multiple vector sizes. Maybe this could be more like `bool supportsDimension(int dim)`, then `doEmbed` could also include a dimension parameter.

src/main/java/com/yelp/nrtsearch/server/field/VectorFieldDef.java
```diff
@@ -494,16 +563,21 @@ VectorSearchType getSearchType(VectorIndexingOptions vectorIndexingOptions) {
 
     @Override
     void parseVectorField(String value, Document document) {
-      float[] floatArr = null;
-      if (hasDocValues() && docValuesType == DocValuesType.BINARY) {
+      float[] floatArr;
+      if (!looksLikeVectorJson(value)) {
```
aprudhomme, line 567: I'm not sure this auto-detection is needed. If a field sets an embedding provider, it should be assumed that all data should be processed by that provider.

src/main/java/com/yelp/nrtsearch/server/field/VectorFieldDef.java
```diff
@@ -494,16 +563,21 @@ VectorSearchType getSearchType(VectorIndexingOptions vectorIndexingOptions) {
...
+        if (embeddingProvider == null) {
+          throw new IllegalStateException(
+              "No embedding provider configured for field: " + getName());
+        }
+        floatArr = embeddingProvider.embed(value);
+      } else {
         floatArr = parseVectorFieldToFloatArr(value);
```
aprudhomme, line 574: Rather than doing different handling, it may be cleaner to create a core provided `EmbeddingProvider` that turns a json array string into its embedding. This could even serve as the field default implementation.

src/main/java/com/yelp/nrtsearch/server/embedding/EmbeddingProviderFactory.java
sarthakn7: I had added the factory so that the same EmbeddingProvider can be re-used with different parameters, but I think that can be done directly. Removed it.

src/main/java/com/yelp/nrtsearch/server/embedding/EmbeddingProvider.java
sarthakn7: The plugin implementer can create multiple `EmbeddingProvider`s with different dimensions and different names. If there is common code, the plugin implementer can still structure it so that different EmbeddingProviders call the common code. I wanted to have an EmbeddingProvider do a fixed dimension by design.

src/main/java/com/yelp/nrtsearch/server/field/VectorFieldDef.java
sarthakn7: This plugin is intended for use in testing, and it's possible that the vector is provided via addDocument but in some cases it only the text is provided.

sarthakn7: I can still remove the auto-detection if you think it won't work well.

src/main/java/com/yelp/nrtsearch/server/field/VectorFieldDef.java
sarthakn7: A wrapper provider would still need the same branching logic to decide whether to parse JSON or delegate to the real embedding provider, just in a different layer. The current approach keeps it simple with a single check in parseVectorField.

PR #1017, opened 2026-09-02
Add searchStream bidirectional streaming RPC for Query-Then-Fetch

dhruva-r:
## Description

Adds a `searchStream` bidirectional streaming RPC that splits a search into two round-trips, so a coordinator can decide **globally** which documents each shard should fetch and log.

1. Client sends a `SearchRequest`. The shard runs search + rescore, returns scored hit ids — **no field fetch** — and pauses with the searcher held open.
2. Client merges every shard's ranking, then sends back a `ReducedHitList` naming exactly which doc ids this shard should return and which it should log.
3. The shard fetches only those documents, logs only the requested ones, responds, and closes.

Recall-phase responses carry facets, collector results and search state; the fetch response doesn't repeat them. Client deadlines are enforced via `DeadlineUtils` at the same points the unary path checks them, and the recall response uses the same hit window unary would (`max(topHits, hitsToLog + startHit)`) rather than the widest rescorer window.

**The unary `search` path is unchanged.** The `SearchHandler` diff is visibility-only — 6 declarations `private` -> package-private, zero body changes — and the `HitsLoggerFetchTask` change is purely additive.

## Testing


clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -1728,3 +1728,62 @@ message KnnQuery {
...
+        // Initial request: execute search and rescore, pause before fetch.
+        // The server holds the searcher open and returns scored hit IDs.
+        SearchRequest searchRequest = 1;
+        // Continue with a reduced hit list after coordinator merge/dedup.
+        // The server fetches and logs only the specified documents.
+        ReducedHitList reducedHitList = 2;
+    }
+}
+
+// Coordinator-provided lists of document IDs for selective fetch and logging. Sent after the
```
waziqi89: Coordinator is not known to the public

clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -1728,3 +1728,62 @@ message KnnQuery {
...
+// coordinator has merged the RESCORE phase responses of all shards into a single global ranking.
+//
+// All ids must be lucene doc ids that this shard returned in its own RESCORE phase response;
+// ids the shard did not return are rejected with INVALID_ARGUMENT rather than silently ignored,
+// so that a coordinator bug surfaces as an error instead of as missing log records.
+//
+// luceneDocIdsToLog is expected to be a superset of luceneDocIdsToReturn: the coordinator decides
```
waziqi89: for general purpose, two lists do no have to have any overlap or superset relation. The only thing we may want to assert is that both lists are subset of the previously returned list from server.

clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -1728,3 +1728,62 @@ message KnnQuery {
...
+// globally which documents to log, and returns a (usually smaller) page of them to the caller.
+// Ids in luceneDocIdsToReturn that are absent from luceneDocIdsToLog are still fetched and
+// returned, they are simply not logged.
+message ReducedHitList {
+    // Lucene doc IDs whose fields should be fetched and returned over the wire, in the order
```
waziqi89, line 1757: Just wondering: why order matters? Supporting this order may add your code complexity.

clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -1728,3 +1728,62 @@ message KnnQuery {
...
+// Server-to-client message for the searchStream RPC.
+message StreamSearchResponse {
+    // The phase that was last completed when the server produced this response.
+    enum Phase {
+        PHASE_UNSPECIFIED = 0;
+        // First-pass retrieval completed (reserved for future decomposition).
+        SEARCH = 1;
+        // Rescoring completed; hit IDs and scores are available.
+        RESCORE = 2;
+        // Fetch completed; full document fields are available.
```
waziqi89, line 1776: Maybe it shall be called "FETCH_AND_LOG". Not 100% sure as log is a special task of fetch, but in searchRequest, they were shown separately.

clientlib/src/main/proto/yelp/nrtsearch/search.proto
dhruva-r: I'll remove references to internal yelp technologies from comments and other places

src/main/java/com/yelp/nrtsearch/server/handler/SearchStreamHandler.java
```diff
@@ -0,0 +1,732 @@
...
+import org.apache.lucene.search.TopDocs;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Handler for the bidirectional streaming searchStream RPC, which implements the shard side of a
+ * query-then-fetch (QTF) search.
+ *
+ * <p>The client sends a {@link SearchRequest}; the server executes recall and rescoring, then
+ * responds with hit ids and ranking info but no hit fields. The client merges those responses
```
waziqi89, line 72: I think you need hit field, at least the primary ID to do dedup

waziqi89: This might be a bit weird as the retrieveField is the searchRequest is the final retrieveField, and we don't a have intermediate retrievedField to fill in proto

src/main/java/com/yelp/nrtsearch/server/handler/SearchStreamHandler.java
```diff
@@ -0,0 +1,732 @@
...
+    this.globalState = globalState;
+    this.searchHandler = searchHandler;
+    this.maxConcurrentStreams = maxConcurrentStreams;
+    this.idleTimeoutMs = idleTimeoutMs;
+    this.concurrencyLimiter = new Semaphore(maxConcurrentStreams);
+    AtomicInteger threadId = new AtomicInteger();
+    // Idle timeouts only need to release a searcher and write a status, so a small pool is
+    // enough. A slow release can still delay other sessions' timeouts, which delays cleanup but
+    // does not affect any in flight request.
+    this.timeoutScheduler =
```
waziqi89, line 115: I think your implementation on timer is a bit hard to understand. CC suggests to use `AtomicReference<ScheduledFuture<?>> idleTimer = new AtomicReference<>();`
as these: https://fluffy.yelpcorp.com/i/CQjTTf8gscnrm0Z3V39Vd2q4vB5DCDMr.html , https://fluffy.yelpcorp.com/i/fjwzW0XSH75lKMFkwPqS5wWdKVHZb6sl.html

src/main/java/com/yelp/nrtsearch/server/handler/SearchStreamHandler.java
```diff
@@ -0,0 +1,732 @@
...
+    return maxConcurrentStreams - concurrencyLimiter.availablePermits();
+  }
+
+  /** Stop the idle timeout scheduler. Any open sessions are left to be closed by grpc. */
+  public void shutdown() {
+    timeoutScheduler.shutdownNow();
+  }
+
+  /** The per-stream session, holding the searcher and search state across round-trips. */
+  private class SearchStreamSession implements StreamObserver<StreamSearchRequest> {
```
waziqi89, line 164: Do you consider using io.grpc.Context to carry the session context like SearchContext? Then you don't have to create this context-awaring streamObserver.

Lots of things are actually accessible from SearchContext, so you don't have to put them in Context one by one

src/main/java/com/yelp/nrtsearch/server/handler/SearchStreamHandler.java
```diff
@@ -0,0 +1,732 @@
...
+      // the merge, so the merged page would not be the true global page.
+      if (searchRequest.getStartHit() != 0) {
+        throw Status.INVALID_ARGUMENT
+            .withDescription(
+                "startHit must be 0 for searchStream; apply the offset to the merged ranking "
+                    + "instead, since a per shard offset does not compose across shards")
+            .asRuntimeException();
+      }
+
+      // A second search request on the same stream replaces the first, e.g. to re-query with
```
waziqi89, line 253: optional: This is can be an error for now if later this makes your implementation harder

src/main/java/com/yelp/nrtsearch/server/handler/SearchHandler.java
```diff
@@ -394,7 +393,7 @@ public SearchResponse handle(IndexState indexState, SearchRequest searchRequest)
    * @throws ExecutionException on error when performing parallel fetch
    * @throws InterruptedException if parallel fetch is interrupted
    */
-  private void fetchFields(SearchContext searchContext)
+  void fetchFields(SearchContext searchContext)
```
waziqi89, line 396: I am not sure it's good time out move those critical util method out to a util class, or just create a copy of them so that newer classes can directly use those utils instead of carrying a copy of searchHandler

src/main/java/com/yelp/nrtsearch/server/handler/SearchStreamHandler.java
```diff
@@ -0,0 +1,732 @@
...
+
+      DeadlineUtils.checkDeadline("SearchStreamHandler: reduced hit list", diagnostics, "SEARCH");
+      long fetchStartTime = System.nanoTime();
+
+      Map<Integer, ScoreDoc> firstPassHits = new LinkedHashMap<>();
+      for (ScoreDoc scoreDoc : currentHits.scoreDocs) {
+        firstPassHits.put(scoreDoc.doc, scoreDoc);
+      }
+
+      List<Integer> logIds = reducedHitList.getLuceneDocIdsToLogList();
```
waziqi89, line 452: I think if you to check null and empty.
null means keep the current hit as it is.
empty means empty.

src/main/java/com/yelp/nrtsearch/server/logging/HitsLoggerFetchTask.java
```diff
@@ -63,6 +63,33 @@ public void processAllHits(SearchContext searchContext, List<SearchResponse.Hit.
...
+   * <p>The {@code hitsToLog > 0} disable switch honored by {@link #processAllHits(SearchContext,
+   * List)} still applies: only the per-shard truncation is bypassed, not logging itself. {@link
+   * #hitsToLog} is read directly rather than through {@link SearchContext#getHitsToLog()} so the
+   * result does not depend on whether this task is currently attached to the context's {@link
+   * com.yelp.nrtsearch.server.search.FetchTasks}.
+   *
+   * @param searchContext search context
+   * @param hits hits to log
+   */
+  public void logHits(SearchContext searchContext, List<SearchResponse.Hit.Builder> hits) {
```
waziqi89, line 84: I think the better idea is to hijack the searchContext during the build phase for QTF to make it INTEGER.MAX_VALUE instead of creating a new method in this base class

src/test/java/com/yelp/nrtsearch/server/handler/SearchStreamHandlerTest.java
```diff
@@ -0,0 +1,791 @@
+/*
```
waziqi89, line 1: I will review the tests once we have finalize the impl decision

src/main/java/com/yelp/nrtsearch/server/handler/SearchStreamHandler.java
waziqi89: As mentioned in sync , you can just wrap the searchRequest with another message containing RetrievedFields in streamPhase

clientlib/src/main/proto/yelp/nrtsearch/search.proto
waziqi89: Logically speaking, is the relative precedence between hits in shard-local order not the same as global order?

Should we return two SETs then?

src/main/java/com/yelp/nrtsearch/server/search/FetchTasks.java
```diff
@@ -153,7 +168,7 @@ public void processAllHits(SearchContext searchContext, List<SearchResponse.Hit.
     // hitsLogger should be the last fetch task to run because it might need shared data from other
     // plugins, including
     // other fetch task plugins
-    if (hitsLoggerFetchTask != null) {
+    if (hitsLoggerFetchTask != null && !hitsLoggerSuppressed) {
```
waziqi89, line 171: nit: maybe add another processAllHits(searchContext, hits, bool: skipLogging) makes it clearer

clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -1728,3 +1728,89 @@ message KnnQuery {
     // Strategy to use when applying filter
     FilterStrategy filter_strategy = 9;
 }
+
+// Client-to-server message for the searchStream bidirectional streaming RPC.
+message StreamSearchRequest {
+    oneof streamPhase {
+        // Initial request: execute search and rescore, pause before fetch.
+        // The server holds the searcher open and returns scored hit IDs.
+        RecallRequest recallRequest = 1;
```
waziqi89: I think it might be called ranking instead of recall

clientlib/src/main/proto/yelp/nrtsearch/search.proto
```diff
@@ -1728,3 +1728,89 @@ message KnnQuery {
...
+message RecallRequest {
+    // The search to execute. Its retrieveFields describe the hits of the final
+    // (FETCH_AND_LOG phase) response, and are not filled on the RESCORE phase response.
+    SearchRequest searchRequest = 1;
+    // Fields to fill on the hits of the RESCORE phase response, in addition to the ranking info
+    // that is always present. Typically the primary key, which the client needs in order to
+    // deduplicate documents that more than one shard returned before deciding what each shard
+    // should fetch and log. Keep this list small: these fields are fetched for every recalled hit,
+    // while SearchRequest.retrieveFields are only fetched for the documents that survive the merge.
+    repeated string retrieveFields = 2;
```
waziqi89: As referred in the later docstring, we can name it as `intermediateRetrieveFields`

src/main/java/com/yelp/nrtsearch/server/handler/SearchStreamHandler.java
```diff
@@ -0,0 +1,858 @@
...
+      responseObserver.onError(
+          Status.UNAVAILABLE.withDescription("Server is shutting down").asRuntimeException());
+      return new NoOpStreamObserver<>();
+    }
+    return session;
+  }
+
+  /** Number of streams currently holding a permit. */
+  @VisibleForTesting
+  int getActiveStreams() {
```
waziqi89: This is not used yet? Is it created for metrics?

src/main/java/com/yelp/nrtsearch/server/handler/SearchStreamHandler.java
```diff
@@ -0,0 +1,858 @@
...
+ * responds with hit ids, ranking info and any intermediate fields the client asked for, but no
+ * {@link SearchRequest#getRetrieveFieldsList() retrieveFields}. The client merges those responses
+ * across all shards into a single global ranking and sends back a {@link ReducedHitList} naming the
+ * documents this shard should fetch and log. Only then does the server fetch fields and invoke the
+ * {@link HitsLoggerFetchTask}, so a document is logged once globally rather than once per shard.
+ *
+ * <p>The Lucene searcher acquired for the first message is held open until the stream ends, so both
+ * phases see the same index commit and lucene doc ids remain valid as document references.
+ */
+public class SearchStreamHandler {
```
waziqi89: This handler does not actually inherit handler base class. Shall we move it to a new package?

waziqi89: Then maybe we can move something to a base bidistreamhandler class? At least in the future

src/main/java/com/yelp/nrtsearch/server/handler/SearchStreamHandler.java
```diff
@@ -0,0 +1,858 @@
...
+          .run();
+    }
+
+    /**
+     * Fetch fields for, and log, exactly the documents the client selected after its global merge,
+     * then respond with the documents it asked to have returned.
+     */
+    private void handleReducedHitList(ReducedHitList reducedHitList) throws Exception {
+      if (searchContext == null || currentHits == null) {
+        closeWithError(
```
waziqi89: this is an abuse of this method. You throw the grpc.status as runtimeException everywhere else at the handling level, and let the onNext to close the stream.

src/main/java/com/yelp/nrtsearch/server/handler/SearchStreamHandler.java
```diff
@@ -0,0 +1,858 @@
...
+      setResponseCompression(searchRequest.getResponseCompression(), responseObserver);
+
+      IndexState indexState = globalState.getIndex(searchRequest.getIndexName());
+      if (indexState == null) {
+        throw Status.NOT_FOUND
+            .withDescription("Index " + searchRequest.getIndexName() + " not found")
+            .asRuntimeException();
+      }
+      ShardState shardState = indexState.getShard(0);
+      indexState.verifyStarted();
```
waziqi89: this is not an grpc status exception, then will be returned as INTERNAL error

src/main/java/com/yelp/nrtsearch/server/handler/SearchStreamHandler.java
```diff
@@ -0,0 +1,858 @@
...
+        IndexState indexState = searchContext.getIndexState();
+        if (indexState.getWarmer() != null) {
+          indexState.getWarmer().addSearchRequest(searchRequest);
+        }
+      } catch (Exception e) {
+        logger.error("Unable to add warming query", e);
+      }
+    }
+
+    synchronized void startIdleTimeout() {
```
waziqi89: Wonder why can't those two methods merge

src/main/java/com/yelp/nrtsearch/server/handler/SearchStreamHandler.java
```diff
@@ -0,0 +1,858 @@
...
+    }
+
+    /**
+     * Reject any doc id this shard did not return during the recall phase, so that a client bug
+     * surfaces as an error rather than as silently missing hits or log records.
+     */
+    private void verifyKnownDocIds(
+        Set<Integer> knownDocIds, Set<Integer> logIds, Set<Integer> returnIds) {
+      List<Integer> unknown = new ArrayList<>();
+      for (int docId : logIds) {
```
waziqi89: I think Sarthak mentioned that if you want to use Guava to simplify those complement/union/intersect operations, you can add it to the dependency

src/main/java/com/yelp/nrtsearch/server/handler/SearchStreamHandler.java
```diff
@@ -0,0 +1,858 @@
...
+      addToWarmer(searchRequest);
+
+      DeadlineUtils.checkDeadline("SearchStreamHandler: recall response", diagnostics, "SEARCH");
+
+      responseObserver.onNext(
+          StreamSearchResponse.newBuilder()
+              .setPhase(StreamSearchResponse.Phase.RESCORE)
+              .setSearchResponse(responseBuilder)
+              .build());
+      scheduleIdleTimeout();
```
waziqi89: Can we move this "startIdlerTimer" to the onNext? I think all steps except exceptions want to start Timer in the end and you've already cancelled the timer in that method

clientlib/src/main/proto/yelp/nrtsearch/search.proto
waziqi89: Forget how this was named. But I think SQ called ranking + recall, where recall is "fetch"

clientlib/src/main/proto/yelp/nrtsearch/search.proto
dhruva-r: >Logically speaking, is the relative precedence between hits in shard-local order not the same as global order?

We shall return two sets

src/main/java/com/yelp/nrtsearch/server/handler/SearchStreamHandler.java
waziqi89: If the concern is that we want to call onComplete for the one case, I think you can add a if-condition, everything else runs reset timer, reducedList runs onComplete

clientlib/src/main/proto/yelp/nrtsearch/search.proto
dhruva-r: I will change it to ranking instead

src/main/java/com/yelp/nrtsearch/server/handler/SearchStreamHandler.java
dhruva-r: It is used in the SearchStreamHandlerTest

src/main/java/com/yelp/nrtsearch/server/handler/SearchStreamHandler.java
dhruva-r: We can merge those methods

waziqi89 (dismissed):

waziqi89 (approved):

PR #1023, opened 2026-09-10
Support multi-level nested documents in OBJECT field type (PR1: indexing)

aprudhomme:
I am trying to clean up nested documents and add support for multiple nested levels. This is being done in steps on the `development` branch.

----

Replace createChildDocument with createChildDocumentBlock in ObjectFieldDef so that inner OBJECT fields with nestedDoc=true produce their own Lucene child documents instead of being flattened into the parent's child document. Inner children are prepended before their parent in the block, satisfying Lucene's block-join ordering invariant at every nesting depth.

Replace addGlobalNestedDocumentOffsets with addNestedDocumentOffsets in AddDocumentHandler. The new algorithm processes the child-document list right-to-left and assigns each doc an offset pointing to its immediate parent (keyed by _nested_path) rather than always pointing to the root document. Single-level nesting produces identical offsets to the previous implementation.

### DIFF
--- a/src/main/java/com/yelp/nrtsearch/server/field/ObjectFieldDef.java
+++ b/src/main/java/com/yelp/nrtsearch/server/field/ObjectFieldDef.java
@@ -95,32 +95,85 @@ public class ObjectFieldDef extends IndexableFieldDef<Struct> {
       List<Map<String, Object>> fieldValueMaps = new ArrayList<>();
       fieldValues.stream().map(e -> GSON.fromJson(e, Map.class)).forEach(fieldValueMaps::add);
 
-      int totalDocs = fieldValueMaps.size();
-      List<Document> childDocuments = new ArrayList<>(totalDocs);
-
+      List<Document> allDocuments = new ArrayList<>();
       for (Map<String, Object> fieldValueMap : fieldValueMaps) {
-        childDocuments.add(createChildDocument(fieldValueMap, facetHierarchyPaths));
+        allDocuments.addAll(createChildDocumentBlock(fieldValueMap, facetHierarchyPaths));
       }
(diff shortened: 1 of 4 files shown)


dhruva-r (approved):
Overall lgtm, what is the plan for the other PRs ?

src/main/java/com/yelp/nrtsearch/server/field/ObjectFieldDef.java
```diff
@@ -95,32 +95,85 @@ public void parseFieldWithChildren(
...
+      if (childField.getValue() instanceof ObjectFieldDef childObjectDef
+          && childObjectDef.isNestedDoc()) {
+        // Inner nested OBJECT: recursively create child document blocks
+        List<Map<String, Object>> innerValues = new ArrayList<>();
+        if (rawValue instanceof Map) {
+          innerValues.add((Map<String, Object>) rawValue);
+        } else if (rawValue instanceof List) {
+          innerValues.addAll((List<Map<String, Object>>) rawValue);
+        } else {
+          throw new IllegalArgumentException("Invalid data");
```
sarthakn7: Should this error be more descriptive?
```suggestion
         throw new IllegalArgumentException(
      "Expected Map or List for field '" + key + "', got: " + rawValue.getClass().getName());
```

aprudhomme:
> Overall lgtm, what is the plan for the other PRs ?

There are two intended follow up PRs:
- Makes NestedQuery and InnerHit work correctly for inner nested levels
- Verifies _PARENT., _CHILDREN., and KNN work correctly; documents known limitations

sarthakn7 (approved):

PR #1026, opened 2026-09-14
Support multi-level nested documents in OBJECT field type (PR2: queries)

aprudhomme:
Add IndexState.getFieldBaseNestedPath(String, DocLookup) overload so that QueryNodeMapper can resolve the immediate parent path of a nested field using the request-scoped DocLookup without needing to thread IndexState through.

Fix QueryNodeMapper.getNestedQuery to use the immediate parent nested path as the ToParentBlockJoinQuery parent filter instead of always using _root. For a single-level path like "orders", getFieldBaseNestedPath returns "_root" so behavior is unchanged. For a multi-level path like "orders.items", getFieldBaseNestedPath returns "orders", causing items to be joined to their order docs before order docs are joined to root — enabling the canonical nested NestedQuery pattern: NestedQuery(orders, NestedQuery(orders.items, <query>)).


src/test/java/com/yelp/nrtsearch/server/field/MultiLevelNestedTest.java
```diff
@@ -340,6 +346,273 @@ public void testUpdateReplacesAllNestedLevels() throws Exception {
...
+            .setNestedQuery(NestedQuery.newBuilder().setPath("orders").setQuery(innerNestedQuery))
+            .build();
+
+    SearchResponse response = doSearch(outerNestedQuery, List.of("doc_id"));
+    assertDocIds(response, "2");
+  }
+
+  /** Cross-doc isolation: querying for "doohickey" (only in doc_id=2) must return only doc_id=2. */
+  @Test
+  public void testNestedNestedQueryCrossDocIsolation() {
```
sarthakn7: This test is the same as the test above it

sarthakn7 (commented):
Lgtm other than the duplicated test

aprudhomme:
Testing failure looks like the example-plugin issue I just fixed on main

sarthakn7 (approved):

